PR #2, opened 2021-06-25
[WIP] Add initial migration framework + copy existing table structure

lindgrenj6:
Just opening this for now - I used DataGrip to export the current table structure, then wrapped the `up` migration with `CREATE TABLE IF EXISTS` so it won't touch existing dbs but will allow a fresh startup. Currently this is all being done in the `runMigrations` function in the main.go file, we can enable it when we want. This is just more of a "yeah ok heres how we would do this" sort of thing.

---

One fun thing I found is that https://github.com/golang-migrate/migrate actually uses the same `schema_migrations` table as rails, and the migrations will **fail unless that table is recreated**. So basically once we switch over to go migrations there is no going back since we'll lose all the history. I added an ENV var check so we _can_ drop that table one time, but never again hopefully.

\# Things I want to do before actually merging this:
- [ ] change the `fk_rails_8675309` names to actual human-readable ones

EDIT: potentially look at this, designed for GORM: https://github.com/go-gormigrate/gormigrate


lindgrenj6:
Closing in favor of #72

PR #3, opened 2021-06-28
[RHCLOUD-18323] Add Prometheus Metric Exporting

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-18323

Using echo built-in metrics exporter for the grunt work, but using the regular prometheus http handler so we can serve it on a different port.

Definitely just to get discussion goging - will probably move code around to be cleaner.


lindgrenj6:
Ok I modernized this after backlog grooming this morning. A few things to note:

1. Spinning up a separate echo router to listen on `:9000` for the `/metrics` endpoint, echo can't listen on more than one port by default
2. Inserting the prometheus middleware in the API router to record all the requests, then serving those metrics on the separate router
3. Removed the pretty ascii + port output (sorry!) to make the log output straight json when operating in a clowderized env

MikelAlejoBR (approved):
This is interesting :eyes:

lpichler (approved):

lpichler:
There is conflict.

lindgrenj6:
rebased, good to go.

PR #10, opened 2021-08-02
Add support for the Application model

lindgrenj6:
Fairly self-explanatory. I did find out that the silly `Pause` struct I created isn't necessary anymore, since we're using the `gorm.DeletedAt` struct instead. So there are a few changes unrelated, but mostly the same. 

Also looking into how the `gorm.DeletedAt` struct works - not actually sure yet. 

\# TODO:
- [x] DAO
- [x] tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 .idea/
+.vscode/
+.scannerwork/
 sources-api-go
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"fmt"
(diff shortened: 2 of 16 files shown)


lindgrenj6:
So I think we don't actually want the features that `gorm.DeletedAt` brings us - we actually want the paused things to show up in results - whereas just having a `PausedAt` field shows up like normal. I pulled out some of the DeletedAt code here quick. Other than the small refactors its largely the same code as the other endpoints.

lpichler (approved):
Looks good, works perfectly 👍

PR #11, opened 2021-08-02
Fix TenantDAO to actually look up the tenant right

lindgrenj6:
I keep finding more and more bugs with my initial implementation, this was a problem with how it was looking it up + the error that could potentially get returned.

Now we're set. I will probably be making sure we're using the tenant_id properly in some of the DAOs later this week.


lpichler (approved):
Looks good 👍

PR #13, opened 2021-08-04
Add support for hitting the db during tests (aka integration testing)

lindgrenj6:
Ok this is a large one. 

This adds support for running tests against the _actual_ db by running `go test -integration` instead of the regular. I converted the admittedly hacky JSON test data types in the handler tests to regular struct arrays so we can insert them OR use them in the mockDaos. 

Also finding out that GORM itself needs the IDs spelled like `ID` instead of `Id` in order to get associations to work, so I think I got most of those here.


lpichler (approved):
This is awesome 👍 

Looks good, it works great 👍

PR #14, opened 2021-08-04
Add 'sort_by' query parameter support

lindgrenj6:
Added `sort_by` support on list operations, since we have customers that use that a lot. 

also moved the `enforceTenancy` call out of the base group and only included it where it's needed (Source + Application for now, don't need it on the *Type endpoints)

### DIFF
--- a/routes.go
+++ b/routes.go
@@ -15,6 +15,12 @@ import (
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
+var listMiddleware = []echo.MiddlewareFunc{
+	middleware.SortAndFilter, middleware.Pagination,
+}
+
+var tenancyWithListMiddleware = append([]echo.MiddlewareFunc{enforceTenancy}, listMiddleware...)
+
 func setupRoutes(e *echo.Echo) {
 	e.GET("/openapi.json", func(c echo.Context) error {
 		out, err := redis.Client.Get("openapi").Result()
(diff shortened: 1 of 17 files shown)


lpichler:
Looks good and usage of sort_by param in queries will be added to this PR ? (iif I don't miss something)

lindgrenj6:
@lpichler added the `Order` call to the filtering file as well. Should be ready to review.

routes.go
```diff
@@ -32,50 +38,44 @@ func setupRoutes(e *echo.Echo) {
...
+	v3.GET("/application_authentications", ApplicationAuthenticationList, tenancyWithListMiddleware...)
+	v3.GET("/application_authentications/:id", ApplicationAuthenticationGet, enforceTenancy)
 
-	v3.GET("/app_meta_data", MetaDataList, middleware.ParseFilter, middleware.ParsePagination, withoutTenancy)
-	v3.GET("/app_meta_data/:id", MetaDataGet, withoutTenancy)
-	v3.GET("/application_types/:application_type_id/app_meta_data", ApplicationTypeListMetaData, middleware.ParseFilter, middleware.ParsePagination)
+	// AppMetaData
+	v3.GET("/app_meta_data", MetaDataList, listMiddleware...)
+	v3.GET("/app_meta_data/:id", MetaDataGet)
+	v3.GET("/application_types/:application_type_id/app_meta_data", ApplicationTypeListMetaData, listMiddleware...)
```
lpichler, line 73: it needs `tenancyWithListMiddleware` other it ends with error.

lpichler:
@lindgrenj6 

1. I was trying during testing 
```
GET http://localhost:{{port}}/api/sources/v3.1/application_types?sort_by=display_name
```

and from logs I cannot see sql query with "ORDER BY" statement only:

`SELECT * FROM \"application_types\"  LIMIT 100`

@lindgrenj6 is it same in your local env. ?

it looks that [applyFilters](https://github.com/RedHatInsights/sources-api-go/blob/main/dao/application_type_dao.go#L39)
method needs to return pointer of changed query - and it is not enough to just pass pointer as parameter. Probably it should be  enough but or there is something missing me - anyway it doesn't generate `ORDER BY`.

Obviously I am not sure whether also filters which are using `Where` clause (applyFilters) are working - I also don't see `WHERE` clause in generated SQL queries.

2. I also got response with improperly generated links:
```
....
 "links": {
    "first": "/api/sources/v3.1/application_types?sort_by=display_name/?limit=100&offset=0",
    "last": "/api/sources/v3.1/application_types?sort_by=display_name/?limit=100&offset=100"
  }
}

```

routes.go
lindgrenj6: Neither of these types need tenant_id - I set it to `-1` hardcoded and it seems to have worked.

lindgrenj6:
@lpichler I was able to fix the filter/sorting not working by [changing the function signature of the `applyFilters` function](https://github.com/RedHatInsights/sources-api-go/pull/14/commits/bbed6ac5b386f262f3cb440e7c51eaf21ed225db). Apparently when we call `query.Where(...)` it returns a new pointer, it doesn't modify the one in place. Sorting is now working as expected.

I'll go through and look at my link generation and see if I can fix that too while I'm in here - its something to do with the default limit/offset.

lpichler:
thanks for explanation about pointers and changing signature and it works great.

> I'll go through and look at my link generation and see if I can fix that too while I'm in here - its something to do with the default limit/offset.

@lindgrenj6 Are you going to do it in follow-up PR ?  if so I can merge this one.

lindgrenj6:
@lpichler yes I can do it in a follow-up since this PR is getting large as it is already.

PR #18, opened 2021-08-11
Add Authentication Support backed via Vault

lindgrenj6:
Here it is. The big one.

This PR ~switches out~ implements the `Authentication` model endpoints + backend. But instead of pointing at the regular db we're pointing at a hashicorp vault instance! 

---

The hardest part about this was dealing with the fact that vault is kind of a K/V store rather than a traditional RDMS system which results in us needing to list the keys (for a tenant), then find the one that matches the UID passed in. It also gets even _more_ difficult when the polymorphism piece comes in for this model - currently I have the direct parent in the name of the key (so fetching the key is easy) but keeping track of which Source they all belong to will be a less-fun thing that I'll tackle in the future. 

- Currently all of the CRUD operations are supported - without subcollections. I will finish the subcollections in a follow-up PR. 
- There also is no validations on whether or not the subresource exists - which results in some silly silly keys being created - but they can be listed/fetched fine!
- no error handling or switches for turning vault access on/off - want to add those in this too so we can deploy the repo and it won't crash if someone requests authentications before vault is set up. 

Among other things I also added a way to run vault in dev mode locally via `make vault` which runs the vault on localhost:8200 with a password of `root`.


lpichler:
@lindgrenj6 it looks and works great 👍 I found one inconsistency  with rails server - for creation - `resource_id` has to be string for rails but for GO it works only when `resource_id` is integer - despite what is correct I think it should be now same as in rails - so it will not break systems where it is used. (it is also in openapi)


--
NOTE: @lindgrenj6 I was just thinking if would be beneficial to create also DB objects for `authentifications` (without sensitive data which needs to be stored in Vault) and this object would fetch and access sensitive data from Vault. I think it could help with some operations like GET or filtering, sorting and maybe more. Of course it could also brings other problems - I was just curious what do you think about this way. thanks!

lindgrenj6:
@lpichler 
> for creation - resource_id has to be string for rails but for GO it works only when resource_id is integer

Correct, initially I just did `int`s but in Mikel's PR here #27 he is adding the util functions to accept `interface{}` so it mimics the rails app. If this one goes in first I'll do a follow-up once the other gets merged, or the other way around.

>I was just thinking if would be beneficial to create also DB objects for authentifications (without sensitive data which needs to be stored in Vault) and this object would fetch and access sensitive data from Vault

I've been thinking about that too. We could also use Redis as a super-quick store for extra info as well. The problem with storing some Authentications in the db and others in vault is that we then have to store _somewhere_ where they live - which adds a 3rd place for the data, so it just gets messy.

I do think we'll do something along those lines like a cache in the future - it might be needed. At least as far as filters go I don't think anyone is filtering data coming from the authentications endpoint but I could be wrong.

lpichler (approved):
@lindgrenj6 thank you for your answers and I agree

can you please rebase ? then I can merge it 👍 :shipit:

lindgrenj6:
@lpichler rebased and ready to

PR #19, opened 2021-08-26
Add missing record for testAuthentication in main_test.go

lpichler:
it was failing in my local env when I had empty test DB - I am not sure how it is possible that tests were passing before on my machine.


lindgrenj6 (approved):
Ahhh great catch - all because we're switching out the backend for Authentications.

PR #20, opened 2021-08-31
Add subcollections

lpichler:
This PR adds followings subcollections:

- GET `/application_types/:id/app_meta_data`
- GET `/source_types/:id/sources`
- GET `/sources/:id/application_types`
- GET `/application_types/:id/sources`
- GET `/sources/:id/applications`
- GET `/sources/:id/endpoints`

---

## RelationObject
PR also adds file `model/relation_object.go` where are some functions which helps to do SQL queries with using GORM.

It allows to do queries to get (for example) relation `sources` of `source_type` record  (`/source_types/:id/sources`) and also

offer to query (for example) relation `sources` of `application_type` record even when we need to go through `applications` table. 

**So it can do `has_many` relation  and also `has_many` with joining table `many2many` option in gorm tags in fields of structs which represent models.**

```go

import (
)

sourceType, _ := m.NewRelationObject(<primary object>, <tenant id>, DB.Debug())
query := sourceType.HasMany(<related object>, DB.Debug())
...

// => "SELECT * FROM \"sources\" WHERE \"sources\".\"source_type_id\" = 2
```
`<primary object>` - it is object of model from which relations are obtained 
`<tenant id>` - tenant id when it is need from joins - it applicable only for `through` relation
`<related object>` - object of model which specify model of queried relations - name of relation -  
this requires to specify type in tag of model struct field:

- if it is just 1:N - there no need to specify tag to relation:

```go

type SourceType struct {
	Id        int64     `gorm:"primarykey" json:"id"`
        ...
	Sources []Source // <--- THIS IS THE RELATION
}

```

- if it is just M:N

```go

type ApplicationType struct {
	//fields for gorm
	Id        int64     `gorm:"primarykey" json:"id"`
        ...
	Sources      []*Source `gorm:"many2many:applications;"` // <--- THERE ARE MENTIONED TAGS
}
```

### HasMany

`/api/sources/v3.1/source_types/2/sources`

```go

import (
)

sourceType, _ := m.NewRelationObject(m.SourceType{Id: 2}, 1 /*tenantID*/, DB.Debug())
query := sourceType.HasMany(&m.Source{}, DB.Debug())

```

```go

type SourceType struct {
	Id        int64     `gorm:"primarykey" json:"id"`
        ...
	Sources []Source // <--- THIS IS THE RELATION
}

```

generates:
```sql
SELECT * FROM sources WHERE sources.source_type_id = 2 AND sources.tenant_id = 1 
```

### Many2Many (HasMany with through )

`/api/sources/v3.1/source_types/2/sources`

```go

import (
)

appType, _ := m.NewRelationObject(m.ApplicationType{Id: 2}, 1 /* TenantID */, DB.Debug())
query := appType.HasMany(&m.Source{}, DB.Debug())

```

```go

type ApplicationType struct {
	//fields for gorm
	Id        int64     `gorm:"primarykey" json:"id"`
        ...
	Sources      []*Source `gorm:"many2many:applications;"` // <--- THERE ARE MENTIONED TAGS
}
```

generates:

```sql
SELECT DISTINCT sources.last_checked_at, sources.tenant_id, sources.paused_at, sources.id, sources.updated_at, 
sources.name, sources.source_type_id, sources.availability_status, sources.last_available_at, sources.uid, sources.imported, 
sources.source_ref, sources.app_creation_workflow, sources.created_at, sources.version 
FROM sources INNER JOIN applications
ON sources.id = applications.source_id AND applications.tenant_id = 1 WHERE 
applications.application_type_id = 1 AND sources.tenant_id = 1

```

--- 
### Links
- partly https://issues.redhat.com/browse/RHCLOUD-15000
   - in other PR would be implemented:
      -  GET /applications/:id/authentications
      - GET /endpoints/:id/authentications
       - GET /sources/:id/authentications
      
- point 7- https://issues.redhat.com/browse/RHCLOUD-15682


routes.go
```diff
@@ -40,6 +40,9 @@ func setupRoutes(e *echo.Echo) {
 	v3.POST("/sources", SourceCreate)
 	v3.PATCH("/sources/:id", SourceEdit)
 	v3.DELETE("/sources/:id", SourceDelete)
+	v3.GET("/sources/:id/application_types", ApplicationTypeList, middleware.ParseFilter, middleware.ParsePagination)
+	v3.GET("/sources/:id/applications", ApplicationList, middleware.ParseFilter, middleware.ParsePagination)
+	v3.GET("/sources/:id/endpoints", EndpointList, middleware.ParseFilter, middleware.ParsePagination)
```
lindgrenj6: So for these subcollections - rather than specifying `/primary/:id/subcollection`, we should do `/primary/:primary_id/subcollection` e.g: `/sources/:source_id/applications`.

We should then pass to a specific function like `SourceListApplications` which knows it's looking for `c.PathParam("source_id")` _and_ it knows its a subcollection - that way we don't have if/else logic and path parsing in the handler function, just let echo do the path parsing for us. This keeps the handler functions from getting too large - if we want shared code between them it should be middleware in the stack. 

I _think_ the `RequestURL` struct wouldn't be needed anymore either if we separated them into different handler functions. Since we'd know 1) that the request is for a subcolleciton and 2) what the primary resource is

lpichler: yea this is probably better and straightforward => fixed 👍 thanks!

lindgrenj6 (approved):

PR #21, opened 2021-09-13
Add build_deploy & pr_check.sh

lpichler:
cc @lindgrenj6 

### Links

https://issues.redhat.com/browse/RHCLOUD-15920

### DIFF
--- a/pr_check.sh
+++ b/pr_check.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+# --------------------------------------------
+# Options that must be configured by app owner
+# --------------------------------------------
+APP_NAME="sources"  # name of app-sre "application" folder this component lives in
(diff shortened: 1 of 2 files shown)


pr_check.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
lindgrenj6, line 1: Just need to replace this file with one like this: https://github.com/RedHatInsights/sources-monitor-go/blob/main/pr_check.sh

We don't want to deploy the eph env for this or anything yet just because we don't have proper smoke tests quite yet.

lindgrenj6 (commented):
Just need to replace the file with one like sources-monitor-go has and this is good to go!

lpichler:
@lindgrenj6 done, thanks 👍

lindgrenj6 (approved):

PR #22, opened 2021-09-13
Refactor DAO methods to return int64 instead of *int64

MikelAlejoBR:
Refactors the "list" methods of the DAOs to return values instead of pointers. In the case of having errors, the zero value for the `int64` is returned.

[[RHCLOUD-16095]](https://issues.redhat.com/browse/RHCLOUD-16095)

### DIFF
--- a/application_authentication_handlers.go
+++ b/application_authentication_handlers.go
@@ -52,7 +52,7 @@ func ApplicationAuthenticationList(c echo.Context) error {
 		out[i] = *a.ToResponse()
 	}
 
-	return c.JSON(http.StatusOK, util.CollectionResponse(out, c.Path(), int(*count), limit, offset))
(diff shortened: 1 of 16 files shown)


lindgrenj6:
This looks great - I was just thinking about it some more during review and since we're just using `int` for limit/offset we can probably do that with the count too. 

If we do approach 3.4B of any model we can just go through and change int -> int64 all at once instead of just in certain spaces.

MikelAlejoBR:
In this case [`gorm` forces us to pass an `*int64` variable to the `.Count()` function](https://github.com/go-gorm/gorm/blob/v1.21.11/finisher_api.go#L368), so I don't think we can return an `int` unless we convert the `int64` to an `int`... But then we would have to handle the potential truncates and so on.

lindgrenj6 (approved):
Schweeet! Thanks for getting it done so quick. Yay pointers!

PR #24, opened 2021-09-22
Add Cloud Watch Log Hook

lpichler:
1. Change way of configuration for middleware logger
2. Added method `AddHooksTo(logger *logrus.Logger, config *appconf.SourcesApiConfi)`  which adds  _log hooks_  to `logger`
   A.  method `AddHooksTo` is used for all logging instances:  general logger, echo logger, gorm logger and middleware logger(now this logger tied to general logger)

## Links

https://issues.redhat.com/browse/RHCLOUD-15683


logger/log_writter.go
```diff
@@ -0,0 +1,41 @@
+package logger
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/sirupsen/logrus"
+)
+
+type LogWritter struct {
```
lindgrenj6: Can we rename this to `LogWriter`? I think thats what you were going for unless I'm missing something (filename too).

lindgrenj6 (commented):
Just one small typo I think in the struct name - but looks good to me otherwise!

logger/log_writter.go
lpichler: sure, I fixed it 👍

logger/log_writter.go
```diff
@@ -0,0 +1,41 @@
+package logger
```
lindgrenj6: I hate to be that guy but the filename is still `log_writter.go` :grimacing:

lindgrenj6: I edited it in github since apparently I can do that. Merging....

lindgrenj6 (approved):

logger/log_writter.go
lpichler: oh I am sorry I thought that the file had correct name 🤦 thanks @lindgrenj6 !

PR #27, opened 2021-09-27
feature: SourceCreateRequest validation

MikelAlejoBR:
Validates the "SourceCreateRequest" struct's fields following the guidelines set by the [legacy Ruby code](https://github.com/RedHatInsights/sources-api/blob/d1d7d5dfefab699fe30853f82ec0a3d401ad27dd/app/models/source.rb#L8-L24).

[[RHCLOUD-16353]](https://issues.redhat.com/browse/RHCLOUD-16353)


MikelAlejoBR:
Force pushing to solve some of the `fmt` and `golangci-lint` issues. Nonetheless @lindgrenj6, I don't really understand what "File is not `gci`-ed is. Apparently it takes care of the [import order](https://github.com/daixiang0/gci), but the imports seem to be in order? I must be missing something here. Can you help me with this, please?

Thanks in advance.

lindgrenj6:
Ahhh yeah - so basically that is just making sure the import order is the same in all files so the compiler can compile faster. 

I have included a makefile directive to autofix gci, just type `make gci` and it should auto-fix it for you.

MikelAlejoBR:
Thanks Jacob.

I've just learned something about the "echo" framework. It seems that if you specify `interface{}` as the type of the member of the struct, you'll get a `float64`. Here you can see the variable I declared for the test, and the unmarshalled content:

sourceTypeInt

In the screenshot below, the debugger tells me that the variable is an `interface{} | float64`, and the type switch confirms it.

sourceTypeFloat

If I pass a `*string` it gets unmarshalled to `string` as well. If I change the type to an `*int64` or a `*string` I get the expected types, but not otherwise. I'm thinking on maybe considering those two cases as well, and extracting the type switch to a `util`s function to make it easier to follow.

What do you think?

lindgrenj6:
Yeah - that sounds good to me. make a function `util.InterfaceToInt(i interface{}) (int64, error)` that handles all the types that make sense (ints, floats, strings, thats about it i think). We'll be able to use that in all the create functions I think. 

---

One tiny thing while I'm looking at this, I was more implying that we should have some util functions like `util.SliceContainsString(slice, str) bool`, and that the source constants (app_creation_workflow, etc) would live in the model package like:
```
const (
	AccountAuthorization = "account_authorization"
	ManualConfiguration = "manual_configuration"
)
var SourceAppCreationWorkflow = []string{AccountAuthorization, ManualConfiguration}
```

Then we'd call it like `util.SliceContainsString(SourceAppCreationWorkflow, req.AppCreationWorkflow)` from the validation function(s).

This way the constants are generally contained in one package (model) rather than across multiple.

MikelAlejoBR:
All right, will do! Thanks!

MikelAlejoBR:
I'm removing the WIP status because I think it is ready to be reviewed!

model/source_http.go
```diff
@@ -14,7 +21,48 @@ type SourceCreateRequest struct {
...
+	}
+
+	// If no valid UUID is provided in the request, we generate one
+	if req.Uid == nil || *req.Uid == "" {
+		id := uuid.New()
+		stringId := id.String()
+		req.Uid = &stringId
+	}
+
+	var validWorkflowStatuses = []string{AccountAuth, ManualConfig}
```
lindgrenj6: We can make this a `var` at the package level - that way it is only alloc'd once rather than every time we're validating something (wish we could make it a const - but we can't do constant slices/arrays :sob:)

Same with the availability status array - but throw that (+ the consts above) in a file called `availability_status.go`, since that will be shared code between basically all of the data models.

util/slice_utils.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import "sort"
+
+// IsStringPresentInSlice returns true if the specified target is present in the given slice. Requires the slice to
+// be ordered in ascending order, due to the underlying requirement with SearchStrings.
+func IsStringPresentInSlice(target string, slice []string) bool {
```
lindgrenj6: For this function signature - can we reverse the order of the params (and probably rename to the opposite too)? Main reason for asking is thats how `strings.HasPrefix(s, prefix string)` accepts parameters plus how [many other APIs](https://pkg.go.dev/strings@go1.17.1) in Go are structured. It would be good to continue that here.

util/slice_utils.go
```diff
@@ -0,0 +1,17 @@
...
+	index := sort.SearchStrings(slice, target)
+
+	return index < len(slice) && slice[index] == target
```
lindgrenj6: I'm not a huge fan of the `sort` package specifically for this design (it doesn't return false, it returns an index which we then have to check). 

It does look like it does a binary search which theoretically should be faster: https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/sort/search.go;drc=refs%2Ftags%2Fgo1.17;l=59

But the fact that we have to sort the array before searching negates any benefits imo.

Can we change this to just a for loop looping through the array? That way if the item isn't found we just return false rather than having to check the index and the value at said index.

util/parser.go
```diff
@@ -0,0 +1,48 @@
+package util
+
+import (
+	"fmt"
+	"strconv"
+)
+
+// InterfaceToInt64 takes an interface and returns an int64 for any float64, int64 or string received -whether they
+// are pointers or not-.
+func InterfaceToInt64(i interface{}) (int64, error) {
```
lindgrenj6, line 10: This is great, exactly what I was looking for!

lindgrenj6 (commented):
Awesome first pass - this is going to be a great pattern established for the entire app.

Great job on the tests!

model/source_http.go
MikelAlejoBR: All right. I've left `validWorkflowStatuses` slice on the `source_http.go`, because after doing a quick lookup I haven't found any more usages of `AppCreationWorkflow` on the repository. I'm thinking that maybe I should move the constants of that array to the `source_http.go` file, since I feel it makes more sense to have everything in the same file, instead of having the constants on `source.go` and then the slice in `source_http.go`. What do you think?

util/slice_utils.go
MikelAlejoBR: I knew this was going to come back at me. I felt SO weird writing that function like that... like after doing everything I was "pff, this feels odd". Should have changed it before.

util/slice_utils.go
MikelAlejoBR: Yeah, I was really unsure about using `sort` too because of that "needs the slice to be sorted" requirement... All right, changed the implementation to a simple for loop.

util/parser.go
MikelAlejoBR: Great :tada:

model/source_http.go
```diff
@@ -14,7 +25,46 @@ type SourceCreateRequest struct {
...
+
+// Validate validates that the required fields of the SourceCreateRequest request hold proper values. In the specific
+// case of the UUID, if an empty or nil one is provided, a new random UUID is generated and appended to the request.
+func (req *SourceCreateRequest) Validate() error {
+	if req.Name == nil || *req.Name == "" {
+		return fmt.Errorf("name cannot be empty")
+	}
+
+	// If no valid UUID is provided in the request, we generate one
+	if req.Uid == nil || *req.Uid == "" {
```
MikelAlejoBR: Here the code doesn't check that the given UUID is well-formatted UUID\[1\]. I was thinking on adding a regex to check that, what do you think @lindgrenj6 ?

\[1\]: https://en.wikipedia.org/wiki/Universally_unique_identifier#Format

model/source_http.go
```diff
@@ -14,7 +25,46 @@ type SourceCreateRequest struct {
...
 
-	SourceTypeID *int64 `json:"source_type_id"`
+	SourceTypeID    *int64      `json:"-"`
+	SourceTypeIDRaw interface{} `json:"source_type_id"`
+}
+
+// Validate validates that the required fields of the SourceCreateRequest request hold proper values. In the specific
+// case of the UUID, if an empty or nil one is provided, a new random UUID is generated and appended to the request.
+func (req *SourceCreateRequest) Validate() error {
+	if req.Name == nil || *req.Name == "" {
```
lpichler: @MikelAlejoBR why it requires both checks ? isn't enough just e.g. `req.Name == nil` ?

model/source_http.go
```diff
@@ -14,7 +25,46 @@ type SourceCreateRequest struct {
...
+	}
+
+	// If no valid UUID is provided in the request, we generate one
+	if req.Uid == nil || *req.Uid == "" {
+		id := uuid.New()
+		stringId := id.String()
+		req.Uid = &stringId
+	}
+
+	if !util.SliceContainsString(validWorkflowStatuses, req.AppCreationWorkflow) {
```
lpichler: in ruby version of sources api is also valid when `app_creation_workflow` is not specified - it defaults to `manual_configuration`

lpichler:
@MikelAlejoBR I was not able to check creation:

```
POST http://localhost:{{port}}/api/sources/v3.1/sources

{
  "name": "Test3aaa",
  "source_type_id": "8",
  "app_creation_workflow": "manual_configuration"
}
```

Response:
```
{
  "errors": [
    {
      "detail": "Internal Server Error: ERROR: insert or update on table \"sources\" violates foreign key constraint \"fk_rails_f830a376e4\" (SQLSTATE 23503)",
      "status": "500"
    }
  ]
}
```

it is probably because of missing `tenant_id` which should be determined from `getSourceDao(c)` - so source have to be created with 	attribute `TenantID: *sourcesDB.Tenant()` - not sure if this is subject of this PR 
I added it:

```
diff --git a/source_handlers.go b/source_handlers.go
index f3c6f30..0896be4 100644
--- a/source_handlers.go
+++ b/source_handlers.go
@@ -190,6 +190,7 @@ func SourceCreate(c echo.Context) error {
                AvailabilityStatus: m.AvailabilityStatus{
                        AvailabilityStatus: input.AvailabilityStatus,
                },
+               TenantID: *sourcesDB.Tenant(),
                SourceTypeID: *input.SourceTypeID,
        }

```

^^^ @lindgrenj6 hopefully it is correct fix.


@MikelAlejoBR - 

I was able to add sources with same name despite [this code](https://github.com/RedHatInsights/sources-api/blob/d1d7d5dfefab699fe30853f82ec0a3d401ad27dd/app/models/source.rb#L23) - but it should be possible to add source with same name but under different tenant (https://github.com/RedHatInsights/sources-api/blob/d1d7d5dfefab699fe30853f82ec0a3d401ad27dd/app/models/source.rb#L24).

so it looks that It needs to be fixed - thanks!

model/source_http.go
MikelAlejoBR: [You are right](https://github.com/RedHatInsights/sources-api/blob/5befbc497719db35758488535a44afcb5f82000b/db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb#L3). Thank you for pointing this out, I'll default this to `manual_configuration`.

model/source_http.go
MikelAlejoBR: The Rails app actually had an `:allow_blank => false` for the `:name` attribute, so that's why we're performing both checks: [models/source.rb#L23](https://github.com/RedHatInsights/sources-api/blob/5befbc497719db35758488535a44afcb5f82000b/app/models/source.rb#L22)

model/source_http.go
MikelAlejoBR: [You are right](https://github.com/RedHatInsights/sources-api/blob/5befbc497719db35758488535a44afcb5f82000b/db/migrate/20210125173413_add_app_creation_workflow_to_sources.rb#L3). Thanks for pointing this out, I'll default the parameter to `manual_configuration` if it comes empty / nil.

lindgrenj6:
@lpichler 
>it is probably because of missing tenant_id which should be determined from getSourceDao(c)

Ahhh yeah it looks like I forgot to set that there - we should set it in the `SourceDao.Create` function though - that way it always gets set even if we use it outside the handler.

model/source_http.go
MikelAlejoBR: My bad, the UUID should be automatically generated[1], because the spec doesn't say anything about accepting an `uuid` key in the request body.

\[1\]: https://github.com/RedHatInsights/sources-api/blob/d1d7d5dfefab699fe30853f82ec0a3d401ad27dd/app/models/source.rb#L8

MikelAlejoBR:
Rebased to include the latest upstream changes

MikelAlejoBR:
I just came up with a little problem regarding circular imports. I was thinking on injecting the `SourceDao` in the `model/source_http.go#Validate` function like this:

```go
func (req *SourceCreateRequest) Validate(sourceDao dao.SourceDao)
```

The problem is that I cannot import the "dao" package, because then this package imports "model" back, and thus creates a circular imports issue. Maybe I should extract the `Validate` function to a `validation` package, and create a `validation/source.go` service there?

MikelAlejoBR:
I'm sorry about the noise. I had to rebase again to include the changes of [pull/35](https://github.com/RedHatInsights/sources-api-go/pull/35), as I was having some issues with the integration tests. Basically the latest commits add what @lpichler mentioned: the source's name gets validated against the database, to check whether it is unique in the tenant or not.

MikelAlejoBR:
Rebased main to include the latest `testutils` package's utility functions

dao/source_dao.go
```diff
@@ -89,3 +90,11 @@ func (s *SourceDaoImpl) Delete(id *int64) error {
 func (s *SourceDaoImpl) Tenant() *int64 {
 	return s.TenantID
 }
+
+func (s *SourceDaoImpl) NameExistsInCurrentTenant(name string) bool {
+	src := &m.Source{Name: name}
+	result := DB.Where("name LIKE ? AND tenant_id = ?", name, s.TenantID).First(src)
```
lpichler: Can we have "=" instead of LIKE here ?

MikelAlejoBR: Good idea. Done!

lpichler (approved):
LGTM 👍 thanks!

PR #29, opened 2021-10-04
Change CollectionResponse to generate links based on query parameters

lindgrenj6:
A few changes in this PR (probably best to review by commit):

1. Fixing the links as mentioned in #14 but wanted to do a follow-up.
2. Went through and changed out the `SubCollectionList` count return type from *int64 -> int64. Mikel's PR changing the return types went in while this was in review.


lpichler (approved):
it works great 👍 thanks!

PR #32, opened 2021-10-05
Improve logging for failing requests

lpichler:
This adds logging for error messages - it improves to determine sources of issues.

it changes
```
util.ErrorDoc("Bad Request", "400")
```
to 

```
errorLog := util.ErrorLog{Logger: c.Logger(), LogMessage: err.Error()
errorLog.ErrorDocument()
```

`ErrorLog` can be initialised with status, message (which is returned in API response) and LogMessage(which is visible in logs).
if `LogMessage` is empty it takes message. There are also some defaults values in `(e *ErrorLog) ErrorDocument()`

### DIFF
--- a/application_authentication_handlers.go
+++ b/application_authentication_handlers.go
@@ -43,8 +43,10 @@ func ApplicationAuthenticationList(c echo.Context) error {
 
 	applications, count, err := applicationDB.List(limit, offset, filters)
 	if err != nil {
-		return c.JSON(http.StatusBadRequest, util.ErrorDoc("Bad Request", "400"))
+		errorLog := util.ErrorLog{Logger: c.Logger(), LogMessage: err.Error()}
+		return c.JSON(http.StatusBadRequest, errorLog.ErrorDocument())
 	}
+
 	c.Logger().Infof("tenant: %v", *applicationDB.Tenant())
 
 	out := make([]interface{}, len(applications))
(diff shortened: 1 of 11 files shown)


lpichler:
@lindgrenj6 ~~I am not sure if I did anything but there is only one check in this PR. Do you have idea where are checks for test, fmt and `golangci-lint` ?~~ 

EDIT: they appeared.

lindgrenj6 (changes requested):
I see what you're trying to do here - but luckily we don't need to pass around the logger instance since it is living at the package level of the `logger` package.

This implements the same functionality without having to have a separate struct - since the desired side-effect is that each error is logged: 
```diff
diff --git a/util/error_document.go b/util/error_document.go
index 903d612..3bf409b 100644
--- a/util/error_document.go
+++ b/util/error_document.go
@@ -1,5 +1,7 @@
 package util
 
+import l "github.com/RedHatInsights/sources-api-go/logger"
+
 type Error struct {
 	Detail string `json:"detail"`
 	Status string `json:"status"`
@@ -9,6 +11,8 @@ type ErrorDocument struct {
 }
 
 func ErrorDoc(message, status string) *ErrorDocument {
+	l.Log.Error(message)
+
 	return &ErrorDocument{
 		[]Error{{
 			Detail: message,
```

Maybe we can change the first arg to `err.Error()` everywhere that was just "Bad Request" before as well. Thats just leftover from when I initially hacked this all together.

lpichler:
this is replaced by https://github.com/RedHatInsights/sources-api-go/pull/33

PR #33, opened 2021-10-06
Improve logging for failing requests

lpichler:
1. this changes message "Bad Request" to `err.Error()`
2. this adds log error messages in `ErrorDoc` function


lindgrenj6 (approved):

lindgrenj6:
Hm just one failure - I think we need to init the logger for the tests.

lpichler:
> Hm just one failure - I think we need to init the logger for the tests.

yes we need to add `init` only for tests 👍 now it is fixed otherwise it works well.

lindgrenj6 (approved):

PR #37, opened 2021-10-06
Update handling of errors in handlers to return 400/404 instead of 500

lindgrenj6:
Inspired by @lpichler 's recent changes regarding error handling and since the RDU2 datacenter is down - I went through and updated some of the handlers to return 400 when failing to parse the ID and a 404 if we fail to find specific records. 

cc @MikelAlejoBR you're not showing up in the list so i can't add you as a reviewer - but have a looksee.


lpichler (approved):

MikelAlejoBR (approved):
Looks good to me :+1:

PR #39, opened 2021-10-13
Refactor: extract common test code to a testutils package

MikelAlejoBR:
The reason for this change is that otherwise these functions would have to be rewritten over and over again for each package. The idea is to place the code under the `internal/` directory so it doesn't get exported —which wouldn't make sense anyway as these settings apply specifically to this project—.

Basically:

- It self-contains the errors that can be thrown from the database functions. We were only doing an `if err != nil { log & exit }` for the errors, which made the callers have to do that when there's usually no other recovery than manually checking what's up with the database.
- It extracts the test data or fixtures to a file and exports them, to make importing from other packages easier.
- Replaces some common request and context set up with a helper function.

### Discussions
1. I replaced the `Fprintf & Exit 1` calls with `log.Fprintf` calls, which combine the former calls in one. The output ends up being formatted as JSON, as the rest of the logs of the APP. I guess that's why you were using the two calls, but anyway I'd like to ask: should I keep the two calls?
2. I initially thought of the `ParseFlags` function as one function that would have the following signature: `func ParseFlags() bool`. It'd directly create the database if `-createdb`is passed, and return a `bool` indicating whether the tests should include the integration ones. That would spare us from having an `if createDb` on every test package. The only "problem" is that I don't like having a function called `ParseFlags` which also takes care of initializing the database. Any suggestions around this? :thinking: 

[[RHCLOUD-16574]](https://issues.redhat.com/browse/RHCLOUD-16574)

### DIFF
--- a/internal/testutils/parser.go
+++ b/internal/testutils/parser.go
@@ -1,0 +1,15 @@
+package testutils
+
+import "flag"
+
+// ParseFlags parses the flags for the "go test" command. The "-createdb" indicates that a test database should be
+// created, whilst the "-integration" flag indicates that integration tests should be run along with the unit tests.
+// It returns a (bool, bool) which represents the status of ("-createdb", "-integration") accordingly.
+func ParseFlags() (bool, bool) {
+	createDb := flag.Bool("createdb", false, "create the test database")
+	integration := flag.Bool("integration", false, "run unit or integration tests")
+
+	flag.Parse()
+
+	return *createDb, *integration
+}
--- a/internal/testutils/request.go
+++ b/internal/testutils/request.go
@@ -1,0 +1,24 @@
+package testutils
+
+import (
+	"io"
+	"net/http/httptest"
+
+	"github.com/labstack/echo/v4"
+)
+
+var echoInstance = echo.New()
+
+// CreateTestContext sets up a new echo context with the parameters given, and returns the context itself and the
+// response recorder.
+func CreateTestContext(method string, path string, body io.Reader, context map[string]interface{}) (echo.Context, *httptest.ResponseRecorder) {
+	request := httptest.NewRequest(method, path, body)
+	recorder := httptest.NewRecorder()
+	echoContext := echoInstance.NewContext(request, recorder)
+
+	for k, v := range context {
+		echoContext.Set(k, v)
+	}
+
+	return echoContext, recorder
+}
--- a/main_test.go
+++ b/main_test.go
@@ -8,7 +8,6 @@ import (
 	"github.com/RedHatInsights/sources-api-go/dao"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils"
 	l "github.com/RedHatInsights/sources-api-go/logger"
-	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
 )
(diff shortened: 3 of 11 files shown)


MikelAlejoBR:
Hello @lpichler and @lindgrenj6.

I think this is ready for review, but I can't assign you as reviewers for some reason :/

Thanks in advance!

internal/testutils/parser.go
```diff
@@ -0,0 +1,15 @@
+package testutils
+
+import "flag"
+
+// ParseFlags parses the flags for the "go test" command. The "-createdb" indicates that a test database should be
+// created, whilst the "-integration" flag indicates that integration tests should be run along with the unit tests.
+// It returns a (bool, bool) which represents the status of ("-createdb", "-integration") accordingly.
+func ParseFlags() (bool, bool) {
```
lindgrenj6, line 8: What would you think about changing this to return a `map[string]string` (or `map[string]bool`)?

That way if we were to add more flags (like vault potentially) they could all just be returned in the map from this function. Just a thought.

EDIT: maybe we wouldn't need the createDb flag anymore - but we could potentially add some more in the future so maybe the point is still valid? not really. lol

internal/testutils/request.go
```diff
@@ -0,0 +1,24 @@
...
+	"net/http/httptest"
+
+	"github.com/labstack/echo/v4"
+)
+
+var echoInstance = echo.New()
+
+// CreateTestContext sets up a new echo context with the parameters given, and returns the context itself and the
+// response recorder.
+func CreateTestContext(method string, path string, body io.Reader, context map[string]interface{}) (echo.Context, *httptest.ResponseRecorder) {
```
lindgrenj6, line 14: :tada: this is so nice!

main_test.go
```diff
@@ -1,76 +1,50 @@
...
 	mockApplicationTypeDao dao.ApplicationTypeDao
 	mockEndpointDao        dao.EndpointDao
 	mockSourceTypeDao      dao.SourceTypeDao
 	mockApplicationDao     dao.ApplicationDao
 	mockMetaDataDao        dao.MetaDataDao
-
-	testDbName = "sources_api_test_go"
 )
 
 func TestMain(t *testing.M) {
```
lindgrenj6, line 24: 10/10 would have you refactor again. ALL THE RED.

lindgrenj6 (commented):
> I replaced the Fprintf & Exit 1 calls with log.Fprintf calls, which combine the former calls in one. The output ends up being formatted as JSON, as the rest of the logs of the APP. I guess that's why you were using the two calls, but anyway I'd like to ask: should I keep the two calls?

That looks good to me - I always forget about `log.Fatal` because we really don't use it much. 

> I initially thought of the ParseFlags function as one function that would have the following signature: func ParseFlags() bool. It'd directly create the database if -createdb is passed, and return a bool indicating whether the tests should include the integration ones. That would spare us from having an if createDb on every test package. The only "problem" is that I don't like having a function called ParseFlags which also takes care of initializing the database. Any suggestions around this? thinking

Hmm. Yeah that is a good question. Personally I like specifically passing `-createdb` mostly just because it isn't touching the db without you explicitly telling it to (same behavior in rails). On the other hand it would be convenient to have it just _try_ to create the db when running tests (since it needs it anyway) and erroring if it fails. 

I think I want it to look at the `-integration` flag and check if the db exists and then create it if it's missing. Just because rails requires a `db:setup` doesn't mean we need to follow that here.

main_test.go
MikelAlejoBR: 🟢🟢
🔴🔴🔴🔴🔴

:+1:

internal/testutils/parser.go
MikelAlejoBR: I see your `map` and raise you a `struct`. That way we can simplify checking which flag was passed by using Go's type system: instead of doing `if flags["createDb"] {...}` we can simply do `if flags.CreateDb {...}`. We can also avoid typos this way as well —not that they'll be a big deal for when we're testing but...—. What do you think?

Regarding removing the `createDb` flag... I'm not sure about it. I mean, we could automatically create the DB if you passed the `integration` flag, but then we should handle the potential error of an already existing database. Also, if it already exists, should we also truncate the stuff or leave it there? If we don't, we would have to skip reinserting the data... and so on and so forth. In my opinion it seems like a lot of work and not much gain, but we could open up a discussion about it :smiley::+1:

lindgrenj6: Yeah lets just leave it for now. It isn't hurting anything being explicit like that. We can revisit if there is a good reason to remove it.

lindgrenj6 (approved):

PR #41, opened 2021-10-15
Add Kafka package

lpichler:
Add Kafka package with possibility to receive and send message via Kafka.

This Kafka package uses https://github.com/segmentio/kafka-go and provider provider uniform api to producing and consuming messages.

### Produce Messages
```go

// init manager
producerConfig := kafka.ProducerConfig{Topic: "platform.sources.status"}
kafkaConfig := kafka.Config{KafkaBrokers: []string{"localhost:9092"}, ProducerConfig: producerConfig}
kf := &kafka.Manager{Config: kafkaConfig}

// prepare message to send

// we will put on Kafka object of this struct
type StatusMessage struct {
  ResourceType string `json:"resource_type"`
  ResourceID   string `json:"resource_id"`
  Status       string `json:"status"`
  Error        string `json:"error"`
} 

// prepare message to send
m := &kafka.Message{}

// add headers
headers := []kafka.Header{
  {Key: "event_type", Value: []byte("Source.update")},
  {Key: "encoding", Value: []byte("json")},
}
m.AddHeaders(headers)

// add values from struct instance to message
statusMessage := StatusMessage{ResourceID: "3", ResourceType: "Source", Status: "available"}
m.AddValueAsJSON(statusMessage)

// put message on Kafka queue
kf.Produce(m)

```

### Consume Messages
```go
type StatusMessage struct {
  ResourceType string `json:"resource_type"`
  ResourceID   string `json:"resource_id"`
  Status       string `json:"status"`
  Error        string `json:"error"`
} 

consumerConfig := kafka.ConsumerConfig{Topic: "platform.sources.status", GroupID: "sources-api-status-worker"}
kafkaConfig := kafka.Config{KafkaBrokers: []string{"localhost:9092"}, ConsumerConfig: consumerConfig}

kf := &kafka.Manager{Config: kafkaConfig}

kf.Consume(func(message kafka.Message) {
  var m StatusMessage
  err := message.ParseTo(&m)
  If err != nil {
	return
  }

  fmt.Println(m)
})

```

### Both
```go
consumerConfig := kafka.ConsumerConfig{Topic: "platform.sources.status", GroupID: "sources-api-status-worker"}
producerConfig := kafka.ProducerConfig{Topic: "platform.sources.status"}
kafkaConfig := kafka.Config{KafkaBrokers: []string{"localhost:9092"}, ProducerConfig: producerConfig, ConsumerConfig: consumerConfig}

kf := &kafka.Manager{Config: kafkaConfig}
...
```

## Types

`Manager` is struct has methods to init **Kafka Writter** and **Reader** https://github.com/segmentio/kafka-go from and method to producer and consume messages. This struct also holds necessary **config** information.

`Config` - struct has attribute for kafka broker addresses and configs for topic and consumer group: `ConsumerConfig` and `ProducerConfig`

`Header` and `Message` are alias for `kafka.Header` and `kafka.Message`.

**Usage needs only types from this Kafka package it doesn't require import of https://github.com/segmentio/kafka-go**

### DIFF
--- a/kafka/manager.go
+++ b/kafka/manager.go
@@ -2,21 +2,28 @@ package kafka
 
 import (
 	"context"
+	"fmt"
+
 	"github.com/segmentio/kafka-go"
 )
 
-func (manager *Manager) Produce(message *Message) {
+func (manager *Manager) Produce(message *Message) error {
 	if manager.Producer() == nil {
-		return
+		return fmt.Errorf("producer is not initialized")
 	}
 
 	if !message.isEmpty() {
-		manager.Producer().WriteMessages(context.Background(),
+		err := manager.Producer().WriteMessages(context.Background(),
 			kafka.Message{
 				Headers: message.Headers,
 				Value:   message.Value,
 			})
+		if err != nil {
+			return err
+		}
 	}
+
+	return nil
 }
 
 func (manager *Manager) Producer() *kafka.Writer {
@@ -29,16 +36,16 @@ func (manager *Manager) Producer() *kafka.Writer {
 	}
 
 	manager.producer = &kafka.Writer{
-		Addr:  kafka.TCP(manager.Config.KafkaBrokers[0]),
+		Addr:  kafka.TCP(manager.Config.KafkaBrokers...),
 		Topic: manager.Config.ProducerConfig.Topic,
 	}
 
 	return manager.producer
 }
 
-func (manager *Manager) Consume(consumerHandler func(Message)) {
+func (manager *Manager) Consume(consumerHandler func(Message)) error {
 	if manager.Consumer() == nil {
-		return
+		return fmt.Errorf("consumer is not initialized")
 	}
 
 	for {
--- a/kafka/manager.go
+++ b/kafka/manager.go
@@ -1,0 +1,65 @@
+package kafka
+
+import (
+	"context"
+	"github.com/segmentio/kafka-go"
+)
+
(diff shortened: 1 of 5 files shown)


kafka/manager.go
```diff
@@ -0,0 +1,72 @@
...
+	if manager.producer != nil {
+		return manager.producer
+	}
+
+	if len(manager.Config.KafkaBrokers) == 0 {
+		return nil
+	}
+
+	manager.producer = &kafka.Writer{
+		Addr:  kafka.TCP(manager.Config.KafkaBrokers[0]),
```
lindgrenj6: So [looking at the API](https://github.com/segmentio/kafka-go/blob/d8dd9b00cf48af7abc5144bbc574ccf03c3199a0/address.go#L9) we can pass in all the brokers, like so:
```
Addr:  kafka.TCP(manager.Config.KafkaBrokers...),
```

That way the writer will have knowledge of all the brokers when there are multiple.

lindgrenj6 (commented):
Just one tiny change on the producer declaration - other than that this looks great! :muscle: 

EDIT: also I will be moving the superkey implementation over to this when I get time, importing it like so: `github.com/RedHatInsights/sources-api-go/kafka`

kafka/manager.go
lpichler: done, thanks!

lindgrenj6 (approved):

PR #43, opened 2021-10-27
Endpoint validation

MikelAlejoBR:
Implements the equivalent endpoint validation of [sources-api/app/models/endpoint.rb](https://github.com/RedHatInsights/sources-api/blob/master/app/models/endpoint.rb) plus some extra validations. The validations that have been implemented are the following ones:

## Validations and defaults
- `SourceID`
    - Required, non-zero and positive number.
- `Default`
    - An endpoint can't be set as "default" if there's another endpoint for the given source marked as "default".
    - Marked as default if there aren't any other endpoints for the given source.
- `Role`
    - There must not exist another endpoint for the given source with the same role name.
- `Scheme`
    - Not required.
    - If it is `nil`, or it is not a valid scheme, it defaults to `https`
-  `Host`
    - Not required.
    - Can't be longer than 255 chars (RFC2181). Can't have labels longer than 63 chars (RFC2181).
    - Must be a valid host (RFC1034 and RFC2181)
- `Port`
    - Not required. If provided, must be lower than MAXPORT.
    - Defaults to `443`
- `VerifySsl`
    - Not required.
    - Defaults to `true`
- `CertificateAuthority`
    - Not required.
    - Must not be null if `VerifySsl` is set as `true`.
- `AvailabilityStatus`
    - Required.
    - Must be one of `("", "available", "unavailable")`.

[[RHCLOUD-16766]](https://issues.redhat.com/browse/RHCLOUD-16766)

### DIFF
--- a/dao/endpoint_dao.go
+++ b/dao/endpoint_dao.go
@@ -84,7 +84,7 @@ func (a *EndpointDaoImpl) Tenant() *int64 {
 func (a *EndpointDaoImpl) CanEndpointBeSetAsDefaultForSource(sourceId int64) bool {
 	endpoint := &m.Endpoint{}
 	// add double quotes to the "default" column to avoid any clashes with postgres' "default" keyword
-	result := DB.Where("\"default\" = true AND source_id = ?", sourceId).First(&endpoint)
+	result := DB.Where(`"default" = true AND source_id = ?`, sourceId).First(&endpoint)
 
 	return result.Error == nil
 }
--- a/dao/endpoint_dao.go
+++ b/dao/endpoint_dao.go
@@ -80,3 +80,27 @@ func (a *EndpointDaoImpl) Delete(id *int64) error {
 func (a *EndpointDaoImpl) Tenant() *int64 {
 	return a.TenantID
 }
+
+func (a *EndpointDaoImpl) CanEndpointBeSetAsDefaultForSource(sourceId int64) bool {
+	endpoint := &m.Endpoint{}
+	// add double quotes to the "default" column to avoid any clashes with postgres' "default" keyword
+	result := DB.Where("\"default\" = true AND source_id = ?", sourceId).First(&endpoint)
+
+	return result.Error == nil
+}
+
+func (a *EndpointDaoImpl) IsRoleUniqueForSource(role string, sourceId int64) bool {
+	endpoint := &m.Endpoint{}
+	result := DB.Where("role = ? AND source_id = ?", role, sourceId).First(&endpoint)
+
+	// If the record doesn't exist "result.Error" will have a "record not found" error
+	return result.Error != nil
+}
+
+func (a *EndpointDaoImpl) SourceHasEndpoints(sourceId int64) bool {
+	endpoint := &m.Endpoint{}
+
+	result := DB.Where("source_id = ?", sourceId).First(&endpoint)
+
+	return result.Error == nil
+}
--- a/model/endpoint_http.go
+++ b/model/endpoint_http.go
@@ -27,15 +27,16 @@ type EndpointResponse struct {
 }
 
 type EndpointCreateRequest struct {
-	Default              *bool   `json:"default"`
-	ReceptorNode         *string `json:"receptor_node"`
-	Role                 *string `json:"role"`
-	Scheme               *string `json:"scheme"`
-	Host                 *string `json:"host"`
-	Port                 *int64  `json:"port"`
-	Path                 string  `json:"path"`
-	VerifySsl            *bool   `json:"verify_ssl"`
-	CertificateAuthority *string `json:"certificate_authority"`
-	AvailabilityStatus   string  `json:"availability_status"`
-	SourceID             *string `json:"source_id"`
+	Default              *bool       `json:"default"`
+	ReceptorNode         *string     `json:"receptor_node"`
+	Role                 *string     `json:"role"`
+	Scheme               *string     `json:"scheme"`
+	Host                 *string     `json:"host"`
+	Port                 *int64      `json:"port"`
+	Path                 string      `json:"path"`
+	VerifySsl            *bool       `json:"verify_ssl"`
+	CertificateAuthority *string     `json:"certificate_authority"`
+	AvailabilityStatus   string      `json:"availability_status"`
+	SourceID             int64       `json:"-"`
+	SourceIDRaw          interface{} `json:"source_id"`
 }
--- a/model/endpoint_http.go
+++ b/model/endpoint_http.go
@@ -27,16 +27,15 @@ type EndpointResponse struct {
 }
 
 type EndpointCreateRequest struct {
-	Default                 *bool   `json:"default"`
-	ReceptorNode            *string `json:"receptor_node"`
-	Role                    *string `json:"role"`
-	Scheme                  *string `json:"scheme"`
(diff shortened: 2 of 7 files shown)


MikelAlejoBR:
@lindgrenj6, I was thinking on leaving this PR as-is instead of adding more stuff to it —spec, routes and handler code—, because I have the feeling that this is already pretty big. What do you think?

Also, another note: does it make sense to have the `AvailabilityStatusError` member on `*Create` models? Because I understand that it makes sense to have it when querying for the object itself, to know why my endpoint or source is unavailable... But on creation?

dao/endpoint_dao.go
```diff
@@ -80,3 +80,27 @@ func (a *EndpointDaoImpl) Delete(id *int64) error {
 func (a *EndpointDaoImpl) Tenant() *int64 {
 	return a.TenantID
 }
+
+func (a *EndpointDaoImpl) CanEndpointBeSetAsDefaultForSource(sourceId int64) bool {
+	endpoint := &m.Endpoint{}
+	// add double quotes to the "default" column to avoid any clashes with postgres' "default" keyword
+	result := DB.Where("\"default\" = true AND source_id = ?", sourceId).First(&endpoint)
```
lindgrenj6: small nit (and it's taking me forever to type this because github markdown overtakes it): 
You can use [string literals](https://golang.org/ref/spec#String_literals) so that you don't have to escape the quotes. Basically just use backticks intstead of quotes.

model/endpoint_http.go
```diff
@@ -25,3 +25,18 @@ type EndpointResponse struct {
...
+	ReceptorNode            *string `json:"receptor_node"`
+	Role                    *string `json:"role"`
+	Scheme                  *string `json:"scheme"`
+	Host                    *string `json:"host"`
+	Port                    *int64  `json:"port"`
+	Path                    string  `json:"path"`
+	VerifySsl               *bool   `json:"verify_ssl"`
+	CertificateAuthority    *string `json:"certificate_authority"`
+	AvailabilityStatus      string  `json:"availability_status"`
+	AvailabilityStatusError string  `json:"availability_status_error"`
```
lindgrenj6: correct - this field doesn't quite make sense for a create request. I know it was in the spec, but the spec was generated...and generated incorrectly!

model/endpoint_http.go
```diff
@@ -25,3 +25,18 @@ type EndpointResponse struct {
...
+	SourceID                *string `json:"source_id"`
```
lindgrenj6: For this field we might want to use the same logic and helper methods we have in the Source creation workflow so we can accept integers AND strings for the foreign key reference.

service/endpoint_validation.go
```diff
@@ -0,0 +1,117 @@
...
+	defaultScheme    = "https"
+	defaultPort      = int64(443)
+	defaultVerifySsl = true
+	maxFqdnLength    = 255 // RFC2181
+	maxLabelLength   = 63  // RFC2181
+	maxPort          = 65535
+)
+
+func ValidateEndpointCreateRequest(dao dao.EndpointDao, ecr *model.EndpointCreateRequest) error {
+	sourceId, err := util.InterfaceToInt64(ecr.SourceID)
```
lindgrenj6: Ok sweet - just need to change the field to `interface{}` in the create request and do the same `*Raw` pattern.

service/endpoint_validation.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return fmt.Errorf("the provided source ID is not valid")
+	}
+
+	if sourceId < 1 {
+		return fmt.Errorf("invalid source id")
+	}
+
+	if ecr.Default == nil {
```
lindgrenj6: On the rails side we are defaulting to true if it isn't present, here: https://github.com/RedHatInsights/sources-api/blob/dca0a0fd9d829e2765389aa418348b99719948b8/app/models/endpoint.rb#L16

It looks like you're already checking for multiple defaults and setting to true if not - so we just need to make sure this field isn't required on create. If the spec said it was required its probably wrong for create requests - its just always there on response because we set a default.

service/endpoint_validation.go
```diff
@@ -0,0 +1,117 @@
...
+		return fmt.Errorf("a default endpoint already exists for the provided source")
+	}
+
+	if !dao.SourceHasEndpoints(sourceId) {
+		*ecr.Default = true
+	}
+
+	if ecr.ReceptorNode == nil || *ecr.ReceptorNode == "" {
+		return fmt.Errorf("the receptor node cannot be empty")
+	}
```
lindgrenj6: Receptor node is not required - spec is probably wrong for create requests.

service/endpoint_validation.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if ecr.Role == nil || *ecr.Role == "" {
+		return fmt.Errorf("role cannot be empty")
+	}
```
lindgrenj6: role actually can be empty - empty role is default, but uniqueness per tenant is still enforced.

service/endpoint_validation.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if ecr.Scheme == nil || !schemeRegexp.MatchString(*ecr.Scheme) {
+		tmp := defaultScheme
+		ecr.Scheme = &tmp
+	}
+
+	if ecr.Host == nil || *ecr.Host == "" {
+		return fmt.Errorf("the host cannot be empty")
+	}
```
lindgrenj6: host actually can be empty too (for sources where they're just connecting via a receptor node they won't have host present).

lindgrenj6 (changes requested):
Just a few comments - I know the spec was probably your source but unfortunately its wrong. The model has most of the behavior (or lack thereof). 

Also I'm ok with a followup PR on the actual handlers - this PR is pretty large already.

dao/endpoint_dao.go
MikelAlejoBR: Got it, thanks!

model/endpoint_http.go
MikelAlejoBR: Great! Removing...

model/endpoint_http.go
MikelAlejoBR: All right, doing it now.

service/endpoint_validation.go
MikelAlejoBR: All right, now is set as optional.

service/endpoint_validation.go
MikelAlejoBR: My bad. I actually thought it was weird to validate for uniqueness when we're accepting empty strings, so I made the change to actually not accept them. Now it accepts empty strings and checks for uniqueness!

service/endpoint_validation.go
MikelAlejoBR: Got it, now we accept empty hosts, and validations only take place when receiving a non-empty one.

service/endpoint_validation.go
MikelAlejoBR: All right, removed that "required" constraint.

MikelAlejoBR:
Thanks for the review @lindgrenj6. I've implemented all the requested changes :+1:

lindgrenj6 (approved):
Looks great!

PR #45, opened 2021-11-04
Code simplification - check is solved on another place

petracihalova:
When ID will be again 0 then you get "not found" message + status 400. So it is not needed to have this additional check in code.

https://issues.redhat.com/browse/RHCLOUD-16849


lpichler (approved):
Looks great! Thanks !!! 👍

lindgrenj6 (approved):

PR #46, opened 2021-11-05
[RHCLOUD-16868] Setting the correct response status

petracihalova:
One task for changes in handlers and one task for added tests
https://issues.redhat.com/browse/RHCLOUD-16868
https://issues.redhat.com/browse/RHCLOUD-16726

### DIFF
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
@@ -55,8 +56,12 @@ func SourceListApplications(c echo.Context) error {
 	applications, count, err = applicationDB.SubCollectionList(m.Source{ID: id}, limit, offset, filters)
 
 	if err != nil {
+		if errors.Is(err, util.ErrNotFoundEmpty) {
+			return err
+		}
 		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
 	}
+
 	c.Logger().Infof("tenant: %v", *applicationDB.Tenant())
 
 	out := make([]interface{}, len(applications))
@@ -118,8 +123,12 @@ func ApplicationGet(c echo.Context) error {
 	c.Logger().Infof("Getting Application ID %v", id)
 
 	app, err := applicationDB.GetById(&id)
+
 	if err != nil {
-		return c.JSON(http.StatusNotFound, util.ErrorDoc(err.Error(), "404"))
+		if errors.Is(err, util.ErrNotFoundEmpty) {
+			return err
+		}
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
 	}
 
 	return c.JSON(http.StatusOK, app.ToResponse())
--- a/dao/endpoint_dao.go
+++ b/dao/endpoint_dao.go
@@ -3,7 +3,6 @@ package dao
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/RedHatInsights/sources-api-go/util"
 
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
@@ -17,7 +16,7 @@ func (a *EndpointDaoImpl) SubCollectionList(primaryCollection interface{}, limit
 	endpoints := make([]m.Endpoint, 0, limit)
 	sourceType, err := m.NewRelationObject(primaryCollection, *a.TenantID, DB.Debug())
 	if err != nil {
-		return nil, 0, util.NewErrNotFound("endpoint")
+		return nil, 0, util.NewErrNotFound("source")
 	}
 
 	query := sourceType.HasMany(&m.Endpoint{}, DB.Debug())
--- a/dao/endpoint_dao.go
+++ b/dao/endpoint_dao.go
@@ -3,6 +3,7 @@ package dao
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/RedHatInsights/sources-api-go/util"
 
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
@@ -14,11 +15,11 @@ type EndpointDaoImpl struct {
 
 func (a *EndpointDaoImpl) SubCollectionList(primaryCollection interface{}, limit int, offset int, filters []util.Filter) ([]m.Endpoint, int64, error) {
 	endpoints := make([]m.Endpoint, 0, limit)
-
 	sourceType, err := m.NewRelationObject(primaryCollection, *a.TenantID, DB.Debug())
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, util.NewErrNotFound("endpoint")
 	}
+
 	query := sourceType.HasMany(&m.Endpoint{}, DB.Debug())
 	query = query.Where("endpoints.tenant_id = ?", a.TenantID)
 
@@ -55,8 +56,11 @@ func (a *EndpointDaoImpl) List(limit int, offset int, filters []util.Filter) ([]
 func (a *EndpointDaoImpl) GetById(id *int64) (*m.Endpoint, error) {
 	app := &m.Endpoint{ID: *id}
 	result := DB.First(&app)
+	if result.Error != nil {
+		return nil, util.NewErrNotFound("endpoint")
+	}
 
-	return app, result.Error
+	return app, nil
 }
 
 func (a *EndpointDaoImpl) Create(app *m.Endpoint) error {
--- a/dao/meta_data_dao.go
+++ b/dao/meta_data_dao.go
@@ -2,7 +2,6 @@ package dao
 
 import (
 	"fmt"
-	"github.com/RedHatInsights/sources-api-go/util"
 
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
@@ -16,7 +15,7 @@ func (a *MetaDataDaoImpl) SubCollectionList(primaryCollection interface{}, limit
 	metadatas := make([]m.MetaData, 0, limit)
 	collection, err := m.NewRelationObject(primaryCollection, -1, DB.Debug())
 	if err != nil {
-		return nil, 0, util.NewErrNotFound("application")
+		return nil, 0, util.NewErrNotFound("application type")
 	}
 
 	query := collection.HasMany(&m.MetaData{}, DB.Debug())
@@ -54,7 +53,7 @@ func (a *MetaDataDaoImpl) GetById(id *int64) (*m.MetaData, error) {
 	metaData := &m.MetaData{ID: *id}
 	result := DB.First(&metaData)
 	if result.Error != nil {
-		return nil, util.NewErrNotFound("application")
+		return nil, util.NewErrNotFound("metadata")
 	}
 
 	return metaData, nil
--- a/dao/meta_data_dao.go
+++ b/dao/meta_data_dao.go
@@ -2,6 +2,7 @@ package dao
 
 import (
 	"fmt"
+	"github.com/RedHatInsights/sources-api-go/util"
 
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
@@ -15,7 +16,7 @@ func (a *MetaDataDaoImpl) SubCollectionList(primaryCollection interface{}, limit
 	metadatas := make([]m.MetaData, 0, limit)
 	collection, err := m.NewRelationObject(primaryCollection, -1, DB.Debug())
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, util.NewErrNotFound("application")
 	}
 
 	query := collection.HasMany(&m.MetaData{}, DB.Debug())
@@ -52,8 +53,11 @@ func (a *MetaDataDaoImpl) List(limit int, offset int, filters []util.Filter) ([]
 func (a *MetaDataDaoImpl) GetById(id *int64) (*m.MetaData, error) {
 	metaData := &m.MetaData{ID: *id}
 	result := DB.First(&metaData)
+	if result.Error != nil {
+		return nil, util.NewErrNotFound("application")
+	}
 
-	return metaData, result.Error
+	return metaData, nil
 }
 
 func (a *MetaDataDaoImpl) Create(metaData *m.MetaData) error {
--- a/dao/mock_daos.go
+++ b/dao/mock_daos.go
@@ -31,8 +31,6 @@ type MockMetaDataDao struct {
 }
 
 func (src *MockSourceDao) SubCollectionList(primaryCollection interface{}, limit, offset int, filters []util.Filter) ([]m.Source, int64, error) {
-	count := int64(1)
-
 	var sources []m.Source
 
 	for index, i := range src.Sources {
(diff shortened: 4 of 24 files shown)


lindgrenj6 (changes requested):
So - the idea is right here, but the implementation can be changed. This can be accomplished using a custom `error` type. Let me outline it here.

---

In Go an `error` is a type that implements the `Error` interface, the interface is actually very simple:
```go
type error interface {
	Error() string
}
```
see this blog post for a bit more in-depth: https://go.dev/blog/error-handling-and-go

---

For this specific use case I've been wanting to make a middleware that checks on a few different error types like not found, bad request, etc and handle them all in the middleware so we don't have to have `return c.JSON(...)` everywhere, we can instead just return the error in the controllers.

We can get started with that though by adding a `util.NotFoundError` struct that implements the `error` interface, and then returning that. Instead of comparing the message to a standard string we can just return a `NotFoundError` that has information about what wasn't found - that way we don't lose that information. That can be done with `errors.Is()` https://pkg.go.dev/errors#Is where you can check to see if an error is a certain type. In this case we would see if the error is not found and return a 404, otherwise 400. 

----

Like I said eventually we will probably extract this all to a middleware but for now we can just use a custom error throughout the controllers. Hopefully this isn't too confusing! Let me know if you need more guidance!

application_handlers.go
```diff
@@ -55,8 +56,12 @@ func SourceListApplications(c echo.Context) error {
 	applications, count, err = applicationDB.SubCollectionList(m.Source{ID: id}, limit, offset, filters)
 
 	if err != nil {
+		if errors.Is(err, &util.NotFoundError{}) {
```
lindgrenj6: Great! One final little tidbit as far as the comparisons go - we can have a variable over in util that represents this empty error, that way we don't have to rewrite the & and {} over and over again. I'll provide examples at the end of the review.

dao/mock_daos.go
```diff
@@ -64,7 +64,7 @@ func (src *MockSourceDao) GetById(id *int64) (*m.Source, error) {
 		}
 	}
 
-	return nil, fmt.Errorf("source not found")
+	return nil, fmt.Errorf("record not found")
```
lindgrenj6: So - for this I would probably make the `ErrNotFound` struct contain a string variable that is the type that wasn't found. For example we'd `return nil, util.ErrNotFound{Type: "source"}` instead and make the Error() method be `return type + " not found"`

middleware/error_handling.go
```diff
@@ -12,10 +13,17 @@ func HandleErrors(next echo.HandlerFunc) echo.HandlerFunc {
...
-			)
+			var statusCode int
+			var message interface{}
+
+			if errors.Is(err, &util.NotFoundError{}) {
+				statusCode = http.StatusNotFound
+				message = util.ErrorDoc(err.Error(), "404")
+			} else {
+				statusCode = http.StatusInternalServerError
+				message = util.ErrorDoc(fmt.Sprintf("Internal Server Error: %v", err.Error()), "500")
```
lindgrenj6: I like what you did here a lot, this is great. This way we can just let the error bubble up to the middleware which handles the return!

util/error_document.go
```diff
@@ -20,3 +22,10 @@ func ErrorDoc(message, status string) *ErrorDocument {
 		}},
 	}
 }
+
+type NotFoundError struct {
+}
+
+func (e *NotFoundError) Error() string {
+	return "record not found"
+}
```
lindgrenj6: So right here I would flesh this struct out a bit more, here are my suggestions:

1. Change the name to `ErrNotFound` to follow Go conventions (example here: https://pkg.go.dev/io/fs#ErrPermission)
2. Add an `Is` method so that `errors.Is` will check more than just the type + string, so we can make the string whatever we want and be dynamically generated when we call `Error()`

lindgrenj6 (commented):
So this is very very very close - just a few more things to make the NotFoundError a bit more idiomatic. I'm super happy about the middleware piece! That way we can bubble up the not found error type and it just does the right thing. 

I did a bit of an example program here to illustrate how `errors.Is` works since it isn't very useful unless the error has the Is method on it. 

https://gist.github.com/lindgrenj6/571699f5af0f033cb96cfbfa8b513171

Let me know if you have any questions!

application_handlers.go
petracihalova: Thanks for comments. My first understanding was when we want to compare error type on this place then we need the variable err contains this new ErrNotFound type, right? And place where this value is created is model/relation_object.go : 127. I changed there

petracihalova: Thanks for comments. My first understanding was when we want to compare error type on this place then we need the variable err contains this new ErrNotFound type, right? And place where this value is created is model/relation_object.go : 127. But then I had meeting with Libor and his suggestion was that the creation of concrete error type should be on handler level - so e.g. in application_handlers.go file. But in this case the variable err doesnt contain ErrNotFound type. So second idea is to compare error strings in handlers (because what else can I do?) and then create error type (now only ErrNotFound, but in future ErrBadRequest ....) and probably wrap this in separate function, because this is needed several times in each handler.

MikelAlejoBR: >  My first understanding was when we want to compare error type on this place then we need the variable err contains this new ErrNotFound type, right? And place where this value is created is model/relation_object.go : 127.

Yes, the `err` would be a `ErrNotFound` in [model/relation_object.go#L123](https://github.com/RedHatInsights/sources-api-go/blob/main/model/relation_object.go#L123).

> But then I had meeting with Libor and his suggestion was that the creation of concrete error type should be on handler level - so e.g. in application_handlers.go file. But in this case the variable err doesnt contain ErrNotFound type.

Maybe it's because I come from a different programming language, but personally I would create the concrete error types in the lower levels and then bubble them up to the handlers —or wherever the error must be caught—, and then transform those error types to proper responses.

application_handlers.go
```diff
@@ -55,8 +56,12 @@ func SourceListApplications(c echo.Context) error {
 	applications, count, err = applicationDB.SubCollectionList(m.Source{ID: id}, limit, offset, filters)
 
 	if err != nil {
+		if errors.Is(err, util.ErrNotFoundEmpty) {
+			return err
+		}
 		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
```
lindgrenj6, line 62: :tada: looks great. Can't wait until we don't have to have this if anymore (once we flesh out bad request handling)

internal/testutils/request/request.go
```diff
@@ -22,3 +37,33 @@ func CreateTestContext(method string, path string, body io.Reader, context map[s
...
+	for k, v := range reqPar.Context {
+		echoContext.Set(k, v)
+	}
+
+	for k, v := range reqPar.Headers {
+		request.Header.Set(k, v)
+	}
+
+	return recorder, request
+}
```
lindgrenj6: So - this is fairly similar to what we have at [testutils.CreateTestContext](https://github.com/RedHatInsights/sources-api-go/blob/e5ad9bbed6af58bf92fb9540cefebac7fd42069c/internal/testutils/request/request.go#L14) minus the header piece. I think we should do what we've done in other pieces of the tests and just add the headers when necessary. We don't really need headers when testing that a controller returns 404.

internal/testutils/test_template.go
```diff
@@ -0,0 +1,35 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/RedHatInsights/sources-api-go/util"
+)
+
+func NotFoundTest(t *testing.T, rec *httptest.ResponseRecorder, expectedResults map[string]interface{}) {
+	if rec.Code != expectedResults["Code"] {
```
lindgrenj6: Since this test function knows we're testing for not found - we probably don't need to pass in "expected results".

We always will know:
1. the code is 404
2. the status will be `"404"`
3. the detail will end with "not found" (check out [strings.HasSuffix](https://pkg.go.dev/strings@go1.17.6#HasSuffix)

as long as these conditions are met it's a successful 404 handle.

main_test.go
```diff
@@ -79,3 +83,22 @@ func AssertLinks(t *testing.T, path string, links util.Links, limit int, offset
...
+	e := echo.New()
+
+	v3 := e.Group("/api/sources/v3.1", middleware.HandleErrors, middleware.ParseHeaders)
+	v3.GET(reqPar.Route, handlerFunc, MiddlewareFunc(flags.Integration)...)
+
+	recorder, request := request.CreateTestRecorder(e, reqPar)
+	e.ServeHTTP(recorder, request)
+
+	return recorder
+}
```
lindgrenj6: So instead of making another echo router with all the middleware of the regular app we can just work with contexts wrapped with middleware functions. I did something similar [here](https://github.com/RedHatInsights/sources-api-go/blob/e5ad9bbed6af58bf92fb9540cefebac7fd42069c/middleware/error_handling_test.go#L22).

---

Basically: somewhere shared (not sure where we're putting all this) write a function like this:
```go
func ErrorHandlingContext(handler echo.Handlerfunc) func(echo.Context) error {
    return func(c echo.Context) error {
        return middleware.HandleErrors(c)
    }
}
```

then construct a context with `testutils.CreateContext`, and call it like this: 
```go
notFoundSourceGet := ErrorHandlingContext(SourceGet)

err := notFoundSourceGet(c) // where c is something created earlier.
```

and since these are 404 tests just be sure to use an id that won't exist. The DAO will return a not found error and this tiny stack will do the right thing! From there you can use the (updated) function that tests that recorder for not found errors. 

---

Getting closer!

lindgrenj6 (commented):
It's getting really late for me - but this is so close. The handlers are all correct (imo) we just need to change how the tests were written to follow the same pattern as the rest of the handler tests!

dao/meta_data_dao.go
```diff
@@ -15,7 +15,7 @@ func (a *MetaDataDaoImpl) SubCollectionList(primaryCollection interface{}, limit
 	metadatas := make([]m.MetaData, 0, limit)
 	collection, err := m.NewRelationObject(primaryCollection, -1, DB.Debug())
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, util.NewErrNotFound("application")
```
MikelAlejoBR: Shouldn't this be an `"application_type"` instead of an `"application"`? This gets called from the `ApplicationTypeListMetaData` handler, which lists the metadatas according to an application type. So here gets checked if the application type exists before listing its metadatas, right?

dao/meta_data_dao.go
```diff
@@ -52,8 +52,11 @@ func (a *MetaDataDaoImpl) List(limit int, offset int, filters []util.Filter) ([]
 func (a *MetaDataDaoImpl) GetById(id *int64) (*m.MetaData, error) {
 	metaData := &m.MetaData{ID: *id}
 	result := DB.First(&metaData)
+	if result.Error != nil {
+		return nil, util.NewErrNotFound("application")
```
MikelAlejoBR: In here, shouldn't it be `NewErrNotFound("metadata")` instead?

dao/mock_daos.go
```diff
@@ -42,11 +40,17 @@ func (src *MockSourceDao) SubCollectionList(primaryCollection interface{}, limit
...
 				sources = append(sources, src.Sources[index])
 			}
+		default:
+			return nil, 0, fmt.Errorf("unexpected primary collection type")
 		}
+	}
 
+	count := int64(len(sources))
+	if count == 0 {
+		return nil, count, &util.ErrNotFound{Type: "source"}
```
MikelAlejoBR: I would use the `util.NewErrNotFound("source")` here, to make the most of that new function!

dao/mock_daos.go
```diff
@@ -64,7 +68,7 @@ func (src *MockSourceDao) GetById(id *int64) (*m.Source, error) {
 		}
 	}
 
-	return nil, fmt.Errorf("source not found")
+	return nil, &util.ErrNotFound{Type: "source"}
```
MikelAlejoBR: Here too, plus the other few places below where this raw struct is used instead

util/error_document.go
```diff
@@ -20,3 +27,19 @@ func ErrorDoc(message, status string) *ErrorDocument {
...
+	Type string
+}
+
+func (e ErrNotFound) Error() string {
+	return fmt.Sprintf("%s not found", e.Type)
+}
+
+func (e ErrNotFound) Is(err error) bool {
+	return reflect.TypeOf(err) == reflect.TypeOf(e)
+}
```
MikelAlejoBR, line 41: I don't fully understand the purpose of this function. The thing I understand is that I can check if another error is of the type "ErrNotFound". But wouldn't be using `errors.Is(someError, util.ErrNotFoundEmpty)` instead?

petracihalova: I also used errors.Is() before as you describe but Jacob explained me here https://gist.github.com/lindgrenj6/571699f5af0f033cb96cfbfa8b513171

MikelAlejoBR: Oh, that's very interesting! Thank you!

dao/meta_data_dao.go
petracihalova: done

dao/mock_daos.go
petracihalova: done

dao/mock_daos.go
petracihalova: fixed all occurrences

dao/meta_data_dao.go
petracihalova: I changed this to `return nil, 0, util.NewErrNotFound("metadata")` because of  function's return value ... but this is good question what is expected in this case (`"application type not found"` or `"metadata not found"`) when no metadatas are found for not existing application type (I think when you call `api/sources/v3.1/application_types/{{id}}/app_meta_data` there is no check if application type ID exists or not) ... final decision will affect all SubCollectionList methods .... EDIT: Libor showed me where we check if app type ID exists ... thanks @lpichler

internal/testutils/test_template.go
petracihalova: done

dao/meta_data_dao.go
petracihalova: I think we can return "metadata not found" and add some verification to the handler (before calling the SubCollectionList method) if the application type ID exists, and if not, then we can return "application type not found" ... but this can be solved in the next PR , what do you think? @lindgrenj6 @MikelAlejoBR @lpichler ... EDIT: I apologize for the chaos 😁😁😁 Libor explained it to me

petracihalova: ok, so I had meeting with Libor and he explained me this and I hope I understand it better now, so yes @MikelAlejoBR, there should be an "application_type"

petracihalova:
```
api/sources/v3.1/application_types/{{id}}/sources
api/sources/v3.1/source_types/{{id}}/sources
```
both use SubCollectionList() in source_dao.go, so for NotFound case we need to return "application type not found" or "source type not found" ... my idea was that we can use baseObject field in RelationObject struct, but it is private ... so I create new method which returns text representation of object type

and...
I know this PR is mess a little bit ... I think in the end I can squash second and third commit if review will be ok

lindgrenj6 (approved):

dao/endpoint_dao.go
```diff
@@ -14,11 +14,11 @@ type EndpointDaoImpl struct {
 
 func (a *EndpointDaoImpl) SubCollectionList(primaryCollection interface{}, limit int, offset int, filters []util.Filter) ([]m.Endpoint, int64, error) {
 	endpoints := make([]m.Endpoint, 0, limit)
-
 	sourceType, err := m.NewRelationObject(primaryCollection, *a.TenantID, DB.Debug())
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, util.NewErrNotFound("source type")
```
MikelAlejoBR: I think that this one is a "source" instead of a "source type".  The reason is that this comes from the "sources/:source_id/endpoints" route, which would list all the endpoints from a given source.

MikelAlejoBR (changes requested):
Just a minor comment. Other than that, looks good to me!

dao/endpoint_dao.go
petracihalova: yes i agree .. brilliant eye ... i checked it before and i thought i changed it but i missed it ... thanks

petracihalova:
Checks failed but it will be solved by https://github.com/RedHatInsights/sources-api-go/pull/84

lpichler (approved):
LGTM 👍 Good job 👏

PR #48, opened 2021-11-08
feature: include SonarQube's scanner on pr checks

MikelAlejoBR:
Includes SonarQube's scanner to be run along with the other stuff in the pr checks. 

[[RHCLOUD-16840]](https://issues.redhat.com/browse/RHCLOUD-16840)


MikelAlejoBR:
Essentially the PR is ready to be reviewed. You can ignore the "ci.ext.devshift" check because it is no more as I changed that external job for the internal one.

sonarqube.bash
```diff
@@ -0,0 +1,87 @@
...
+# use it in this script.
+java_home="/usr/lib/jvm/jre-1.8.0"
+
+"$java_home"/bin/keytool \
+  -alias "RH-IT-Root-CA" \
+  -file "$rh_it_root_ca_file" \
+  -import \
+  -keystore "$rh_it_keystore_file" \
+  -noprompt \
+  -storepass "$rh_it_keystore_pass"
```
lindgrenj6: :sweat:

sonarqube.bash
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
lindgrenj6: What would you think about hiding this file in a `scripts/` folder or something? Like I can understand having `pr_check.sh` and `build_deploy.sh` in the root since jenkins needs them - but this one doesn't need to be in the root necessarily?

lindgrenj6 (commented):
Looks good with one minor thing - nbd in case I'm crazy.

lpichler:
what is status of this PR, is it ready to _GO_ ?

MikelAlejoBR:
Not yet, sorry. I'm marking it as a draft because I'm trying the latest changes. I'm trying to set up a pipeline which first runs the "pr_check" and once that one succeeds, then launches another pipeline for the SonarQube scanner. However I'm finding it hard to do it, so if I don't make it, I'll just create two parallel jobs for them and that's it.

Essentially the bash code won't change — the only thing I'll do is move it to "scripts/", which I haven't done yet 😅

sonarqube.bash
MikelAlejoBR: Yep...

sonarqube.bash
MikelAlejoBR: Yeah, I agree. Moving it now!

MikelAlejoBR:
All right it seems that all pipelines passed so I'd say it is ready to be reviewed :+1:

lpichler (approved):
👏

lpichler:
When there is some issue what sonarcube reports about PR - then It will be in https://ci.int.devshift.net/job/RedHatInsights-sources-api-go-pr-check/69/console or I have to find it in somewhere https://sonarqube.corp.redhat.com/ ?

MikelAlejoBR:
The scanner only scans and uploads the results to the SonarQube instance, so unless there's a connection or certificate error I don't think the Jenkins job will fail much. You can find the results of the analyses by searching the name of the projects on SonarQube, e.g.: by searching "sources-api-go" on SQ you'll find the analyses for this project.

lpichler:
I am probably missing something but who runs ` scripts/sonarqube.bash` ?

lpichler:
> The scanner only scans and uploads the results to the SonarQube instance, so unless there's a connection or certificate error I don't think the Jenkins job will fail much. You can find the results of the analyses by searching the name of the projects on SonarQube, e.g.: by searching "sources-api-go" on SQ you'll find the analyses for this project.

yes it looks it is here - https://sonarqube.corp.redhat.com/dashboard?id=console.redhat.com%3Asources-api-go
so does it mean that I cannot check results for opened PR ? (Sonarqube is totally new for me )

MikelAlejoBR:
> I am probably missing something but who runs ` scripts/sonarqube.bash` ?

We have a Jenkins job which runs that script, the same way as the other jobs that directly run the "pr_check" and "build" scripts. I'll write some docs about it referencing the exact callers so we all can understand this better.

> > The scanner only scans and uploads the results to the SonarQube instance, so unless there's a connection or certificate error I don't think the Jenkins job will fail much. You can find the results of the analyses by searching the name of the projects on SonarQube, e.g.: by searching "sources-api-go" on SQ you'll find the analyses for this project.
> 
> yes it looks it is here - https://sonarqube.corp.redhat.com/dashboard?id=console.redhat.com%3Asources-api-go so does it mean that I cannot check results for opened PR ? (Sonarqube is totally new for me )

Next to the project's name, you can switch branches to see the results of the analysis for the PR. Essentially, the scanner has a reference branch —which usually is the main/master one— and compares that to the PR's branch. That way, you can see if the code quality improved or decreased compared to the main branch.

lpichler:
Thank you very much for answers @MikelAlejoBR now I can see it 👍

lindgrenj6 (approved):

PR #49, opened 2021-11-18
Availability Status Listener

lpichler:
- [x] [this](https://github.com/RedHatInsights/sources-api-go/pull/51)  needs to be merged first.
---

Converts ruby (Availability Status Listener https://github.com/RedHatInsights/sources-api/blob/master/lib/availability_status_listener.rb) to GO

[Checks](https://github.com/RedHatInsights/sources-api/blob/master/lib/availability_status_listener.rb#L43-L48) skipped for `x-rh-identity`, `x-rh-sources-account-number` in headers of Kafka message

Missing:
- [x] Tests

Following:
- add `raiseForUpdate` to missing places https://issues.redhat.com/browse/RHCLOUD-17055
- possibility to disable Availability Status Listener in ruby repo by ENV
- go routines for `RaiseEventForUpdate` https://issues.redhat.com/browse/RHCLOUD-17056
- Authentication and ApplicationAuthentication are not included https://issues.redhat.com/browse/RHCLOUD-17059

### DIFF
--- a/dao/application_dao.go
+++ b/dao/application_dao.go
@@ -93,12 +93,12 @@ func (a *ApplicationDaoImpl) BulkMessage(id *int64) (map[string]interface{}, err
 }
 
 func (a *ApplicationDaoImpl) FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error {
-	src, err := a.GetById(id)
-	if err == nil {
-		err = DB.Model(src).Updates(updateAttributes).Error
+	result := DB.Model(&m.Application{ID: *id}).Updates(updateAttributes)
+	if result.RowsAffected == 0 {
+		return fmt.Errorf("application not found %v", id)
 	}
 
-	return err
+	return nil
 }
 
 func (a *ApplicationDaoImpl) FindWithTenant(id *int64) (*m.Application, error) {
--- a/dao/application_dao.go
+++ b/dao/application_dao.go
@@ -89,7 +89,7 @@ func (a *ApplicationDaoImpl) BulkMessage(id *int64) (map[string]interface{}, err
 		return nil, resource.Error
 	}
 
-	return BulkMessageFrom(&application.Source)
+	return BulkMessageFromSource(&application.Source)
 }
 
 func (a *ApplicationDaoImpl) FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error {
--- a/dao/application_dao.go
+++ b/dao/application_dao.go
@@ -83,33 +83,13 @@ func (a *ApplicationDaoImpl) Tenant() *int64 {
 
 func (a *ApplicationDaoImpl) BulkMessage(id *int64) (map[string]interface{}, error) {
 	application := &m.Application{ID: *id}
-	resource := DB.Preload("Source.Tenant").Preload("Source.Applications.Tenant").Preload("Source.Endpoints.Tenant").Find(&application)
+	resource := DB.Preload("Source").Find(&application)
 
 	if resource.Error != nil {
 		return nil, resource.Error
 	}
 
-	bulkMessage := map[string]interface{}{}
-	bulkMessage["source"] = application.Source.ToEvent()
-
-	endpoints := make([]m.EndpointEvent, len(application.Source.Endpoints))
-	for i, endpoint := range application.Source.Endpoints {
-		endpoints[i] = *endpoint.ToEvent()
-	}
-
-	bulkMessage["endpoints"] = endpoints
-
-	applications := make([]m.ApplicationEvent, len(application.Source.Applications))
-	for i, application := range application.Source.Applications {
-		applications[i] = *application.ToEvent()
-	}
-
-	bulkMessage["applications"] = applications
-
-	bulkMessage["authentications"] = []m.Authentication{}
-	bulkMessage["application_authentications"] = []m.ApplicationAuthenticationEvent{}
-
-	return bulkMessage, nil
+	return BulkMessageFrom(&application.Source)
 }
 
 func (a *ApplicationDaoImpl) FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error {
--- a/dao/application_dao.go
+++ b/dao/application_dao.go
@@ -114,7 +114,7 @@ func (a *ApplicationDaoImpl) BulkMessage(id *int64) (map[string]interface{}, err
 
 func (a *ApplicationDaoImpl) FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error {
 	src, err := a.GetById(id)
-	if err != nil {
+	if err == nil {
 		err = DB.Model(src).Updates(updateAttributes).Error
 	}
 
--- a/dao/application_dao.go
+++ b/dao/application_dao.go
@@ -1,6 +1,7 @@
 package dao
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/RedHatInsights/sources-api-go/middleware"
@@ -78,4 +79,58 @@ func (a *ApplicationDaoImpl) Delete(id *int64) error {
 
 func (a *ApplicationDaoImpl) Tenant() *int64 {
 	return a.TenantID
+}
+
+func (a *ApplicationDaoImpl) BulkMessage(id *int64) (map[string]interface{}, error) {
+	application := &m.Application{ID: *id}
+	resource := DB.Preload("Source.Tenant").Preload("Source.Applications.Tenant").Preload("Source.Endpoints.Tenant").Find(&application)
+
+	if resource.Error != nil {
+		return nil, resource.Error
+	}
+
+	bulkMessage := map[string]interface{}{}
+	bulkMessage["source"] = application.Source.ToEvent()
+
+	endpoints := make([]m.EndpointEvent, len(application.Source.Endpoints))
+	for i, endpoint := range application.Source.Endpoints {
+		endpoints[i] = *endpoint.ToEvent()
+	}
+
+	bulkMessage["endpoints"] = endpoints
+
+	applications := make([]m.ApplicationEvent, len(application.Source.Applications))
+	for i, application := range application.Source.Applications {
+		applications[i] = *application.ToEvent()
+	}
+
+	bulkMessage["applications"] = applications
+
+	bulkMessage["authentications"] = []m.Authentication{}
+	bulkMessage["application_authentications"] = []m.ApplicationAuthenticationEvent{}
+
+	return bulkMessage, nil
+}
+
+func (a *ApplicationDaoImpl) FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error {
+	src, err := a.GetById(id)
+	if err != nil {
+		err = DB.Model(src).Updates(updateAttributes).Error
+	}
+
+	return err
+}
+
+func (a *ApplicationDaoImpl) FindWithTenant(id *int64) (*m.Application, error) {
+	app := &m.Application{ID: *id}
+	result := DB.Preload("Tenant").Find(&app)
+
+	return app, result.Error
+}
+
+func (a *ApplicationDaoImpl) ToEventJSON(id *int64) ([]byte, error) {
+	app, err := a.FindWithTenant(id)
+	data, _ := json.Marshal(app.ToEvent())
+
+	return data, err
 }
--- a/dao/common.go
+++ b/dao/common.go
@@ -25,7 +25,11 @@ func GetFrom(resourceType string) (*m.EventModelDao, error) {
 }
 
 func BulkMessageFrom(source *m.Source) (map[string]interface{}, error) {
-	result := DB.Preload("Tenant").Preload("Applications.Tenant").Preload("Endpoints.Tenant").Find(&source)
+	result := DB.
+		Preload("Tenant").
+		Preload("Applications.Tenant").
+		Preload("Endpoints.Tenant").
+		Find(&source)
 
 	if result.Error != nil {
 		return nil, result.Error
--- a/dao/common.go
+++ b/dao/common.go
@@ -1,0 +1,25 @@
+package dao
+
+import (
+	"fmt"
+
+	m "github.com/RedHatInsights/sources-api-go/model"
+)
+
+func GetFrom(resourceType string) (*m.EventModelDao, error) {
+	var resource m.EventModelDao
+	switch resourceType {
+	case "Source":
+		resource = &SourceDaoImpl{}
+	case "Endpoint":
+		resource = &EndpointDaoImpl{}
+	case "Application":
+		resource = &ApplicationDaoImpl{}
+	case "Authentication":
+		resource = &AuthenticationDaoImpl{}
+	default:
+		return nil, fmt.Errorf("invalid resource_type (%s) to get DAO instance", resourceType)
+	}
+
+	return &resource, nil
+}
--- a/dao/common.go
+++ b/dao/common.go
@@ -6,7 +6,7 @@ import (
 	m "github.com/RedHatInsights/sources-api-go/model"
 )
 
-func GetFrom(resourceType string) (*m.EventModelDao, error) {
+func GetFromResourceType(resourceType string) (*m.EventModelDao, error) {
 	var resource m.EventModelDao
 	switch resourceType {
 	case "Source":
--- a/dao/common.go
+++ b/dao/common.go
@@ -2,6 +2,7 @@ package dao
 
 import (
 	"fmt"
+
 	m "github.com/RedHatInsights/sources-api-go/model"
 )
 
--- a/dao/common.go
+++ b/dao/common.go
@@ -24,7 +24,7 @@ func GetFromResourceType(resourceType string) (*m.EventModelDao, error) {
 	return &resource, nil
 }
 
-func BulkMessageFrom(source *m.Source) (map[string]interface{}, error) {
+func BulkMessageFromSource(source *m.Source) (map[string]interface{}, error) {
 	result := DB.
 		Preload("Tenant").
 		Preload("Applications.Tenant").
--- a/dao/common.go
+++ b/dao/common.go
@@ -2,7 +2,6 @@ package dao
 
 import (
 	"fmt"
-
 	m "github.com/RedHatInsights/sources-api-go/model"
 )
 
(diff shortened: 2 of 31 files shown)


dao/application_dao.go
```diff
@@ -79,3 +80,57 @@ func (a *ApplicationDaoImpl) Delete(id *int64) error {
 func (a *ApplicationDaoImpl) Tenant() *int64 {
 	return a.TenantID
 }
+
+func (a *ApplicationDaoImpl) BulkMessage(id *int64) (map[string]interface{}, error) {
+	application := &m.Application{ID: *id}
+	resource := DB.Preload("Source.Tenant").Preload("Source.Applications.Tenant").Preload("Source.Endpoints.Tenant").Find(&application)
```
lindgrenj6: Super small suggestion just for readability + styling of these really long method chains:

```suggestion
	resource := DB.
		Preload("Source.Tenant").
		Preload("Source.Applications.Tenant").
		Preload("Source.Endpoints.Tenant").
		Find(&application)
```

Can we apply this to all implementations of this method?

dao/application_dao.go
```diff
@@ -79,3 +80,57 @@ func (a *ApplicationDaoImpl) Delete(id *int64) error {
...
+
+	if resource.Error != nil {
+		return nil, resource.Error
+	}
+
+	bulkMessage := map[string]interface{}{}
```
lindgrenj6: We can do this in a follow-up PR I think - but I really think that there can be a `BulkMessage` struct that codifies the structure of the message.

dao/application_dao.go
```diff
@@ -79,3 +80,57 @@ func (a *ApplicationDaoImpl) Delete(id *int64) error {
...
+	return bulkMessage, nil
+}
+
+func (a *ApplicationDaoImpl) FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error {
+	src, err := a.GetById(id)
+	if err != nil {
+		err = DB.Model(src).Updates(updateAttributes).Error
+	}
+
+	return err
```
lindgrenj6: This can be slightly refactored here:

```suggestion
	result := DB.Model(&m.Application{ID: *id}).Updates(updateAttributes)
	if result.RowsEffected == 0 {
		return fmt.Errorf("application not found %v", id)
	}
	
	return nil
```

We do not need to fetch the app from the db before the update, this should generate the same sql - we can key off of whether any rows were touched in the update - it returns 1 even if no fields get updated.

internal/events/event_stream_producer.go
```diff
@@ -0,0 +1,82 @@
...
+	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/util"
+)
+
+const EventStreamTopic = "platform.sources.event-stream"
+
+var config = c.Get()
+
+func raiseEvent(eventType string, payload []byte, headers []kafka.Header) error {
+	logging.Log.Debugf("\"publishing message to topic \"platform.sources.event-stream\"...")
```
lindgrenj6: For future reference we can use [string literals](https://go.dev/ref/spec#String_literals) here so you don't need to do the escaping, super handy!

Though I'm not sure we need the quotes in the log message - maybe something like this (`%q` adds quotes to the string):

```go
	logging.Log.Debugf("publishing message to topic %q...", EventStreamTopic)
```

main.go
```diff
@@ -1,25 +1,39 @@
...
 func main() {
-	e := echo.New()
-
 	logging.InitLogger(conf)
-	logging.InitEchoLogger(e, conf)
 
 	dao.Init()
 	redis.Init()
 
+	availabilityListener := flag.Bool("listener", false, "run unit or integration tests")
```
lindgrenj6: Flag description here needs updated.

model/application.go
```diff
@@ -29,6 +30,27 @@ type Application struct {
 	ApplicationType   ApplicationType
 }
 
+func (app *Application) ToEvent() *ApplicationEvent {
+	asEvent := AvailabilityStatusEvent{AvailabilityStatus: util.StringValueOrNil(app.AvailabilityStatus.AvailabilityStatus),
+		LastAvailableAt: util.StringValueOrNil(util.FormatTimeToString(app.LastAvailableAt, "2006-01-02 15:04:05 MST")),
+		LastCheckedAt:   util.StringValueOrNil(util.FormatTimeToString(app.LastCheckedAt, "2006-01-02 15:04:05 MST"))}
```
lindgrenj6: A few things:
1. Instead of a helper function we can just use the [`time#Format(string)`](https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/time/format.go;l=558) method. 
2. The time package has [a bunch of handy constants, including one that matches Ruby's time format](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/time/format.go;l=95)

So we can change this to just be something like `app.LastAvailableAt.Format(time.RubyDate)`

model/events.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+type EventModelDao interface {
+	BulkMessage(id *int64) (map[string]interface{}, error)
+	FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error
+	ToEventJSON(id *int64) ([]byte, error)
+}
+
+func UpdateMessage(eventObject *EventModelDao, resourceID int64, resourceType string, attributes []string) ([]byte, error) {
+	updatedAttributes := map[string]interface{}{}
```
lindgrenj6, line 16: For future reference - instead of the double `{}{}` we can do `make(map[string]interface{})`

I think this will be nice to get in a struct too eventually.

statuslistener/statuslistener.go
```diff
@@ -0,0 +1,140 @@
...
+		return
+	}
+
+	if !util.SliceContainsString(m.AvailabilityStatuses, statusMessage.Status) {
+		logging.Log.Errorf("Invalid Status: %s", statusMessage.Status)
+		return
+	}
+
+	updateAttributes := attributesForUpdate(statusMessage)
+	modelEventDao, errorDao := dao.GetFrom(statusMessage.ResourceType)
```
lindgrenj6: Can we use `err` for all the error names? It's a bit difficult to read with multiple assigned errors.

statuslistener/statuslistener.go
```diff
@@ -0,0 +1,140 @@
...
+			return
+		}
+		et := eventType(message)
+		if et != EventAvailabilityStatus {
+			return
+		}
+
+		logging.Log.Infof("Kafka message %s, %s received with payload: %s", message.Headers, message.Key, message.Value)
+
+		headers := headersWithoutEventType(message)
```
lindgrenj6: I'm not sure we need to strip the event_type out - unless we're doing that on the ruby side already. iirc we override the `event_type` when we go to produce it with the `#{model.class}.update` message later anyway.

lpichler:
[this](https://github.com/RedHatInsights/sources-api-go/pull/51)  needs to be merged first.

dao/application_dao.go
lpichler: Done in https://github.com/RedHatInsights/sources-api-go/pull/49/commits/52a842b0c87179917d39620f2a2a925439c9c6bd

dao/application_dao.go
lpichler: This will be done in follow up PR.

dao/application_dao.go
lpichler: Done in https://github.com/RedHatInsights/sources-api-go/pull/49/commits/6a5d9ab142e8cabf74a243d747d19f5701580088

internal/events/event_stream_producer.go
lpichler: this is done in https://github.com/RedHatInsights/sources-api-go/pull/49/commits/52a842b0c87179917d39620f2a2a925439c9c6bd.

+ I will look on string literals to use them eventually in follow up PRs

main.go
lpichler: done in https://github.com/RedHatInsights/sources-api-go/pull/49/commits/02555b9aa0cf61231e5a738fc8d62e7ea15bbb7a

model/events.go
lpichler: This will be done in follow up PR.

statuslistener/statuslistener.go
lpichler: done in https://github.com/RedHatInsights/sources-api-go/pull/49/commits/e8a243e6e01a98c3f604b26d19b756247344ccd6

statuslistener/statuslistener.go
lpichler: yes that's right we don't need it.

I had to rewrite it in function where we are sending to Kafka.

done in https://github.com/RedHatInsights/sources-api-go/pull/49/commits/6416aefc1eb957b7e19da809c61b431d0c03e009

lpichler:
@lindgrenj6 - it is ready for re-review.

What is new ?
- I had to move some functions to have receiver - so I could use interfaces to test the code +  I added all-in-one test
  - https://github.com/RedHatInsights/sources-api-go/pull/49/commits/7d6616e3f9d7db604606662a706df9fc47da54cb
- I Unified bulk message function for Source, Application and Endpoint https://github.com/RedHatInsights/sources-api-go/pull/49/commits/bc451108b9b68e7fd442a6c4a7d741cb522465fb
- I added bunch of commits according to your review but there are small :)
  -  I added to your comments what is done in this PR and what will be done in future PR.


I know this PR is big, so if this PR needs live review let me know!

thanks @lindgrenj6

dao/common.go
```diff
@@ -0,0 +1,59 @@
+package dao
+
+import (
+	"fmt"
+
+	m "github.com/RedHatInsights/sources-api-go/model"
+)
+
+func GetFrom(resourceType string) (*m.EventModelDao, error) {
```
lindgrenj6: Can we rename this to something more descriptive? `GetFrom` is a bit too vague - maybe `GetDaoFromType` or something along those lines

dao/common.go
```diff
@@ -0,0 +1,59 @@
...
+	case "Authentication":
+		resource = &AuthenticationDaoImpl{}
+	default:
+		return nil, fmt.Errorf("invalid resource_type (%s) to get DAO instance", resourceType)
+	}
+
+	return &resource, nil
+}
+
+func BulkMessageFrom(source *m.Source) (map[string]interface{}, error) {
```
lindgrenj6: Similar with this one - maybe `BulkMessageForSource`

internal/events/event_stream_producer.go
```diff
@@ -0,0 +1,101 @@
...
+func (esp *EventStreamSender) RaiseEvent(eventType string, payload []byte, headers []kafka.Header) error {
+	logging.Log.Debugf("publishing message to topic %q...", EventStreamTopic)
+
+	producerConfig := kafka.ProducerConfig{Topic: config.KafkaTopic(EventStreamTopic)}
+	kafkaConfig := kafka.Config{KafkaBrokers: config.KafkaBrokers, ProducerConfig: producerConfig}
+	kf := &kafka.Manager{Config: kafkaConfig}
+
+	m := &kafka.Message{}
+
+	for index, header := range headers {
```
lindgrenj6, line 36: Can probably switch this loop out for `m.GetHeader("event_type")`

model/event_models.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+type AuthenticationEvent struct {
+	ID        string    `json:"id"`
+	CreatedAt time.Time `json:"created_at"`
+
+	Name                    string                 `json:"name"`
+	AuthType                string                 `json:"authtype"`
+	Username                string                 `json:"username"`
+	Password                string                 `json:"password"`
```
lindgrenj6: We should remove this field - we don't produce it for the event stream since its globally readable by anyone on the platform.

model/source.go
```diff
@@ -41,6 +43,34 @@ type Source struct {
 	Endpoints        []Endpoint
 }
 
+func (src *Source) AsJSON() []byte {
```
lindgrenj6: This isn't getting called - can it be removed?

statuslistener/statuslistener.go
```diff
@@ -0,0 +1,153 @@
...
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/util"
+)
+
+const (
+	SourcesStatusTopic      = "platform.sources.status"
+	GroupID                 = "sources-api-status-worker"
+	EventAvailabilityStatus = "availability_status"
+)
```
lindgrenj6: These can probably be un-exported, not sure why they'd need to be accessed from outside this package.

statuslistener/statuslistener.go
```diff
@@ -0,0 +1,153 @@
...
+	if logging.Log == nil {
+		panic("logging is not initialized")
+	}
+
+	consumerConfig := kafka.ConsumerConfig{Topic: config.KafkaTopic(SourcesStatusTopic), GroupID: GroupID}
+	kafkaConfig := kafka.Config{KafkaBrokers: config.KafkaBrokers, ConsumerConfig: consumerConfig}
+
+	kf := &kafka.Manager{Config: kafkaConfig}
+
+	errorConsume := kf.Consume(avs.ConsumeStatusMessage)
```
lindgrenj6: can we rename this error var to just `err` as per conventions?

statuslistener/statuslistener.go
```diff
@@ -0,0 +1,153 @@
...
+	headers := make([]kafka.Header, len(message.Headers))
+	for index, header := range message.Headers {
+		headers[index] = kafka.Header{Key: header.Key, Value: header.Value}
+	}
+
+	return headers
+}
+
+func (avs *AvailabilityStatusListener) eventType(message kafka.Message) string {
+	for _, header := range message.Headers {
```
lindgrenj6: Probably can replace this with `m.GetHeader("event_type")`

statuslistener/statuslistener.go
```diff
@@ -0,0 +1,153 @@
...
+
+	if !util.SliceContainsString(m.AvailabilityStatuses, statusMessage.Status) {
+		logging.Log.Errorf("Invalid Status: %s", statusMessage.Status)
+		return
+	}
+
+	updateAttributes := avs.attributesForUpdate(statusMessage)
+	modelEventDao, err := dao.GetFrom(statusMessage.ResourceType)
+	if err != nil {
+		logging.Log.Error(err.Error())
```
lindgrenj6: I think you can just pass `err` and it'll call the `Error()` method for you.

statuslistener/types.go
```diff
@@ -0,0 +1,8 @@
+package statuslistener
+
+type StatusMessage struct {
```
lindgrenj6, line 3: :tada: I am so happy to see this! It's finally codified.

util/parser.go
```diff
@@ -3,8 +3,11 @@ package util
 import (
 	"fmt"
 	"strconv"
+	"time"
 )
 
+var RecordDateTimeFormat = "2006-01-02 15:04:05 MST"
```
lindgrenj6, line 9: I _think_ this is similar to RFC3339, and there is a constant here: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/time/format.go;l=101

That matches what the current API is returning, they even have a `RubyDate` constant too if we wanted to use that.

statuslistener/statuslistener.go
```diff
@@ -0,0 +1,153 @@
...
+	if errorConsume != nil {
+		logging.Log.Errorf("Consumer kafka message error: %s", errorConsume.Error())
+	}
+}
+
+func (avs *AvailabilityStatusListener) ConsumeStatusMessage(message kafka.Message) {
+	var statusMessage StatusMessage
+	err := message.ParseTo(&statusMessage)
+	if err != nil {
+		logging.Log.Errorf("Error in parsing status message %s", err.Error())
```
lindgrenj6: So for this kind of log its probably better to use `%v` (its automatic per the type) that way you only have to pass in `err` instead of calling `err.Error()`.

dao/common.go
lpichler: to be honest in addition I was thinking how to remove "Get" part as it looks to me redundant but then "Dao" would became also redundant :)  :
`dao.DaoFrom("Source")` :) 

Descriptive part is here also first parameter - so it should be understandable enough - anyway I change it as well for certitude to `GetFromResourceType`

internal/events/event_stream_producer.go
lpichler: Here I am no sure  - how do you mean it ?

`m.GetHeader("event_type")` finds and returns in m.Headers value with "event_type" key.

but what I need here is to replace value with key "event_type"  in `headers` as header with key "event_type" is already present in `headers` because those headers came from message sent to kafka with status message - header with "event_type" is required here(./statuslistener/statuslistener.go:64)

model/source.go
lpichler: sure, it was leftover - good eye.

statuslistener/statuslistener.go
lpichler: (function eventType removed)

util/parser.go
lpichler: Currently returned by our ruby/rails(and this RecordDateTimeFormat) availability status listener: `2021-12-21 11:25:36 UTC`
time.RFC3339: `2021-12-21T11:26:40Z`
time.RubyDate: `Tue Dec 21 11:27:53 +0000 2021`

Of course I would rather use existing constant but it is not same - so it could potentially break systems which are using it and it is main reason why I defined own.

Let me know whether we can change it - but as I am looking on format constants - I still prefer `RecordDateTimeFormat` :) it has good readability

lpichler:
@lindgrenj6 I created new commits according to your suggestions (except to 2 - https://github.com/RedHatInsights/sources-api-go/pull/49#discussion_r772589582 and https://github.com/RedHatInsights/sources-api-go/pull/49#discussion_r772596027 )
thanks

internal/events/event_stream_producer.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (esp *EventStreamProducer) RaiseEventForUpdate(resourceID int64, resourceType string, updateAttributes []string, headers []kafka.Header) error {
+	allowed := esp.RaiseEventAllowed(resourceType, updateAttributes)
+	eventModelDao, err := dao.GetFromResourceType(resourceType)
+	if err != nil {
+		return err
+	}
+
+	resourceJSON, errEvent := (*eventModelDao).ToEventJSON(&resourceID)
```
lindgrenj6: Can we rename the error to `err`?

internal/events/event_stream_producer.go
```diff
@@ -0,0 +1,101 @@
...
+	if errEvent != nil {
+		return errEvent
+	}
+
+	err = esp.RaiseEventIf(allowed, resourceType+".update", resourceJSON, headers)
+	if err != nil {
+		return err
+	}
+
+	message, errMessage := m.UpdateMessage(eventModelDao, resourceID, resourceType, updateAttributes)
```
lindgrenj6: Same here.

model/events.go
```diff
@@ -0,0 +1,33 @@
...
+	updatedAttributes["updated"] = map[string]interface{}{resourceType: map[string]interface{}{strconv.Itoa(int(resourceID)): attributes}}
+
+	bulkMessage, errorBulkMessage := (*eventObject).BulkMessage(&resourceID)
```
lindgrenj6: Same here

lindgrenj6 (commented):
Just a few errors that need to be renamed - other than that looks good!

lpichler:
@lindgrenj6 PR rebased and err variables unified.

lindgrenj6 (approved):

PR #51, opened 2021-12-08
Test subpackages   in integration testing

lpichler:
- we were not testing submodules in integration testing by command `go test ./... --integration`
- it reveals couple places which requires changes:
   - Failing test `TestDefaultEndpointAlreadyExists`: 
     - Fix `Endpoints#CanEndpointBeSetAsDefaultForSource` checks for existence of records by **count** query
     - Extend test data for endpoints (`TestEndpointData`) so we can simulate that default endpoints with certain SourceId already exists (check is present in `Endpoints#CanEndpointBeSetAsDefaultForSource`)
- Tests were failing for submodules but arguments were not processed
   - error message: `flag provided but not defined: -integration` 
   - `testutils.ParseFlags()` was not used because of circle dependency error
   - some submodules required also `os.Exit(t.Run())`
- `go test ./... --integration` added to `pr_check.sh` 
- fixed sporadic failure most likely caused that tests for [different packages run in parallel](https://github.com/golang/go/blob/4df662fb373480b5055e645120558bb536fae42c/src/cmd/go/internal/test/test.go#L255)
  -  solution is that for each package is created own schema and thanks to that tests from various packages cannot affect each other 

### Sporadic failure Reproducer
- run db from container as we have [here](https://github.com/RedHatInsights/sources-api-go/blob/main/pr_check.sh#L33-L49) 
- run test multiple times:
```
go test ./... --integration -v
```
- it should fails sporadically and even each failing run fails thanks to different test

### DIFF
--- a/dao/endpoint_dao.go
+++ b/dao/endpoint_dao.go
@@ -2,6 +2,7 @@ package dao
 
 import (
 	"fmt"
+
 	"github.com/RedHatInsights/sources-api-go/middleware"
 	m "github.com/RedHatInsights/sources-api-go/model"
 )
--- a/dao/endpoint_dao.go
+++ b/dao/endpoint_dao.go
@@ -2,7 +2,6 @@ package dao
 
 import (
 	"fmt"
-
 	"github.com/RedHatInsights/sources-api-go/middleware"
 	m "github.com/RedHatInsights/sources-api-go/model"
 )
@@ -85,10 +84,10 @@ func (a *EndpointDaoImpl) Tenant() *int64 {
 
 func (a *EndpointDaoImpl) CanEndpointBeSetAsDefaultForSource(sourceId int64) bool {
 	endpoint := &m.Endpoint{}
+
 	// add double quotes to the "default" column to avoid any clashes with postgres' "default" keyword
 	result := DB.Where(`"default" = true AND source_id = ?`, sourceId).First(&endpoint)
-
-	return result.Error == nil
+	return result.Error != nil
 }
 
 func (a *EndpointDaoImpl) IsRoleUniqueForSource(role string, sourceId int64) bool {
--- a/internal/testutils/database.go
+++ b/internal/testutils/database.go
@@ -29,8 +29,6 @@ func ConnectToTestDB() {
 
 	// Set the dao.DB in case any tests want to use it
 	dao.DB = db
-	// Migrate the schema for the first time
-	MigrateSchema()
 }
 
 // CreateFixtures creates the following fixtures for the database —listed in order—:
--- a/internal/testutils/database.go
+++ b/internal/testutils/database.go
@@ -98,6 +98,8 @@ func DropSchema() {
 // MigrateSchema migrates all the models.
 func MigrateSchema() {
 	err := dao.DB.AutoMigrate(
+		&m.Tenant{},
+		
 		&m.SourceType{},
 		&m.ApplicationType{},
 
--- a/internal/testutils/database.go
+++ b/internal/testutils/database.go
@@ -10,13 +10,17 @@ import (
 	"github.com/labstack/gommon/log"
 	"gorm.io/driver/postgres"
 	"gorm.io/gorm"
+	"gorm.io/gorm/schema"
 )
 
 var testDbName = "sources_api_test_go"
 
 // ConnectToTestDB connects to the test database, populates the "dao.DB" member, and runs a schema migration.
-func ConnectToTestDB() {
-	db, err := gorm.Open(postgres.Open(testDbString(testDbName)), &gorm.Config{})
+func ConnectToTestDB(dbSchema string) {
+	db, err := gorm.Open(postgres.Open(testDbString(testDbName)), &gorm.Config{NamingStrategy: schema.NamingStrategy{
+		TablePrefix: dbSchema + ".",
+	}})
+
 	if err != nil {
 		log.Fatalf("db must not exist - create the database '%s' first with '-createdb'. Error: %s", testDbName, err)
 	}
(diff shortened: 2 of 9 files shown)


dao/endpoint_dao.go
```diff
@@ -84,11 +85,14 @@ func (a *EndpointDaoImpl) Tenant() *int64 {
...
-	endpoint := &m.Endpoint{}
+	var count int64
 	// add double quotes to the "default" column to avoid any clashes with postgres' "default" keyword
-	result := DB.Where(`"default" = true AND source_id = ?`, sourceId).First(&endpoint)
+	result := DB.Model(&m.Endpoint{}).Where(`"default" = true AND source_id = ?`, sourceId).Count(&count)
+	if result.Error != nil {
+		logging.Log.Errorf("cannot check whether endpoint can be set %s", result.Error.Error())
+	}
 
-	return result.Error == nil
```
MikelAlejoBR, line 91: Regarding this, wouldn't changing the condition to `result.Error != nil` do the job? I'm thinking that the way I wrote it, it was wrong because it was doing the following:

| Default endpoints in database | `.Find` returns result | `result.Error` is `nil` | `CanEndpointBeSetAsDefaultForSource`? | 
|-------------------------------|----------------------|---------------------|-------------------------------------|
| yes                           | yes                  | yes                  | true                                |
| no                            | no                   | no                 | false                               |

... when it should return the opposite thing. `true` when there are no default endpoints, and `false` when there is at least one. `.Count` might be easier to follow though...

internal/testutils/fixtures.go
```diff
@@ -27,9 +29,22 @@ var TestApplicationData = []m.Application{
 	{ID: 2, Extra: datatypes.JSON("{\"extra\": false}"), ApplicationTypeID: 1, SourceID: 1, TenantID: 1},
 }
 
+var (
+	CreatedAt, _ = time.Parse("2006-01-02T15:04:05.999Z", "2021-11-26 18:47:19 CET")
+	UpdatedAt    = CreatedAt
+	port         = 80
+	defaultValue = true
```
MikelAlejoBR: Here I'm noticing that both of the endpoints are created as "default". It seems that the table allows it, although we don't allow it when validating a new endpoint creation. Maybe we should set one of them to `false` and the other one to `true`? What do you think?

MikelAlejoBR:
I'm also thinking... maybe we could move the `internal/testutils/parser.go` file to `internal/parser/parser.go` or even create subdirectories for each of the utils, because if we do that, we will be able to reuse the `parser.go` utilities without hitting an import cycle:

```console
.
└── testutils
    ├── database
    │   └── database.go
    ├── fixtures
    │   ├── authentication.go
    │   ├── endpoint.go
    │   └── tenant.go
    ├── parser
    │   └── parser.go
    └── request
        └── request.go
```

.... although maybe this could be left for another PR.

dao/endpoint_dao.go
lindgrenj6: I'd also add that `.First()` does a `LIMIT 1` on the query vs `.Count()` doing a full table scan. I generally lean towards `.First()` and comparing the error instead.

@MikelAlejoBR I think you're right - we would need to inverse it.

lpichler: yes, that's right (also LIMIT 1 👍 ) I put it back with `result.Error != nil` . 


**NOT for this PR - my thoughts about it:**
```
func (a *EndpointDaoImpl) CanEndpointBeSetAsDefaultForSource(sourceId int64) bool {
	endpoint := &m.Endpoint{}

	// add double quotes to the "default" column to avoid any clashes with postgres' "default" keyword
	result := DB.Where(`"default" = true AND source_id = ?`, sourceId).First(&endpoint)
	return result.Error != nil
}
```

as I am looking on code - it is harder to understand (for me ) as it is just checking existence of record.
Maybe we can extract method (called "recordExists") and I believe this method can be used also in other places.

But maybe it is understandable enough as it is :)

internal/testutils/fixtures.go
lpichler: that's right and it was fixed 👍

lpichler:
> we could move the `internal/testutils/parser.go` file to `internal/parser/parser.go` or even create subdirectories for each of the utils, because if we do that, we will be able to reuse the `parser.go` utilities without hitting an import cycle:

interesting but yes let's do it in another PR

lpichler:
@MikelAlejoBR @lindgrenj6 thanks  for your review I fixed it and it could be re-reviewed 👍

dao/endpoint_dao.go
MikelAlejoBR: You are right, the naming of the function makes it hard to reason about — in fact I think that's why I got that return value the other way around. I think your idea of renaming it to `RecordExists`, `DefaultEndpointExists` or any other name will make it easier to understand :+1:

internal/testutils/database.go
```diff
@@ -147,3 +151,17 @@ func testDbString(dbname string) string {
...
+func ConnectAndMigrateDB(packageName string) {
+	ConnectToTestDB(packageName)
+
+	if out := dao.DB.Exec("CREATE SCHEMA " + packageName); out.Error != nil {
+		fmt.Println("error in creating schema " + out.Error.Error())
+	}
+
+	MigrateSchema()
+
+	if out := dao.DB.Exec("SET search_path TO \"$user\", " + packageName); out.Error != nil {
```
MikelAlejoBR: You might find it easier using a raw string here, as you can avoid escaping double quotes:

```diff
- dao.DB.Exec("SET search_path TO \"$user\", " + packageName);
+ dao.DB.Exec(`SET search_path TO "$user", ` + packageName);
```

internal/testutils/database.go
```diff
@@ -147,3 +151,17 @@ func testDbString(dbname string) string {
...
+		fmt.Println("error in setting schema" + out.Error.Error())
+	}
```
MikelAlejoBR: I would suggest using [`log.Fatalf`](https://pkg.go.dev/log#Fatalf) here, as I guess it makes sense just to end the program if the schema creation or the migrations don't work:

```go
import "log"

//...

if out := dao.DB.Exec("CREATE SCHEMA " + packageName); out.Error != nil {
	log.Fatalf("error in creating schema: %s", out.Error)
}

MigrateSchema()

if out := dao.DB.Exec(`SET search_path TO "$user", ` + packageName); out.Error != nil {
	log.Fatalf("error in setting schema: %s", out.Error)
}
```

lpichler:
@MikelAlejoBR, I added your suggestions.

in addition:
- I added `IF NOT EXIST` to CREATE SCHEMA
- I simplified `dropSchema` function

lpichler:
[this](https://github.com/RedHatInsights/sources-api-go/pull/51)  needs to be merged first.

lindgrenj6 (approved):
LGTM :+1:

MikelAlejoBR (approved):

PR #54, opened 2021-12-10
Add Message#GetHeader(string) method to fetch headers off of kafka message

lindgrenj6:
As mentioned here: https://github.com/RedHatInsights/sources-superkey-worker/pull/46

Added some tests around it and for a few other methods in the Message struct.


lpichler (approved):
LGTM 👍 thanks also for tests 👍

PR #55, opened 2021-12-13
[RHCLOUD-17247] Fetch tokens from marketplace and store them in an struct

MikelAlejoBR:
Implements a function that calls the marketplace endpoints to get a new token, and extracts the `access_token` and the `expiration` timestamp from the response, to return them in a `BearerToken` struct that we can use latter.

[[RHCLOUD-17247]](https://issues.redhat.com/browse/RHCLOUD-17247)

### DIFF
--- a/marketplace/marketplace_token_provider.go
+++ b/marketplace/marketplace_token_provider.go
@@ -14,7 +14,7 @@ type httpClient interface {
 }
 
 // GetToken sends a request to the marketplace to request a bearer token.
-func GetToken(httpClient httpClient, marketplaceHost string, apiKey string) (BearerToken, error) {
+func GetToken(httpClient httpClient, marketplaceHost string, apiKey string) (*BearerToken, error) {
 	// Reference docs for the request: https://marketplace.redhat.com/en-us/documentation/api-authentication
 	data := url.Values{}
 	data.Set("apikey", apiKey)
@@ -27,7 +27,7 @@ func GetToken(httpClient httpClient, marketplaceHost string, apiKey string) (Bea
 	)
 
 	if err != nil {
-		return BearerToken{}, fmt.Errorf("could not create the request object: %s", err)
+		return nil, fmt.Errorf("could not create the request object: %s", err)
 	}
 
 	// Set the proper headers to accept JSON, and let the server know we're sending urlencoded data.
@@ -36,11 +36,11 @@ func GetToken(httpClient httpClient, marketplaceHost string, apiKey string) (Bea
 
 	response, err := httpClient.Do(request)
 	if err != nil {
-		return BearerToken{}, fmt.Errorf("could not perform the request to the marketplace: %s", err)
+		return nil, fmt.Errorf("could not perform the request to the marketplace: %s", err)
 	}
 
 	if response.StatusCode != http.StatusOK {
-		return BearerToken{}, fmt.Errorf("unexpected status code received from the marketplace: %d", response.StatusCode)
+		return nil, fmt.Errorf("unexpected status code received from the marketplace: %d", response.StatusCode)
 	}
 
 	return DecodeMarketplaceTokenFromResponse(response)
--- a/marketplace/marketplace_token_provider.go
+++ b/marketplace/marketplace_token_provider.go
@@ -9,14 +9,23 @@ import (
 	"github.com/RedHatInsights/sources-api-go/config"
 )
 
-// httpClient abstracts away the client to be used in the GetToken function, and allows mocking it easily for the
+// GetHttpClient variable that holds the function which returns an HttpClient. This allows us to set up in runtime
+// which http client we want for the "GetToken" function, and allows us to mock it easily.
+var GetHttpClient func() HttpClient
+
+// HttpClient abstracts away the client to be used in the GetToken function, and allows mocking it easily for the
 // tests.
-type httpClient interface {
+type HttpClient interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
+// GetHttpClientStdlib returns a "http.Client" with a timeout of 10 seconds.
+func GetHttpClientStdlib() HttpClient {
+	return &http.Client{Timeout: 10}
+}
+
 // GetToken sends a request to the marketplace to request a bearer token.
-func GetToken(httpClient httpClient, apiKey string) (*BearerToken, error) {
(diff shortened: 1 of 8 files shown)


marketplace/token.go
```diff
@@ -0,0 +1,31 @@
...
+// BearerToken represents the bearer token sent by the marketplace, and includes the Unix timestamp of the time when
+// it expires.
+type BearerToken struct {
+	Expiration int    `json:"expiration"`
+	Token      string `json:"access_token"`
+}
+
+// DecodeMarketplaceTokenFromResponse decodes the bearer token and the expiration timestamp from the received
+// response.
+func DecodeMarketplaceTokenFromResponse(response *http.Response) (BearerToken, error) {
```
lindgrenj6: Can we change the return values to `(*BearerToken, error)` that way we can just `return nil, err`?

marketplace/marketplace_token_provider.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// httpClient abstracts away the client to be used in the GetToken function, and allows mocking it easily for the
+// tests.
+type httpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// GetToken sends a request to the marketplace to request a bearer token.
+func GetToken(httpClient httpClient, marketplaceHost string, apiKey string) (BearerToken, error) {
```
lindgrenj6: The change would bubble up here too - that way we dont have to pass around empty structs on error.

lindgrenj6: Since we're passing the interface in - should probably have a test for this too.

marketplace/marketplace_token_provider.go
```diff
@@ -0,0 +1,47 @@
...
+
+	if err != nil {
+		return BearerToken{}, fmt.Errorf("could not create the request object: %s", err)
+	}
+
+	// Set the proper headers to accept JSON, and let the server know we're sending urlencoded data.
+	request.Header.Add("Accept", "application/json")
+	request.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+
+	response, err := httpClient.Do(request)
```
lindgrenj6: Don't forget to add a `defer response.Body.Close()` on the end here - otherwise the response leaks.

lindgrenj6 (commented):
Just a few _super small_ changes - other than that looks good!

marketplace/token.go
MikelAlejoBR: Sure! Done.

marketplace/marketplace_token_provider.go
MikelAlejoBR: Done too!

MikelAlejoBR: Done as well!

marketplace/marketplace_token_provider.go
MikelAlejoBR: Nice catch! I did a manual `close` on the decoding function — which I think that makes more sense, as it's the one opening the response's body as well — and logged the error to catch potential leaks.

lpichler (approved):

MikelAlejoBR:
Rebased `main`

marketplace/marketplace_token_provider.go
```diff
@@ -0,0 +1,49 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/RedHatInsights/sources-api-go/config"
+)
+
+// httpClient abstracts away the client to be used in the GetToken function, and allows mocking it easily for the
+// tests.
+type httpClient interface {
```
lindgrenj6: So one last request here - can we add a `var` that contains the instance of httpClient, so we can just set it once rather than passing it around as an argument. [Kind of like we do with the DAOs](https://github.com/RedHatInsights/sources-api-go/blob/dcc99ae23f9d1295a7ce77fb2167913e6f15322b/authentication_handlers.go#L13) except instead of a function we can just make it a field. You should be able to set that variable to your mocked out clients in the tests.

Set it to `http.Client{Timeout:10*time.Second}` in `main.go` (like we do with the daos), it's [concurrency safe](https://pkg.go.dev/net/http#Client) according to docs so we can just have one client that will be used by potentially multiple threads at once. Maybe in the future we can extract it to a util var or something - but I'm not sure we'll need an http client elsewhere except here. RBAC implements its own in the library for example.

lindgrenj6 (commented):
Just one teeeeny change :pray:

marketplace/marketplace_token_provider.go
MikelAlejoBR: All righty. I did it, but I had to make the function to change the `var` exported, to be able to initialize it from the `main` function.

marketplace/token.go
```diff
@@ -0,0 +1,37 @@
...
+	token := BearerToken{}
+
+	err := json.NewDecoder(response.Body).Decode(&token)
+	if err != nil {
+		return nil, err
+	}
+
+	err = response.Body.Close()
+	if err != nil {
+		log.Fatalf("could not close the marketplace JWT token response's body: %s", err)
```
lindgrenj6: I actually don't think we should fatal log here - maybe return an error instead and make that return a 500 response in the controller later?

marketplace/token.go
```diff
@@ -0,0 +1,37 @@
+package marketplace
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
```
lindgrenj6: Should have caught this earlier - but we should be using the `logging` package instead of the std logger - that way it'll get thrown at cloudwatch and formatted as json.

lindgrenj6 (commented):
Sorry just one more change that I should have noticed earlier but missed. We need to be using the logging package instead of the stdlib logger - that way it formats the logs correctly and sends them to cloudwatch.

marketplace/token.go
MikelAlejoBR: Whoops, you are right. I was intending to use `fmt.Errorf` there but I messed up. Nice catch!

marketplace/token.go
MikelAlejoBR: Yeah this is a mistake related to the previous one. I've removed this from here. I guess that we'll use `logging` extensively on the middleware or on the handlers to return user friendly messages, and log the error messages.

lindgrenj6 (approved):

PR #57, opened 2021-12-21
[RHCLOUD-17248] Marketplace token caching

MikelAlejoBR:
This merge request implements caching the marketplace token on Redis.

[[RHCLOUD-17248]](https://issues.redhat.com/browse/RHCLOUD-17248)

### DIFF
--- a/marketplace/token.go
+++ b/marketplace/token.go
@@ -9,7 +9,7 @@ import (
 // BearerToken represents the bearer token sent by the marketplace, and includes the Unix timestamp of the time when
 // it expires.
 type BearerToken struct {
-	Expiration *int    `json:"expiration"`
+	Expiration *int64  `json:"expiration"`
 	Token      *string `json:"access_token"`
 }
 
--- a/marketplace/token.go
+++ b/marketplace/token.go
@@ -13,6 +13,16 @@ type BearerToken struct {
 	Token      *string `json:"access_token"`
 }
 
+// MarshalBinary implements the "BinaryMarshaller" interface to easily marshal the struct when using the Redis client.
+func (bt BearerToken) MarshalBinary() (data []byte, err error) {
+	return json.Marshal(bt)
+}
+
+// String pretty prints the struct. Required to avoid printing the actual pointer addresses.
+func (bt *BearerToken) String() string {
+	return fmt.Sprintf(`BearerToken{Expiration:%d, Token:"%s"}`, *bt.Expiration, *bt.Token)
+}
+
 // DecodeMarketplaceTokenFromResponse decodes the bearer token and the expiration timestamp from the received
 // response.
 func DecodeMarketplaceTokenFromResponse(response *http.Response) (*BearerToken, error) {
--- a/redis/marketplace_token.go
+++ b/redis/marketplace_token.go
@@ -1,12 +1,31 @@
 package redis
 
 import (
+	"encoding/json"
 	"fmt"
-	"github.com/RedHatInsights/sources-api-go/marketplace"
 	"time"
+
+	"github.com/RedHatInsights/sources-api-go/marketplace"
 )
 
 var redisKeySuffix = "marketplace_token_%d"
+
+// GetToken fetches the token from the Redis cache.
+func GetToken(tenantId int64) (*marketplace.BearerToken, error) {
+	redisKey := fmt.Sprintf(redisKeySuffix, tenantId)
+
+	cachedToken, err := Client.Get(redisKey).Result()
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch the token from Redis: %s", err)
+	}
+
+	token := &marketplace.BearerToken{}
+	if err = json.Unmarshal([]byte(cachedToken), &token); err != nil {
+		return nil, fmt.Errorf("could not unmarshal the cached token: %s", err)
+	}
+
+	return token, err
+}
 
 // SetToken sets the token on the Redis cache.
 func SetToken(tenantId int64, token *marketplace.BearerToken) error {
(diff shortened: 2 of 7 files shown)


lpichler:
CI failure
14:55:02 	imports github.com/RedHatInsights/sources-api-go/internal/testutils: no matching versions for query "latest"
maybe `go mod tidy` could fix it

redis/marketplace_token.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// MarketplaceTokenCacher is an struct which implements the "TokenCacher" interface. This helps in abstracting away the
+// dependencies and making testing easier.
+type MarketplaceTokenCacher struct {
+	TenantID int64
+}
+
+// FetchToken fetches the token from the Redis cache.
+func (mtc *MarketplaceTokenCacher) FetchToken() (*marketplace.BearerToken, error) {
```
lindgrenj6, line 29: Can we add some logging here? That way we have some insight when we're hitting redis.

redis/marketplace_token.go
```diff
@@ -0,0 +1,63 @@
...
+	err := Client.Set(
+		redisKey,
+		token,
+		redisExpiration,
+	).Err()
+
+	if err != nil {
+		return fmt.Errorf(
+			"could not set marketplace token on redis: %s", token,
+		)
```
lindgrenj6: I know this is a super small nit - but can this be on one line? (following the rest of the return calls)

redis/marketplace_token.go
```diff
@@ -0,0 +1,63 @@
...
+	TenantID int64
+}
+
+// FetchToken fetches the token from the Redis cache.
+func (mtc *MarketplaceTokenCacher) FetchToken() (*marketplace.BearerToken, error) {
+	redisKey := fmt.Sprintf(redisKeySuffix, mtc.TenantID)
+
+	cachedToken, err := Client.Get(redisKey).Result()
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch the token from Redis: %s", err)
```
lindgrenj6: Chewing on this a bit - I _think_ redis will return err if the token doesn't exist, e.g. it expired as well. Maybe this error message could be `token not present in redis:`

marketplace/token.go
```diff
@@ -9,10 +9,20 @@ import (
 // BearerToken represents the bearer token sent by the marketplace, and includes the Unix timestamp of the time when
 // it expires.
 type BearerToken struct {
-	Expiration *int    `json:"expiration"`
+	Expiration *int64  `json:"expiration"`
 	Token      *string `json:"access_token"`
 }
 
+// MarshalBinary implements the "BinaryMarshaller" interface to easily marshal the struct when using the Redis client.
+func (bt BearerToken) MarshalBinary() (data []byte, err error) {
```
lpichler, line 17: Where this is used ?

lindgrenj6: I think it's used here: https://github.com/RedHatInsights/sources-api-go/pull/57/files#diff-9f23ee963589d96d28d468ae7ce847cd2dcd6631c19b9d9fc4ce3be971f3cd3cR52

If the struct implements the `BinaryMarshaller` interface redis will use that method to marshal it (per the comment).

redis/marketplace_token.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return token, err
+}
+
+// CacheToken sets the token on the Redis cache.
+func (mtc *MarketplaceTokenCacher) CacheToken(token *marketplace.BearerToken) error {
+	redisKey := fmt.Sprintf(redisKeySuffix, mtc.TenantID)
+
+	tokenExpiration := time.Unix(*token.Expiration, 0)
+	redisExpiration := time.Until(tokenExpiration)
```
lpichler, line 52: when `token.Expiration` is older then `time.Now()` then `redisExpiration` is negative value

 `Client.Set` is treating with negative value as well as with zero - It means when we set token with expired expiration - we will be able to retrieve token.

I am not sure why we would put expired token to Redis :) but maybe it will be better cover when redisExpiration has negative value - maybe don't set it to Redis at all ?

lpichler:
@MikelAlejoBR I had one comment - otherwise - tested and it  works, LGTM 👍

MikelAlejoBR:
Aaaaaaah my bad. I have force pushed when I should have cherry-picked from the main branch, so that your review status doesn't get lost... Sorry about that.

If I recall correctly, the tests were failing because this had to be included in this PR: https://github.com/RedHatInsights/sources-api-go/pull/60 — refactor: replace "testutils" with "testutils/parser".

marketplace/token.go
MikelAlejoBR: Exactly. Actually, Redis requires us to implement that `BinaryMarshaller` interface to be able to cache the token.

redis/marketplace_token.go
MikelAlejoBR: Sure! I added the tenant ID to the message as well.

redis/marketplace_token.go
MikelAlejoBR: Done. I don't know why I did this. My guess is that I tried printing the bearer token's fields manually, and when I finally implemented the `String()` interface for the `BearerToken` type I forgot to turn this back to a "oneliner".

redis/marketplace_token.go
MikelAlejoBR: Yeah, I like it better. It's also more concise.

redis/marketplace_token.go
MikelAlejoBR: Nice catch! I added a check to avoid these situations.

lpichler (approved):
LGTM!

lindgrenj6 (approved):

PR #62, opened 2022-01-05
move tenancy + permission check authorization to `middleware/` package

lindgrenj6:
Been wanting to do this for a bit - but the problem became that I stuck the `Filter` struct in the middleware package. By moving it to `util/` that allowed me to finally put all our middleware in the right pacakge including the tenancy check as well as authorization check. 

Most of the changes are just package renames - best to just check out the changes under `middleware/`

I'll be following this up with changes to allow the system header through - if it contains a `cluster_id` or a `cn` field.

**EDIT**:  I also added the current sources-api behavior to the authorization check - e.g. if there is a system header present we just let the request through.

### DIFF
--- a/application_handlers_test.go
+++ b/application_handlers_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/request"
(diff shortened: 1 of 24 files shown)


MikelAlejoBR (approved):

lpichler (approved):
LGTM

PR #63, opened 2022-01-05
[RHCLOUD-16316] Deploy Availability Status Listener

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-16316

Adding the pod for the availability status listener to the deployment. Also did some tiny refactors on the listener for handling case-insensitive resource names.

I ended up setting the availability status listener min replicas to 0 so we can merge this - we can scale it up and turn off the ruby-side at the same time at a later date. 

~~WIP as I need to test it yet and due to my internet being out I can't.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,9 @@ run: build
 inlinerun:
 	go run `ls *.go | grep -v test`
 
+listener:
+	go run `ls *.go | grep -v test` -listener
+
 container:
 	docker build . -t sources-api-go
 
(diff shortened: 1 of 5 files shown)


MikelAlejoBR:
I would suggest rebasing https://github.com/RedHatInsights/sources-api-go/pull/84, as it seems that we've introduced another `TestMain` function which clashes with the previous one that already existed.

MikelAlejoBR (approved):
LGTM!

lpichler (approved):

lpichler:
Can I merge this now ?

lindgrenj6:
@lpichler yes - I have it scaled to zero by default here: https://github.com/RedHatInsights/sources-api-go/pull/63/files#diff-50b41779d45e37968dc1bef161feebad3447f6c3cf72cfee62db774ecd5dfa4dR168

PR #65, opened 2022-01-06
Lock down xrhid system authorization to only GET/POST

lindgrenj6:
Also added a bunch of tests for it + mocked out the rbac tests.

### DIFF
--- a/middleware/authorization.go
+++ b/middleware/authorization.go
@@ -14,7 +14,11 @@ import (
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
-var PSKS = config.Get().Psks
+var (
+	psks            = config.Get().Psks
+	bypassRbac      = config.Get().BypassRbac
+	rbacClient Rbac = &RbacClient{client: rbac.NewClient(os.Getenv("RBAC_URL"), "sources")}
+)
 
 /*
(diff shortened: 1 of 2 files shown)


middleware/authorization.go
```diff
@@ -50,10 +54,28 @@ func PermissionCheck(next echo.HandlerFunc) echo.HandlerFunc {
 				return fmt.Errorf("error casting identity to struct: %+v", c.Get("identity"))
 			}
 
-			// current sources-api behavior = if there is a system key -> it's authorized.
-			// TODO: make this more specific and do more checks.
+			// checking to see if we're going to change the results since
```
lpichler, line 57: Maybe follow-up but do you think that this logic could be refactored somehow ?

it is well described but there is lot of ifs, switches - it can be hard understand later.

lpichler (approved):
LGTM 👍

middleware/authorization.go
lindgrenj6: Yeah I'm still playing with it - in the process of refactoring a bit still, I can do it in a followup pr though thats nbd.

lindgrenj6: Ended up opening another PR after this one #66, this will probably still get refactored later.

lpichler (approved):
👍

PR #67, opened 2022-01-07
Update UBI to RHEL 8.5 in order to bring in Go 1.16

lindgrenj6:
A few things got moved from `ioutil` to just the `io` and `os` packages. We should use these going forward. 

I also added a container build to the pr_check.sh script - that way even if the action build works the container one might fail due to differing go versions.


lpichler (approved):
LGTM

MikelAlejoBR (approved):

PR #68, opened 2022-01-10
refactor: use 405 status code for invalid http verbs in requests

MikelAlejoBR:
It replaces the "401 Unauthorized" responses with a more accurate "405 Method not allowed" response.


lpichler (approved):

lindgrenj6 (approved):

lindgrenj6:
Thanks for this - probably need to update the API spec eventually to say we can return a 405 - but thats less important.

PR #69, opened 2022-01-10
Add Static Cache for Application/Source Type Id's

lindgrenj6:
This PR adds the static cache we discussed in Slack today. This is what it looks like in memory:
image

---

I have comments in the struct but will add more here. Basically it caches the `SourceType` name -> id mapping as well as the `ApplicationType` name -> id mapping. For ApplicationType though I did not only the path-style name but also the "short" name since we use that in some places in the app.

---

I also have another commit as an example of it being used: limiting satellite listing sources to _only_ the satellite source_type_id.

\# TODO:
- [x] Tests for SourceList with cert-auth being used
- [x] Maybe tests for the typeCache?

### DIFF
--- a/dao/main_test.go
+++ b/dao/main_test.go
@@ -1,0 +1,15 @@
+package dao
+
+import (
+	"os"
+	"testing"
+
+	"github.com/RedHatInsights/sources-api-go/internal/testutils/parser"
+)
+
+func TestMain(t *testing.M) {
+	// we need this to parse arguments otherwise there are not recognized which lead to error
+	_ = parser.ParseFlags()
+
+	os.Exit(t.Run())
+}
--- a/dao/source_type_dao.go
+++ b/dao/source_type_dao.go
@@ -8,7 +8,7 @@ import (
 type SourceTypeDaoImpl struct {
 }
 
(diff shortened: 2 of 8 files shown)


MikelAlejoBR (approved):

lpichler (approved):

lpichler:
@lindgrenj6 as this add test to `dao` package, the package most likely needs main_test.go, for example:


```
package dao

import (
	"os"
	"testing"
)

func TestMain(t *testing.M) {
	// we need this to parse arguments otherwise there are not recognized which lead to error
	_ = parser.ParseFlags()

	os.Exit(t.Run())
}
```

lpichler (changes requested):
@lindgrenj6 as this add test to dao package, the package most likely needs main_test.go, for example:
```go
package dao

import (
	"os"
	"testing"
)

func TestMain(t *testing.M) {
	// we need this to parse arguments otherwise there are not recognized which lead to error
	_ = parser.ParseFlags()

	os.Exit(t.Run())
}
```

lindgrenj6:
@lpichler i swear i added it - must have gotten lost :grin: added either way!

lpichler (approved):

PR #72, opened 2022-01-12
[RHCLOUD-15003] Feature migrations

MikelAlejoBR:
- Uses `golang-migrate`(https://github.com/golang-migrate)
- Creates a new `schema_migrations_go` table to store the 
- Migrates all the `sources-api/db/migrate/*.rb` files to SQL files
    -  Replaces the `yyyy-MM-ddHHMMSS_${description}` format with a sequential `1_${description}` one. The reason for this is that `golang-migrate` stores the versions as an `int8` instead of the `varchar`s that Rails stores. So to make the table contents easier to read, instead of storing gigantic integers, simple-small numbers are stored.
    - Also, in order not to incur in "already exists" errors, some safeguards have been implemented. These check on the prior existence of the tables and foreign key constraints before creating the tables. This way, if the first migration is applied in an already populated database, no error will be thrown, and nothing will be done.

## Links

[RHCLOUD-15003](https://issues.redhat.com/browse/RHCLOUD-15003)


lindgrenj6:
I have a feature request for this: can we support a `-setup` and `-reset` flags where:
- `-setup` would create the db + migrate the db
- `-reset` would drop the db, then run the setup task.

LMK what you think!

MikelAlejoBR:
I added that feature to the code. Now:

* If you pass the `-setup` argument, the schema is migrated.
* If you pass the `-reset` argument, the schema is dropped, recreated and migrated.
* If you don't pass anything, the app just boots.

lpichler:
1) Is this also related to gorm `AutoMigrate`  call ? Shouldn't we stop using it ? 
2) For example if I want to add new column - I need to create sql file in db/migrations/ ?
3) When we will have data migration, is there a way how to test them ?

anyway this is great work 👏

lindgrenj6:
@lpichler 
>Is this also related to gorm AutoMigrate call ? Shouldn't we stop using it ?

AutoMigrate is fine for tests - but we can't/shouldn't use it for prod since we can't control how the migrations work. I think we'll keep it for tests rather than using the actual migrations.

>For example if I want to add new column - I need to create sql file in db/migrations/ ?

Yes, thats how it would work. Basically our migrations would just be sql migrations named appropriately.

>When we will have data migration, is there a way how to test them ?

That would probably be all done locally and spray/pray on stage. Data migrations are tough here though - since it is all SQL we don't have the power of AR when doing actual data migrations. I know there is a [gorm-migrate](https://github.com/go-gormigrate/gormigrate) project but it doesn't have much activity it seems. It would be nice to have the entirety of GORM during migrations I will admit. Maybe we should look into that? 

Opinions?

MikelAlejoBR:
> @lpichler
> 
> > Is this also related to gorm AutoMigrate call ? Shouldn't we stop using it ?
> 
> AutoMigrate is fine for tests - but we can't/shouldn't use it for prod since we can't control how the migrations work. I think we'll keep it for tests rather than using the actual migrations.

In fact I tried to refactor how the schema is migrated in tests, but I was facing some issues with the relative paths of the SQL files, so I gave up for now. I wouldn't mind having our tests migrating the schema every time, so that the migrations get tested too.

However I'm thinking on doing that on a follow-up PR as that would require to refactor multiple things and I wouldn't like to clutter this PR —which is fairly big to review already—.

> > When we will have data migration, is there a way how to test them ?
> 
> That would probably be all done locally and spray/pray on stage. Data migrations are tough here though - since it is all SQL we don't have the power of AR when doing actual data migrations. I know there is a [gorm-migrate](https://github.com/go-gormigrate/gormigrate) project but it doesn't have much activity it seems. It would be nice to have the entirety of GORM during migrations I will admit. Maybe we should look into that?
> 
> Opinions?

When I took a look at `gormigrate` there were two things that I didn't like: it seems to be stalled —last closed issue on 2021-01-02, last non-automatically merged PR on 2021-06-13— and also it uses the GORM API to create migrations, instead of raw SQL. Plus, [they recommend `golang-migrate/migrate` for non-small projects](https://github.com/go-gormigrate/gormigrate#who-is-gormigrate-for).

Personally I really like SQL migrations since that allows us to decouple the migrations from the code that runs them. If we have all the migrations in SQL, today we can use one migration tool, but tomorrow we could use a different one which has many more complex features like Flyway —which isn't even written in Go!—, and the only thing we would have to do is adapt the naming of the files a little bit. But the migrations themselves would be left untouched, and would keep working, regardless of the library.

I personally liked `golang-migrate/migrate` because first of all it provides really basic up/down migration features that expect SQL files. I chose to perform the migrations programmatically —the app runs them every time it boots—, but the migration tool can be run with their CLI if we want to further decouple the migration tool from the project itself, so that's an option too. Finally it looks very active and overall feels like a simple migration tool with the features we need.

I am not sure I fully understand what you mean by "testing data migrations". Are you referring to testing whether the SQL migration does what we expect it to do? Because in that case, it occurs to me that we could use `golang-migrate/migrate`'s [`Migrate(version uint)`](https://pkg.go.dev/github.com/golang-migrate/migrate?utm_source=godoc#Migrate.Migrate) function, which would help creating tests to check the "before" and "after" changes, and see if the schema and the data it contains has been properly migrated.

lindgrenj6:
I think we're talking less about actually testing the migrations more like "how do we modify data if we need to do something other than SQL".

Personally I like them being SQL as well but there might come a time where we need to modify the data in a way that writing some SQL can't quite do for us. The answer might just be we don't have the ability to transform data like we did with ActiveRecord migrations and that just might be the way it is. 

Rails example: https://github.com/RedHatInsights/sources-api/blob/d2b5af911fbcd066ad4e6d27901c594428c8c95c/db/migrate/20211012173623_move_authentications_to_vault.rb

MikelAlejoBR:
Oh, that's a really good point... I see why the "language" migrations instead of the SQL ones make sense... :thinking:

lindgrenj6:
Yeah - it's tough! They're so nice! 

-- 
Jacob Lindgren

Red Hat
***@***.***     M: 402-841-6173

lpichler:
I like GORM in migrations - it gives us some benefits and we can also run raw sql with GORM and @lindgrenj6 provided good example why we should prefer it.

@MikelAlejoBR can we use https://github.com/golang-migrate/migrate with GORM somehow ?

> It would be nice to have the entirety of GORM during migrations I will admit. Maybe we should look into that?

I agree with this, maybe we should start to work on it because we will need it in the end.

I am OK with having AutoMigrate for tests.

MikelAlejoBR:
All right, I've refactored the PR so that we use "gormigrate". The first migration just reads the embedded SQL file because initially I was thinking that I could just translate the schema snapshot to "GORMisms", and make everything fit together. However I've decided to move forward and to finish the PR, and then ask you about it: what do you think? Do I try to translate the SQL to GORM, or do I leave it as is?

Thanks!

dao/db.go
```diff
@@ -41,6 +44,19 @@ func Init() {
 	}
 	rawDB.SetMaxOpenConns(20)
 
+	// Perform database migrations
+	if conf.MigrationsReset {
+		DB.Exec(`DROP SCHEMA "public" CASCADE`)
+		DB.Exec(`CREATE SCHEMA "public"`)
+	}
+
+	if conf.MigrationsSetup || conf.MigrationsReset {
```
lindgrenj6: I think I want the `setup` flag to just create the db - kinda like we do for `sources_api_test`. 

I want the migrations to just always run every time - since the app theoretically shouldn't be running if the db isnt migrated completely. This also prevents us from having to have an initContainer to run the migrations.

lindgrenj6:
I like having the "initial" migration just be in sql - since this is the biggest one too. 

---

The one thing that I think we may have to think about now is when we have multiple replicas running they'll try migrating all at once - we may have to come up with some form of lock. 

Just spitballing here - we could use a redis key to lock. e.g. the first pod that comes up would set the MIGRATION_LOCK key or something - then any others that came up would check for that key first before running migrations. Then when the first one gets done it would remove the key allowing the other pods to continue booting. 

The hard part is failing out - e.g. if the migration fails. Maybe an initContainer would be better for that case then which nullifies my previous comment. Open to ideas here!

MikelAlejoBR:
To avoid any race conditions there exists the a library called ["redsync"](https://github.com/go-redsync/redsync) which implements the algorithm of the ["Distributed locks with Redis"](https://redis.io/topics/distlock) blog post written by the Redis community themselves.

We could set the distributed lock on Redis and make all the pods but one wait, and after that first one runs the migrations the other ones will just skip them.

---

Regarding what happens upon failure: I am wrapping all the migrations in transactions, since PostgreSQL supports ["Transactional DDLs"](https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis#Transactional_DDL), so in case any migration fails the changes are not committed. We could simply `os.Exit(1)` the program so that the pod enters a fail state, after logging what went wrong with the migration.

---

I don't really see it clear if we should use an init container or not. If we do use it and we end up in a "fine but not fine" situation with a migration —it ran fine, but we realized afterwards that for whatever reason it doesn't do what we expected it to do—, we would simply have change the SHA of the App to the previous one whilst keeping the initContainer's SHA as is —so that we can do a rollback with the `down()` code—.

If we do it without an init container we would have to first run the app again with the `--rollback <id>` so we don't lose the last migration's `down()` code and then change the SHA of the app and remove the `--rollback <id>` argument.

It seems like not using an init container is less steps, but I have the feeling that I'm missing something :thinking:

lindgrenj6:
As we discussed yesterday - the "expected" behavior is that the container can just sit there and crashloop on a bad/buggy migration. The possibilities are:
- A bad migration that bugs out will hold us up until a commit to fix the is introduced. Lets say we made an error based on the state of the db, and we just need to fix the migration that was bugging out
- In the case of a migration completing, but we don't like the results (bad data migration for example) we do *not* want to rollback the pod and have the `down()` migrations run. We would instead introduce a new migration moving the schema forward while also fixing the issue

Just wanted to get this down on this PR while its fresh in our minds. 

---

So I think about the only thing here is locking the db while migrations are running. I was debating changing our k8s deployment rollout strategy so basically k8s would just spin one pod down at a time, and thus the migrations would only run that first time. The second, third, Nth pod would not re-migrate. But I'd like to have some form of locking either way. 

I like the idea of locking via Redis - but am not sure I want to pull in a library that basically just:
1. `GET`s for the mutex key in redis, if it's there bail out (or retry every 10s or something)
2. If it isn't, does a `SET` and then proceeds to migrate
3. and does a `DEL` when migrations complete (or errors out, just before `return err` basically)

It's something pretty simple we could just do ourselves in the redis package without pulling in another library (and one less thing for the prodsec team to audit us on).

MikelAlejoBR:
> So I think about the only thing here is locking the db while migrations are running. I was debating changing our k8s deployment rollout strategy so basically k8s would just spin one pod down at a time, and thus the migrations would only run that first time. The second, third, Nth pod would not re-migrate. But I'd like to have some form of locking either way.
> 
> I like the idea of locking via Redis - but am not sure I want to pull in a library that basically just:
> 
>     1. `GET`s for the mutex key in redis, if it's there bail out (or retry every 10s or something)
> 
>     2. If it isn't, does a `SET` and then proceeds to migrate
> 
>     3. and does a `DEL` when migrations complete (or errors out, just before `return err` basically)
> 
> 
> It's something pretty simple we could just do ourselves in the redis package without pulling in another library (and one less thing for the prodsec team to audit us on).

According to the post I linked above, that works if we do have a _"non-distributed system composed of a single, always available, instance"_ and _"where a race condition from time to time is acceptable"_, if we use replicas. Otherwise they recommend using the Redlock algorithm for a proper implementation of a distributed mutex.

The thing is that I don't think that having a race condition when running migrations is acceptable, that's why I was suggesting using the library.

MikelAlejoBR:
All right, I included the locking mechanism for the migrations.

MikelAlejoBR:
All right, I have splitted this PR in two hoping that everything becomes easier to review:

* https://github.com/RedHatInsights/sources-api-go/pull/174
* https://github.com/RedHatInsights/sources-api-go/pull/175

main.go
```diff
@@ -19,9 +21,19 @@ var conf = config.Get()
 func main() {
 	logging.InitLogger(conf)
 
-	dao.Init()
+	setUpDatabase := flag.Bool("setup", false, "create the schema and run the migrations")
+	resetDatabase := flag.Bool("reset", false, "drop the schema, recreate it and run the migrations")
+
+	flag.Parse()
```
lindgrenj6: Final comment: we can move these over to the `config/` package here: https://github.com/RedHatInsights/sources-api-go/blob/62ea4f6533db5dbddb5d5dba4682f6045eca432b/config/config.go#L111

That way they only get parsed once and the config struct has them (like you do below).

lindgrenj6 (approved):
Ok this looks good now. I just have one tiny comment moving the flags to the config package since I was trying to get all of our runtime configuration over there. 

---

@lpichler I am having Mikel split this into multiple PRs so we can get it merged so we can (hopefully) do our first migration over here on the Go side for the `org_id` migration. With the PR in its current state we'll temporarily have 2 migration tables, one for Go and one for Rails. 

After we get done removing the Rails app completely we'll add in the migrations from #175 to remove the rails migration AND rename all the foreign keys from the rails generated to more human readable names. 

---

Please test locally - I did and it appears that it didn't effect any of the current tables than creating a `migrations` table that stores the migrations that have been ran.

main.go
MikelAlejoBR: Done!

dao/db.go
```diff
@@ -46,8 +46,15 @@ func Init() {
 
 	// Perform database migrations
 	if conf.MigrationsReset {
-		DB.Exec(`DROP SCHEMA "public" CASCADE`)
-		DB.Exec(`CREATE SCHEMA "public"`)
+		err = DB.Exec(`DROP SCHEMA "public" CASCADE`).Error
```
lpichler: you have in setup drop database  - Should not be here also dropping database ?

dao/db.go
```diff
@@ -46,8 +46,15 @@ func Init() {
...
+		if err != nil {
+			logging.Log.Fatalf(`Error dropping the "public" schema: %s`, err)
+		}
+
+		err = DB.Exec(`CREATE SCHEMA "public"`).Error
```
lpichler: same here

lpichler:
I have suggestions for feature requests 😊 (in follow up PRs):
`--rollback-to=<MIGRATION_ID_VERSION>` - it goes back to migration with stated ID 
```
example:
./sources-api-go --rollback-to=20220222110000
```

`--rollback-steps=<STEPS>` - it performs  count <STEPS> migrations you want to go back

`--generate-migration=<NAME>` - generate new empty migration with new ID - where <NAME> is name of migration file and migration function 

let me know if you consider any of them useful and we can add it ;) (feel free to change names of flags)

lpichler:
it looks good that `setup` and `reset` need change otherwise looks good 👍 👏

MikelAlejoBR:
Regarding the review comments: I did not write the "drop the database logic" because that implies opening a new connection and issuing the command, because Postgres doesn't allow dropping a database you're currently connected to. I've pushed the code so that the `-reset` flag drops the database.

> I have suggestions for feature requests blush (in follow up PRs): `--rollback-to=<MIGRATION_ID_VERSION>` - it goes back to migration with stated ID
> 
> ```
> example:
> ./sources-api-go --rollback-to=20220222110000
> ```
> 
> `--rollback-steps=<STEPS>` - it performs count migrations you want to go back
> 
> `--generate-migration=<NAME>` - generate new empty migration with new ID - where is name of migration file and migration function
> 
> let me know if you consider any of them useful and we can add it ;) (feel free to change names of flags)

`gormigrate` has a function which allows to do the rollback the way you are describing it: [RollbackTo](https://pkg.go.dev/github.com/go-gormigrate/gormigrate/v2#Gormigrate.RollbackTo), so it shouldn't be too hard to implement.

The same for `generate-migration`, we could have a template and make the command copy and paste it.

However as you say: I would leave that for follow up PRs. Thanks for the review :+1:

dao/db.go
```diff
@@ -41,6 +45,61 @@ func Init() {
...
+
+		// Recreate the database.
+		createDb := fmt.Sprintf(`CREATE DATABASE %s`, conf.DatabaseName)
+		err = adminConnection.Exec(createDb).Error
+		if err != nil {
+			log.Fatalln(err)
+		}
+
+		// Log and exit so that the application can be rerun without the "reset" flag.
+		logging.Log.Info(`The schema "public" has been reset`)
```
lpichler: yes it is true but regard to last changes - I am suggesting change it to "...database XX has been reset"

config/config.go
```diff
@@ -109,13 +111,17 @@ func Get() *SourcesApiConfig {
 	// Parse any Flags (using our own flag set to not conflict with the global flag)
 	fs := flag.NewFlagSet("runtime", flag.ContinueOnError)
 	availabilityListener := fs.Bool("listener", false, "run availability status listener")
+	setUpDatabase := fs.Bool("setup", false, "create the schema and run the migrations")
```
lpichler: schema -> database

config/config.go
```diff
@@ -109,13 +111,17 @@ func Get() *SourcesApiConfig {
...
+	resetDatabase := fs.Bool("reset", false, "drop the schema, recreate it and run the migrations")
```
lpichler: schema -> database

lpichler:
@MikelAlejoBR thanks for adding changes and it would be good if `setup` and `reset` could work even when database `ENV['DATABASE_NAME']` doesn't exist - it fails here https://github.com/MikelAlejoBR/sources-api-go/blob/feature-migrations/dao/db.go#L37

MikelAlejoBR:
I've modified the config so that if the `DATABASE_NAME` environment is missing, a default "sources_api_development" database name is used.

MikelAlejoBR:
> @MikelAlejoBR thanks for adding changes and it would be good if `setup` and `reset` could work even when database `ENV['DATABASE_NAME']` doesn't exist - it fails here https://github.com/MikelAlejoBR/sources-api-go/blob/feature-migrations/dao/db.go#L37

Done! Now you should be able to create/reset the database without any problems! :smiley:

lpichler:
@MikelAlejoBR thanks! it works, nicely done

1) Do we want also to call migrate in `setup` and `reset` ? as it was stated [here](https://github.com/RedHatInsights/sources-api-go/pull/72#issuecomment-1030258726) 

2) `--setup` - Do we want to call migrate when DB is already created ? 

3) Maybe add some log message when DB was successfully migrated ? 
 (we can also introduce a convention that we will use log messages in each migration, like "migration XXXX....started", "migration XXXX....started" )

MikelAlejoBR:
> @MikelAlejoBR thanks! it works, nicely done
> 
>     1. Do we want also to call migrate in `setup` and `reset` ? as it was stated [here](https://github.com/RedHatInsights/sources-api-go/pull/72#issuecomment-1030258726)

Initially I did set it up so that the program would also migrate the database, but then [I was told otherwise](https://github.com/RedHatInsights/sources-api-go/pull/72#discussion_r820977824):

> I think I want the `setup` flag to just create the db - kinda like we do for `sources_api_test`.
> 
> I want the migrations to just always run every time - since the app theoretically shouldn't be running if the db isnt migrated completely. This also prevents us from having to have an initContainer to run the migrations.

Maybe I understood it wrong, but what I get from here is that the commands should simply do their thing and terminate, and that the migrations should run every time the application launches.

However I don't mind changing that. Whatever is most useful for you.

>     2. `--setup` - Do we want to call migrate when DB is already created ?

Per the previous comment, my guess is that we should not, but... :eyes: 

>     3. Maybe add some log message when DB was successfully migrated ?
>        (we can also introduce a convention that we will use log messages in each migration, like "migration XXXX....started", "migration XXXX....started" )

I included a couple of logging statements on the migration. Let me know what you think :+1:

lindgrenj6:
Yeah @MikelAlejoBR you understood it right - to add a bit here we would _run the migrations every time at application startup_, so when `--setup` would be passed, the db would be created, the app would exit, then the app would migrate itself when you run it again. 

in the case of `--reset` it would drop/recreate the db, then migrate itself next startup once again. 

This is basically so we don't have to have an initContainer that basically run `/sources-api-go --migrate` or something like that. It should just do it at startup like we do with the seeding. 

---

I also like the idea of logs, the more logs the better.

lpichler (approved):

PR #73, opened 2022-01-12
Add Create/Edit/Delete operations for the Application Model

lindgrenj6:
Noticed these weren't done when I went to start on the bulk create stuff so I implemented them quick.

I moved around a few of the handlers just to make them ordered a bit easier to understand too (e.g. CRUD first, then subcollections, etc)

\# TODO:
- [x] validation service file that does validations on the application model
- [x] validation tests
- [x] handler tests.

### DIFF
--- a/dao/application_type_dao.go
+++ b/dao/application_type_dao.go
@@ -1,8 +1,11 @@
 package dao
 
 import (
+	"fmt"
+
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
+	"gorm.io/datatypes"
 )
 
 type ApplicationTypeDaoImpl struct {
@@ -72,4 +75,29 @@ func (a *ApplicationTypeDaoImpl) Update(_ *m.ApplicationType) error {
 
 func (a *ApplicationTypeDaoImpl) Delete(_ *int64) error {
 	panic("not needed (yet) due to seeding.")
+}
+
+func (at *ApplicationTypeDaoImpl) ApplicationTypeCompatibleWithSource(typeId, sourceId int64) (bool, error) {
+	source := m.Source{ID: sourceId}
+	result := DB.Preload("SourceType").Find(&source)
+	if result.Error != nil {
+		return false, fmt.Errorf("source not found")
+	}
+
+	// searching for the application type that has the source type's name in its
+	// supported source types column.
+	result = DB.Find(
+		&m.ApplicationType{Id: typeId},
+		datatypes.JSONQuery("supported_source_types").HasKey(source.SourceType.Name),
+	)
+
+	if result.Error != nil {
+		return false, fmt.Errorf("failed to fetch application type %v", typeId)
+	}
+
+	if result.RowsAffected == 0 {
+		return false, nil
+	}
+
+	return true, nil
 }
--- a/dao/application_type_dao.go
+++ b/dao/application_type_dao.go
@@ -77,27 +77,19 @@ func (a *ApplicationTypeDaoImpl) Delete(_ *int64) error {
 	panic("not needed (yet) due to seeding.")
 }
 
-func (at *ApplicationTypeDaoImpl) ApplicationTypeCompatibleWithSource(typeId, sourceId int64) (bool, error) {
+func (at *ApplicationTypeDaoImpl) ApplicationTypeCompatibleWithSource(typeId, sourceId int64) error {
 	source := m.Source{ID: sourceId}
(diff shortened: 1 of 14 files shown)


MikelAlejoBR (approved):

dao/application_type_dao.go
```diff
@@ -69,3 +72,28 @@ func (a *ApplicationTypeDaoImpl) Update(_ *m.ApplicationType) error {
...
+	result := DB.Preload("SourceType").Find(&source)
+	if result.Error != nil {
+		return false, fmt.Errorf("source not found")
+	}
+
+	// searching for the application type that has the source type's name in its
+	// supported source types column.
+	result = DB.Find(
+		&m.ApplicationType{Id: typeId},
+		datatypes.JSONQuery("supported_source_types").HasKey(source.SourceType.Name),
```
MikelAlejoBR, line 91: Probably I'm doing something wrong, but I'd like to ask: does this work as intended? I ask because:

1. It seems that the `supported_source_types` column contains an array, but not a JSON object with keys.
2. I've set up a test database and calling the function concatenating "bogus" to the `source.SourceType.Name`, and the function doesn't return errors.

```diff
	result = DB.Find(
		&m.ApplicationType{Id: typeId},
-		datatypes.JSONQuery("supported_source_types").HasKey(source.SourceType.Name),
+ 		datatypes.JSONQuery("supported_source_types").HasKey(source.SourceType.Name+"bogus"),
       )
```

This is the test I used in `main_test.go`:

```go
func TestCompatibleWithSource(t *testing.T) {
	tenantId := int64(1)
	dao := dao.ApplicationTypeDaoImpl{TenantID: &tenantId}

	database.ConnectToTestDB("public")

	result, err := dao.ApplicationTypeCompatibleWithSource(2, int64(1))
	if err != nil {
		t.Fatal(err)
	}

	if result != true {
		t.Fatal("not true")
	}
}
```

I created the following source before calling the function —source type was `azure`—:

|id|name|uid|created_at|updated_at|tenant_id|source_type_id|version|availability_status|imported|source_ref|last_checked_at|last_available_at|app_creation_workflow|paused_at|
|--|----|---|----------|----------|---------|--------------|-------|-------------------|--------|----------|---------------|-----------------|---------------------|---------|
|1|my sourcy|super wow string|2022-01-14 11:34:09.240|2022-01-14 11:34:09.240|2|4|ab|available|true|there|2022-01-14 11:34:09.240|2022-01-14 11:34:09.240|manual_configuration|2022-01-14 11:34:09.240|

And the following `tenant`:

|id|name|description|external_tenant|created_at|updated_at|
|--|----|-----------|---------------|----------|----------|
|2|my tenant|tenant enant hahaha|there|2022-01-14 11:33:45.637|2022-01-14 11:33:45.637|

lindgrenj6: > 1. It seems that the `supported_source_types` column contains an array, but not a JSON object with keys.

Correct, but the has key function works with it still. iirc a list is still valid json, https://jsonlint.com/ says so haha

> 2. I've set up a test database and calling the function concatenating "bogus" to the source.SourceType.Name, and the function doesn't return errors.

So this was the way I did it initially - I typed `.Find` instead of `.First`, Find just scans into a struct and doesn't populate the Error on the result unless something goes wrong. I'll change it to First, that was my bad. That way we're only checking it once instead of checking the error and then also checking if there were 0 rows returned.

MikelAlejoBR: > > 1. It seems that the `supported_source_types` column contains an array, but not a JSON object with keys.
> 
> Correct, but the has key function works with it still. iirc a list is still valid json, https://jsonlint.com/ says so haha

Ooooh okay, I thought `.HasKey` maybe only worked with JSON keys. TIL!

> > 2. I've set up a test database and calling the function concatenating "bogus" to the source.SourceType.Name, and the function doesn't return errors.
> 
> So this was the way I did it initially - I typed `.Find` instead of `.First`, Find just scans into a struct and doesn't populate the Error on the result unless something goes wrong. I'll change it to First, that was my bad. That way we're only checking it once instead of checking the error and then also checking if there were 0 rows returned.

Great :+1: !

lindgrenj6:
@lpichler can you review this for me quick today? i had to do a long rebase and don't want to have to do it again :sweat_smile:

lpichler:
> @lpichler can you review this for me quick today? i had to do a long rebase and don't want to have to do it again 😅

I was looking on it already but I have no idea why I didn't merge it  :D

lpichler (approved):

PR #74, opened 2022-01-14
[RHCLOUD-17533] Bulk Create

lindgrenj6:
The big one. This reimplements bulk_create on the Go side.

It wasn't too bad really, though I may change the endpoint up in the future to just build an object tree instead of separate arrays, but that would be a breaking change. 

- [x] #73 <- built off this one, so this one needs to go first.
- [x] #100 rebased onto this

\# TODO:
- [x] parse authentications
- [x] authtype validation
- [x] save all sources in transaction, monitor for failure
- [x] ~an ungodly amount of tests~ eventually. if this breaks everything breaks so that'll be the initial tests


model/application_authentication.go
```diff
@@ -15,7 +15,8 @@ type ApplicationAuthentication struct {
 	CreatedAt time.Time `json:"created_at"`
 	UpdatedAt time.Time `json:"updated_at"`
 
-	VaultPath string `json:"vault_path"`
+	// TODO: this doesn't exist yet
+	VaultPath string `json:"vault_path" gorm:"-"`
```
lpichler, line 19: `gorm: "-"` it doesn't create neither the column ?

Why we need `gorm:"-"` here ? I tough we are using it. Statuslistener's tests  started to fail(strange thing is that it is different error that here in CI) with this change.

lindgrenj6: It looks like it's the `AuthenticationUID` that is causing the tests to fail:
```diff
diff --git a/model/application_authentication.go b/model/application_authentication.go
index 50753ac..b979683 100644
--- a/model/application_authentication.go
+++ b/model/application_authentication.go
@@ -26,7 +26,7 @@ type ApplicationAuthentication struct {
 	// TODO: fix correctly PR#40
 	AuthenticationID int64 `json:"authentication_id"`
 
-	AuthenticationUID string `json:"-" gorm:"-"`
+	AuthenticationUID string `json:"-"`
 }
 
 func (aa *ApplicationAuthentication) ToEvent() interface{} {
```

Which is a little unfortunate - the reason I removed the VaultPath was because that column isn't in the DB yet so gorm fails when that column isn't there (e.g. outside the tests). 

The `AuthenticationUID` piece may be something we have to play with a bit more - we could just store it (in the migration on the rails side when we do the data migration from postgres -> vault) but then we'd have duplicated data (VaultPath + UID). 

I'm thinking we'll just store it in one place (Path), then transform it into the ID field that way the frontend doesn't have to change anything (the authentication_id will just return the UUID instead of the string representation of the pkey). We'll also have to remove the foreign key there. All sorts of fun. 

---

All in all I removed the `gorm:"-"` from the AuthenticationUID for now so the tests pass. We'll worry about making the application authentication "correct" when we do the actual migration to vault.

lindgrenj6:
@lpichler @MikelAlejoBR :eye: plox and thanks!

service/bulk_create.go
```diff
@@ -0,0 +1,421 @@
...
+		a.AuthType = auth.AuthType
+		a.Username = auth.Username
+		a.Password = auth.Password
+		a.Extra = auth.Extra
+		a.Name = auth.Name
+		a.TenantID = *tenantID
+
+		// if an id was passed in we're just adding an authentication to
+		// an already-existing resource.
+		id, err := strconv.ParseInt(auth.ResourceName, 10, 64)
```
MikelAlejoBR, line 266: I think this should be `auth.ResourceId`

dao/application_type_dao.go
```diff
@@ -105,18 +111,30 @@ func (a *applicationTypeDaoImpl) Delete(_ *int64) error {
...
-	result = DB.First(
-		&m.ApplicationType{Id: typeId},
-		datatypes.JSONQuery("supported_source_types").HasKey(source.SourceType.Name),
-	)
+	//
+	// initially i tried to use the
+	// datatypes.JsonQuery("application_types.supported_source_types") but that
+	// doesn't work when we're specifying something joined in, in this case
+	// "source_types.name"
+	result := DB.Select("application_types.*").
```
MikelAlejoBR: I think a `.Debug` statement is missing here

dao/source_type_dao.go
```diff
@@ -55,7 +55,14 @@ func (st *sourceTypeDaoImpl) GetById(id *int64) (*m.SourceType, error) {
 	return sourceType, nil
 }
 
-func (st *sourceTypeDaoImpl) Create(_ *m.SourceType) error {
+func (st *sourceTypeDaoImpl) GetByName(name string) (*m.SourceType, error) {
+	sourceType := &m.SourceType{}
+	result := DB.Where("name LIKE ?", "%"+name+"%").First(sourceType)
```
MikelAlejoBR: Missing a `.Debug` statement

service/bulk_create.go
lindgrenj6: Actually this is a fun little piece of functionality that I'm not sure anyone uses - you can attach an authentication to an existing application/endpoint/source. In the bulk create payload you'd just use `{resource_name: "42", resource_type: "Application"}` in the authentication part of the payload. 

I'm not sure anyone actually uses it - but thats what this is for. Checking to see if an int was passed in or not.

MikelAlejoBR: Aaah got it. Thanks for the explanation!

lpichler:
I did:

```
POST http://localhost:{{port}}/api/sources/v3.1/bulk_create

{
  "sources": [
    {
      "name": "my amazon sourceGO",
      "source_type_name": "amazon"
    }
  ],
  "applications": [
    {
      "source_name": "my amazon sourceGO",
      "application_type_name": "cost-management"
    },
    {
      "source_name": "my amazon sourceGO",
      "application_type_name": "cloud-meter"
    }
  ],
  "authentications": [
    {
      "resource_type": "source",
      "resource_name": "my amazon sourceGO",
      "username": "aws::iam::arn::i_cant_remember_the_format",
      "authtype": "arn"
    },
    {
      "resource_name": "cost-management",
      "username": "myUser",
      "password": "myPass"
    },
    {
      "resource_name": "cloud-meter",
      "username": "cloudUser",
      "password": "cloudPass"
    }
  ]
}
```
response
```

HTTP/1.1 500 Internal Server Error
X-Request-Duration-Ms: 28
Date: Fri, 25 Mar 2022 13:51:29 GMT
Content-Length: 202

{
  "errors": [
    {
      "detail": "Internal Server Error: ERROR: insert or update on table \"application_authentications\" violates foreign key constraint \"fk_rails_d709bbbff3\" (SQLSTATE 23503)",
      "status": "500"
    }
  ]
}
```

```
INSERT INTO \"application_authentications\" (\"paused_at\",\"created_at\",\"updated_at\",\"tenant_id\",\"application_id\",\"authentication_id\",\"authentication_uid\") VALUES ('0000-00-00 00:00:00','2022-03-25 14:51:29.833','2022-03-25 14:51:29.833',1,7,0,'3a3f5926-9eff-4fb6-ac38-1b58169d22bb'),('0000-00-00 00:00:00','2022-03-25 14:51:29.833','2022-03-25 14:51:29.833',1,8,0,'7d13a026-aa8c-440a-b48b-ed21efc18c9d') RETURNING \"id\""
```

we need `authentication_id`.
Are we going to leave it as it is until we will remove this foreign key ?

lindgrenj6:
Yeah - basically we're going to need to leave it be until we have the DB DAO merged. It should "just work" with the DB DAO (after adding the field setting) since that key won't be violated anymore. I may just wait until the postgres backed DB DAO is merged to make sure and update this. It'll just be setting that key and then inserting it separately.

lpichler (approved):

model/application_authentication.go
lpichler: yea I think we can do it with just storing in Path.
thanks!

lindgrenj6:
Yeah - basically we're going to need to leave it be until we have the DB
DAO merged. It should "just work" with the DB DAO since that key won't be
violated.
--
Jacob Lindgren

Red Hat

PR #75, opened 2022-01-14
[RHCLOUD-17536] Add /sources/:id/check_availability Endpoint

lindgrenj6:
[RHCLOUD-17536](https://issues.redhat.com/browse/RHCLOUD-17536)

Adds the endpoint to request an availability status for a source.

It basically follows this flow:
1. Lookup source (and basically all of its subresources)
2. Loop through applications, sending http requests to the configured endpoint (set in deploy/clowdapp.yml)
3. Loop through endpoints, checking to see if the source is a satellite source, then proceeding to produce kafka messages for the source.

--- 

Another notable thing is the `SourceDao.GetByIdWithPreload` which basically allows you to pass a variable amount of args and let gorm.io preload the relations. Works great!

\# TODO:
- [x] tests, shouldn't be too bad.

### DIFF
--- a/service/availability_check.go
+++ b/service/availability_check.go
@@ -1,0 +1,140 @@
+package service
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"strconv"
+	"time"
+
+	"github.com/RedHatInsights/sources-api-go/config"
+	"github.com/RedHatInsights/sources-api-go/kafka"
+	l "github.com/RedHatInsights/sources-api-go/logger"
+	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/util"
+)
+
+// storing the satellite topic here since it doesn't change after initial
+// startup.
+var satelliteTopic = config.Get().KafkaTopic("platform.topological-inventory.operations-satellite")
+
+// requests both types of availability checks for a source
+func RequestAvailabilityCheck(source *m.Source) {
+	l.Log.Infof("Requesting Availability Check for Source [%v]", source.ID)
+
+	if len(source.Applications) != 0 {
+		applicationAvailabilityCheck(source)
+	}
+
+	if len(source.Endpoints) != 0 {
+		endpointAvailabilityCheck(source)
+	}
+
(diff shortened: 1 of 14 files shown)


MikelAlejoBR (approved):
LGTM! I like the nice little `GetByIdWithPreload` varadic function!

service/availability_check.go
```diff
@@ -0,0 +1,154 @@
...
+var (
+	// storing the satellite topic here since it doesn't change after initial
+	// startup.
+	satelliteTopic = config.Get().KafkaTopic("platform.topological-inventory.operations-satellite")
+	// default availability checker instance
+	ac availabilityChecker = &availabilityCheckRequester{}
+)
+
+// requests both types of availability checks for a source
+func RequestAvailabilityCheck(source *m.Source) {
```
lpichler, line 35: we have 2 methods with name `RequestAvailabilityCheck` and `requestAvailabilityCheck` - shouldn't we name it differently ?

service/availability_check.go
```diff
@@ -0,0 +1,154 @@
...
+	ctx, done := context.WithTimeout(context.Background(), 10*time.Second)
+	defer done()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, uri.String(), bytes.NewBuffer(raw))
+	if err != nil {
+		l.Log.Warnf("Failed to make request for application [%v], uri [%v]", app.ID, uri.String())
+		return
+	}
+
+	req.Header.Add("x-rh-sources-account-number", source.Tenant.ExternalTenant)
```
lpichler, line 83: Should we add `x-rh-sources-account-number` header also to Kafka message ?

lpichler (approved):
I had 2 questions otherwise it looks and works great 👍

service/availability_check.go
lindgrenj6: Definitely - I renamed it to a more apt `httpAvailabilityRequest`

service/availability_check.go
lindgrenj6: I'm not sure if satellite uses it but it may - so I added it just in case.

lpichler (approved):

PR #76, opened 2022-01-24
[WIP] [RHCLOUD-17059] Add Authentications/ApplicationAuthentications to Records.event 

lpichler:
This PR is split to more smaller PRs:

- [x] 1 https://github.com/RedHatInsights/sources-api-go/pull/77
- [x] 2 https://github.com/RedHatInsights/sources-api-go/pull/78
- [x] 3 https://github.com/RedHatInsights/sources-api-go/pull/79
- [x] 4 https://github.com/RedHatInsights/sources-api-go/pull/80
- [x] 5  https://github.com/RedHatInsights/sources-api-go/pull/87
- [x] 6 https://github.com/RedHatInsights/sources-api-go/pull/98
- [x] 7 https://github.com/RedHatInsights/sources-api-go/pull/117

in ruby project:
https://github.com/RedHatInsights/sources-api/blob/19cf1aa2f6f2ba7f0ce8c17dc8a4662cfabc368d/app/models/concerns/event_concern.rb

Change adds https://github.com/RedHatInsights/sources-api/blob/19cf1aa2f6f2ba7f0ce8c17dc8a4662cfabc368d/app/models/concerns/event_concern.rb#L14-L25
and 
https://github.com/RedHatInsights/sources-api/blob/19cf1aa2f6f2ba7f0ce8c17dc8a4662cfabc368d/app/models/concerns/event_concern.rb#L39-L40

### Links
https://issues.redhat.com/browse/RHCLOUD-17059

PR #77, opened 2022-01-24
Use interface for Vault instance

lpichler:
This change also add mock struct which satisfy interface `LogicalClient` as well as `vault.Logical` - it can be used for testing for Vault-related code.

### Links
Pull out from https://github.com/RedHatInsights/sources-api-go/pull/76

### DIFF
--- a/dao/db.go
+++ b/dao/db.go
@@ -16,7 +16,7 @@ var (
 	DB *gorm.DB
 
 	vaultClient *vault.Client
-	Vault       *vault.Logical
+	Vault       VaultClient
 
 	conf = config.Get()
 )
@@ -55,6 +55,7 @@ func Init() {
 	if err != nil {
 		panic(fmt.Sprintf("Failed to Create Vault Client: %v", err))
 	}
+
 	Vault = vaultClient.Logical()
 }
 
--- a/dao/interfaces.go
+++ b/dao/interfaces.go
@@ -3,6 +3,7 @@ package dao
 import (
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
+	"github.com/hashicorp/vault/api"
 )
 
 type SourceDao interface {
@@ -89,4 +90,11 @@ type SourceTypeDao interface {
 	Create(src *m.SourceType) error
 	Update(src *m.SourceType) error
 	Delete(id *int64) error
+}
+
+type VaultClient interface {
+	Read(path string) (*api.Secret, error)
+	List(path string) (*api.Secret, error)
+	Write(path string, data map[string]interface{}) (*api.Secret, error)
+	Delete(path string) (*api.Secret, error)
 }
--- a/internal/testutils/mocks/mock_vault.go
+++ b/internal/testutils/mocks/mock_vault.go
@@ -1,0 +1,41 @@
+package mocks
+
+import (
+	"fmt"
+
+	"github.com/RedHatInsights/sources-api-go/internal/testutils/fixtures"
+	"github.com/hashicorp/vault/api"
+)
+
+type MockVault struct {
+}
+
+func (m *MockVault) Read(path string) (*api.Secret, error) {
+	if path != fmt.Sprintf("secret/data/%d/%v", fixtures.TestTenantData[0].Id, "") {
+		return nil, nil
+	}
+
+	secret := &api.Secret{}
+	secret.Data = make(map[string]interface{})
+	secret.Data["data"] = map[string]interface{}{}
+	secret.Data["metadata"] = map[string]interface{}{}
+
+	return secret, nil
+}
+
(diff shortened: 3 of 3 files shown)


dao/authentication_dao.go
```diff
@@ -210,7 +210,7 @@ func (a *AuthenticationDaoImpl) Create(auth *m.Authentication) error {
 		return err
 	}
 
-	out, err := Vault.Write(path, data)
+	out, err := (*Vault).Write(path, data)
```
lindgrenj6: So I'm actually kind of mad at myself that I missed this the first time in the availability status listener pr :disappointed:. 

We shouldn't be using pointers of interfaces because basically an interface is a pointer to a struct that satisfies a "signature", so basically an interface type is _already_ a pointer.

For example see here how I instantiated an Rbac interface var: https://github.com/RedHatInsights/sources-api-go/blob/e5ad9bbed6af58bf92fb9540cefebac7fd42069c/middleware/authorization.go#L20

The Interface var is assigned a pointer to a struct that "satisfies" the interface. 

I noticed it last time and thought a bit about it here: https://github.com/RedHatInsights/sources-api-go/blob/e5ad9bbed6af58bf92fb9540cefebac7fd42069c/dao/common.go#L9 I didn't notice that the `EventModelDao` is actually an interface - therefore we shouldn't be returning pointers to it. Basically a pointer to an interface makes it take double de-referencing which doubles the potential for null-pointer errors. This way we won't have to de-ref the pointer before calling methods on the structs as well. 

also: you can see here https://github.com/RedHatInsights/sources-api-go/blob/e5ad9bbed6af58bf92fb9540cefebac7fd42069c/dao/common.go#L13

that basically we're returning a pointer to a pointer. Should have jumped out to me more but I glossed over it the first time.

---

tl; dr: can we refactor these functions/where they're used to not be pointers to interfaces. Everywhere there is `*EventModelDao` needs to just be `EventModelDao`. 

If this doesn't make sense let me know and I can try and explain it better in person, also if you don't want do do the `EventModelDao` changes in this PR thats fine - we can do them in a follow-up.

dao/db.go
```diff
@@ -16,11 +16,15 @@ var (
 	DB *gorm.DB
 
 	vaultClient *vault.Client
-	Vault       *vault.Logical
+	Vault       *LogicalClient
```
lindgrenj6: This is an interface so shouldn't be a pointer like I said earlier - this is what made me realize I missed it last time.

dao/db.go
```diff
@@ -55,7 +59,8 @@ func Init() {
 	if err != nil {
 		panic(fmt.Sprintf("Failed to Create Vault Client: %v", err))
 	}
-	Vault = vaultClient.Logical()
+
+	SetVault(vaultClient.Logical())
```
lindgrenj6: I actually don't think we need the function here - one can just do:
```go
Vault = &vaultClient.Logical()
````
and in tests, just overwrite it in the same way.

File: dao/interfaces.go [hunk 4]
```diff
@@ -90,3 +91,10 @@ type SourceTypeDao interface {
 	Update(src *m.SourceType) error
 	Delete(id *int64) error
 }
+
+type LogicalClient interface {
```
lindgrenj6: Lets rename this to `VaultClient`.

lindgrenj6 (changes requested):

MikelAlejoBR (approved):

dao/authentication_dao.go
lpichler: oh wow that's interesting - it definitely needs to be changed.

Regard to:
https://github.com/RedHatInsights/sources-api-go/blob/e5ad9bbed6af58bf92fb9540cefebac7fd42069c/dao/common.go#L13

It looks same like this 

https://github.com/RedHatInsights/sources-api-go/blob/e5ad9bbed6af58bf92fb9540cefebac7fd42069c/middleware/authorization.go#L20

 I would say it is initialising pointer with address(`&SourceDaoImpl{}`) which seems OK  as `resource` is not pointer on interface but definitely returning values needs to be changed  in the functions and also other occurrence.

I will change `*EventModelDao` in follow up PR and I prefer to do it after series of PRs about adding auths and appauths to listener.

@lindgrenj6 It does make sense and thanks for explanation!

dao/db.go
lpichler: when `Vault` is interface variable then just
`Vault = vaultClient.Logical()`

lpichler:
thanks @MikelAlejoBR for review 👍 

and @lindgrenj6 I updated PR according to your suggestions and now it is very little change

dao/authentication_dao.go
lindgrenj6: Yep that sounds good to me. Just the return type being a pointer is the part I had missed last time - it's easy to add many layers of indirection!

lindgrenj6 (approved):

PR #79, opened 2022-01-24
Move out struct status message to "types" package

lpichler:
WIP:
- [x] https://github.com/RedHatInsights/sources-api-go/pull/78

This change allows to avoid cyclic dependency  issues.

### Links
Pull out from https://github.com/RedHatInsights/sources-api-go/pull/76
- Pull out from https://github.com/RedHatInsights/sources-api-go/pull/76
- PR begins from commit [Move struct StatusMessage to 'types' package](https://github.com/RedHatInsights/sources-api-go/pull/79/commits/564f6d2f176a320323f6e224b4f89f5efb1a91ce)

PR #80, opened 2022-01-24
Add and use struct Resource

lpichler:
WIP:
- [x] https://github.com/RedHatInsights/sources-api-go/pull/79

Add struct Resource and pass its instance to functions
- Pull out from https://github.com/RedHatInsights/sources-api-go/pull/76
- PR begins from commit [Add struct Resource and pass its instance to functions](https://github.com/RedHatInsights/sources-api-go/pull/80/commits/93271c3b96b19333d203696709de2d1d127569e9)

### DIFF
--- a/model/events.go
+++ b/model/events.go
@@ -4,21 +4,31 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+
+	"github.com/RedHatInsights/sources-api-go/util"
 )
 
 type EventModelDao interface {
-	BulkMessage(id *int64) (map[string]interface{}, error)
-	FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error
-	ToEventJSON(id *int64) ([]byte, error)
+	BulkMessage(resource util.Resource) (map[string]interface{}, error)
+	FetchAndUpdateBy(resource util.Resource, updateAttributes map[string]interface{}) error
+	ToEventJSON(resource util.Resource) ([]byte, error)
 }
 
-func UpdateMessage(eventObject *EventModelDao, resourceID int64, resourceType string, attributes []string) ([]byte, error) {
+func UpdateMessage(eventObject *EventModelDao, resource util.Resource, attributes []string) ([]byte, error) {
 	updatedAttributes := map[string]interface{}{}
-	updatedAttributes["updated"] = map[string]interface{}{resourceType: map[string]interface{}{strconv.Itoa(int(resourceID)): attributes}}
 
-	bulkMessage, err := (*eventObject).BulkMessage(&resourceID)
+	resourceID := ""
+	if resource.ResourceUID != "" && resource.ResourceID == 0 {
+		resourceID = resource.ResourceUID
+	} else {
+		resourceID = strconv.Itoa(int(resource.ResourceID))
+	}
+
+	updatedAttributes["updated"] = map[string]interface{}{resource.ResourceType: map[string]interface{}{resourceID: attributes}}
+
+	bulkMessage, err := (*eventObject).BulkMessage(resource)
 	if err != nil {
-		return nil, fmt.Errorf("error in BulkMessage: %s", err.Error())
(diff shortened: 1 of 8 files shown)


lpichler:
4/6 👼

model/events.go
```diff
@@ -4,21 +4,31 @@ import (
...
-	BulkMessage(id *int64) (map[string]interface{}, error)
-	FetchAndUpdateBy(id *int64, updateAttributes map[string]interface{}) error
-	ToEventJSON(id *int64) ([]byte, error)
+	BulkMessage(resource util.Resource) (map[string]interface{}, error)
+	FetchAndUpdateBy(resource util.Resource, updateAttributes map[string]interface{}) error
+	ToEventJSON(resource util.Resource) ([]byte, error)
 }
 
-func UpdateMessage(eventObject *EventModelDao, resourceID int64, resourceType string, attributes []string) ([]byte, error) {
+func UpdateMessage(eventObject *EventModelDao, resource util.Resource, attributes []string) ([]byte, error) {
```
lindgrenj6, line 17: We can make this `EventModelDao` not a pointer anymore since it is an interface (unless we're going to do that in a followup later)

statuslistener/statuslistener.go
```diff
@@ -86,9 +86,9 @@ func (avs *AvailabilityStatusListener) headersFrom(message kafka.Message) []kafk
 }
 
 func (avs *AvailabilityStatusListener) processEvent(statusMessage types.StatusMessage, headers []kafka.Header) {
-	resourceID, err := util.InterfaceToInt64(statusMessage.ResourceID)
-	if err != nil {
-		logging.Log.Errorf("Error parsing resource_id: %s", err.Error())
+	resource := util.Resource{}
+	if resource.Parse(statusMessage) != nil {
```
lindgrenj6: So for this it's a bit more idiomatic go to have a function like `util.ParseStatusMessageToResource` that returns the resource + an error. I could see the error message being what was logged here.

EDIT: It looks like there is an error being returned already! So we can just change the signature of that function.

statuslistener/statuslistener.go
```diff
@@ -104,7 +104,7 @@ func (avs *AvailabilityStatusListener) processEvent(statusMessage types.StatusMe
 		return
 	}
 
-	err = (*modelEventDao).FetchAndUpdateBy(&resourceID, updateAttributes)
+	err = (*modelEventDao).FetchAndUpdateBy(resource, updateAttributes)
```
lindgrenj6: So I think we can change this var from a pointer to an interface to just the interface (once again unless we're gonna do it in a followup).

lindgrenj6 (commented):
Just one small comment really - the pulling out of the pointer-interface can probably wait until later.

util/resource.go
```diff
@@ -0,0 +1,35 @@
...
+	ResourceType  string
+	ResourceID    int64
+	ResourceUID   string
+	TenantID      int64
+	AccountNumber string
+}
+
+func ParseStatusMessageToResource(resource *Resource, statusMessage types.StatusMessage) (*Resource, error) {
+	resource.ResourceUID = statusMessage.ResourceID
+	resource.ResourceType = Capitalize(statusMessage.ResourceType)
```
lpichler, line 21: @lindgrenj6 I moved `Capitalize` here.

model/events.go
lpichler: I tried but I need to play more with *EventModelDao -> EventModelDao, so let's do it in follow up PR.

statuslistener/statuslistener.go
lpichler: done 👍

statuslistener/statuslistener.go
lpichler: (*modelEventDao) -> (modelEventDao) will be in follow up PR.

lpichler:
@lindgrenj6 I addressed your feedback, thanks!

lindgrenj6 (approved):

PR #83, opened 2022-01-26
[RHCLOUD-15002] Seeding

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-15002

This is a quite-old branch but I cleaned it up before opening this.

So far I only have SourceType seeding done - I need to continue and get ApplicationType+metadata done.

EDIT: 1/29/22
Ready to go! I had to duplicate some of the database db util code in the dao's maintest class due to the fact we can't cyclically import it. Other than that this is pretty sweet!

\# TODO:
- [x] SourceType Seeding
- [x] ApplicationType Seeding (done but only through loosey-goosey `map[string]interface{}`), struct isn't cooperating yet
- [x] MetaData Seeding

### DIFF
--- a/dao/seeding.go
+++ b/dao/seeding.go
@@ -212,11 +212,14 @@ func seedMetaData(seedDir string) error {
 		return result.Error
 	}
 
-	err := seedSuperkeyMetadata(seedDir)
+	// load the application types once so we don't load them in a loop
+	appTypes := loadApplicationTypeSeeds()
+
+	err := seedSuperkeyMetadata(seedDir, appTypes)
 	if err != nil {
 		return err
 	}
-	err = seedAppMetadata(seedDir)
+	err = seedAppMetadata(seedDir, appTypes)
 	if err != nil {
 		return err
 	}
@@ -224,7 +227,7 @@ func seedMetaData(seedDir string) error {
 	return nil
 }
 
-func seedSuperkeyMetadata(seedDir string) error {
+func seedSuperkeyMetadata(seedDir string, appTypes map[string]*m.ApplicationType) error {
 	seeds := make(superkeyMetadataSeedMap)
 	data, err := os.ReadFile(seedDir + "superkey_metadata.yml")
 	if err != nil {
@@ -238,11 +241,9 @@ func seedSuperkeyMetadata(seedDir string) error {
 
 	for name, value := range seeds {
 		// first find the application type we are going to set the metadata on
-		var apptype m.ApplicationType
-		result := DB.Where("name = ?", name).First(&apptype)
-		if result.Error != nil {
-			l.Log.Errorf("Failed to find application type %v", name)
-			return result.Error
+		appType, ok := appTypes[name]
+		if !ok {
+			return fmt.Errorf("failed find application type %v", name)
 		}
 
 		// create each "step" as a record in the db
@@ -261,11 +262,11 @@ func seedSuperkeyMetadata(seedDir string) error {
 				Name:              values.Name,
 				Payload:           payload,
 				Substitutions:     substitutions,
-				ApplicationTypeID: apptype.Id,
+				ApplicationTypeID: appType.Id,
 				Type:              "SuperKeyMetaData",
 			}
 
-			result = DB.Create(&metadata)
+			result := DB.Create(&metadata)
 			if result.Error != nil {
 				return result.Error
 			}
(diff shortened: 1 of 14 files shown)


MikelAlejoBR (approved):

dao/seeding.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// loop through all of the seeds
+	// 1. creating the record if they don't exist
+	// 2. updating the ones that do exist
+	// 3. deleting the ones removed from the list
+	for name, values := range seeds {
+		var st *m.SourceType
+
+		// find an existing one in the list of source types
+		for i, sourceType := range sourceTypes {
```
lpichler: @lindgrenj6 

Do we need cycle inside ? 
What about to fetch all sources to map(let's say in variable sourceTypeMap) `map[string]m.SourceType` where index is name of source type - before this `seeds` loop.

I guess then we can ask check  `if sourceTypeMap[name] != nil` and (as you have):
1. If exists then -  update and remove from(maybe set nil) `sourceTypeMap[name]`
2. if don't exists - create 
3. What is remaining in sourceTypeMap - it should be removed.

Let me what do you think but It looks with this we can avoid this loop inside other loop.

dao/seeding.go
```diff
@@ -0,0 +1,185 @@
...
+
+	err = yaml.Unmarshal(data, &seeds)
+	if err != nil {
+		return err
+	}
+
+	for name, values := range seeds {
+		var at m.ApplicationType
+
+		result := DB.Where("name = ?", name).First(&at)
```
lpichler: Maybe it would be possible use way as was mentioned in previous comment - then we can avoid query for each item in `seeds`

dao/seeding.go
```diff
@@ -0,0 +1,185 @@
...
+		at.SupportedAuthenticationTypes = supportedAuthenticationTypes
+		at.Name = name
+
+		result = DB.Save(&at)
+		if result.Error != nil {
+			return result.Error
+		}
+	}
+
+	return nil
```
lpichler, line 205: isn't missing delete part here ?

dao/seeding.go
```diff
@@ -0,0 +1,185 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = yaml.Unmarshal(data, &seeds)
+	if err != nil {
+		return err
+	}
+
+	// TODO: delete entire metadata table and re-seed it every time (id's do not
```
lpichler: will this be in follow up PR ?

lpichler:
@lindgrenj6 do we need also [SKIP_LIST](https://github.com/RedHatInsights/sources-api/blob/014a2dec984de7711425128a2e9b6b7dbeb9996f/app/models/concerns/seeding_concern.rb#L10) from ENV vars ?

dao/seeding.go
lindgrenj6: Well - it's there already, I think you may have had an old version up when you reviewed. It is probably worth a re-review.

dao/seeding.go
lindgrenj6: Same as above - I think you may have had an old version of the code when you reviewed.

lindgrenj6:
> do we need also [SKIP_LIST](https://github.com/RedHatInsights/sources-api/blob/014a2dec984de7711425128a2e9b6b7dbeb9996f/app/models/concerns/seeding_concern.rb#L10) from ENV vars ?

I do need to do that still - thought about it but forgot to implement it. I'll get it added!

dao/seeding.go
lindgrenj6: Ok I went through and changed this - there are 2 new functions that parse the list of existing SourceTypes/ApplicationTypes into a `map[string]*SourceType`.

This way we can check if something exists or not (map will have a nil value for non-existing attributes), and we just delete them out of the hash instead of deleting them out of the slice.

It is a bit longer code-wise but it is pretty straight forward still.

dao/seeding.go
lindgrenj6: Switched it to use the newly added `loadApplicationTypes` function so we only load them from the DB once (after they've been seeded) and pass them to the metadata functions.

lindgrenj6:
Skip list has been added.

dao/seeding.go
lpichler: I see, now I can see changes I am not sure why I saw older changes.

lpichler (approved):
thanks for changes @lindgrenj6! looks and works great 👍

PR #86, opened 2022-01-28
Fix some type asserts and skip running the linter on test files

lindgrenj6:
golangci-lint has been failing for a minute now - took some time to fix the type asserts I had introduced pre-linter and then got copied.

Also added the config file for golangci-lint to not hold the test files as accountable as the actual source files.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,7 @@
+issues:
+  exclude-rules:
+    # skip forcing a type assert on test files. can add more skips for test files as necessary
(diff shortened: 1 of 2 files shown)


MikelAlejoBR:
This is cool. I had to do some weird stuff myself because the linter was complaining... But I didn't think of making it shush lol

Having it silenced for tests seems pretty safe, so... Let's go with it!

MikelAlejoBR (approved):

lpichler (approved):

PR #87, opened 2022-01-28
Update availability status for Authentications

lpichler:
- there are some additions on authentication related structs
- added body of method to update availability status for authentications

### Links 
- Pull out from https://github.com/RedHatInsights/sources-api-go/pull/76

### DIFF
--- a/model/authentication.go
+++ b/model/authentication.go
@@ -1,7 +1,6 @@
 package model
 
 import (
-	"encoding/json"
 	"strconv"
 	"time"
 )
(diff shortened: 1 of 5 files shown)


MikelAlejoBR (approved):
LGTM. There's a couple of things I'd like to comment though:

1. The `time.Parse` part could potentially throw a `panic` if we ever receive a `last_available_at` value which isn't a `string`. I guess that the chance of that happening is minimal, since we're the only ones producing that bit of information. Do you think it is necessary in this case to perform a type check just for safety? I'm debating myself if it is or not, but I'll leave the question here anyway to see what you think. Code panicking example: https://go.dev/play/p/rWY08cUlnNt
2. It feels a little bit odd having to marshal stuff to JSON and then unmarshall it back to the struct in the `UpdateBy` function. There's a library, [github.com/mitchellh/mapstructure](https://github.com/mitchellh/mapstructure) which seems to take an interface and maps it to a struct. However I cannot help but feel that having to use a library just for this seems a little bit odd as well... I'm unsure of what would be the best option here. I would vote for looping through the `map[string]interface{}` and mapping each field to the struct. @lindgrenj6, any thoughts?

lpichler:
@MikelAlejoBR

1. yes, I added handling error for those type assertion for `LastAvailableAt` and `LastCheckedAt`  - now  `time.Parse`  receives only string.

2. I understand but it is nice that `Unmarshall` updates in `auth` only fields in data - anyway I am curious on @lindgrenj6's feedback.

model/authentication.go
```diff
@@ -83,3 +87,16 @@ func (auth *Authentication) ToVaultMap() (map[string]interface{}, error) {
...
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(data, &auth)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
lindgrenj6: I know marhshaling + unmarshaling might be the nicest way to do this here - but I really think we should just do a loop over the map and update the fields. Much like we do for [the `UpdateFromRequest` methods on various models](https://github.com/RedHatInsights/sources-api-go/blob/7790b0e4d534d2ed510c7cd4470f77b4022b6704/model/authentication_http.go#L47-L69). It looks like something similar was done in the `authFromVault()` method too at the top of this PR.

This way it is explicit what we are doing and we don't have to worry about something going wrong with either of the json marshals.

lindgrenj6 (commented):
Just one tiny comment - I think we should be explicit about it by doing a loop over the map rather than doing a marshal/unmarshal from JSON. It keeps the complexity down and we know exactly what is happening.

lpichler:
@lindgrenj6 I stated fields to update explicitly as you suggested, thanks

lindgrenj6 (approved):

PR #88, opened 2022-01-28
Fix linter-fixes fixes: return error on failed cast rather than nothing

lindgrenj6:
I made a mistake in #86, returning nil is fine in the `authFromVault` function - but we need to bubble the error up. Whoops.

### DIFF
--- a/dao/authentication_dao.go
+++ b/dao/authentication_dao.go
@@ -255,7 +255,7 @@ func (a *AuthenticationDaoImpl) Create(auth *m.Authentication) error {
 
 	number, ok := out.Data["version"].(json.Number)
 	if !ok {
-		return nil
+		return errors.New("failed to cast vault version number to string")
 	}
 	auth.Version = number.String()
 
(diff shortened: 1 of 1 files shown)


lpichler (approved):

dao/authentication_dao.go
```diff
@@ -255,7 +255,7 @@ func (a *AuthenticationDaoImpl) Create(auth *m.Authentication) error {
 
 	number, ok := out.Data["version"].(json.Number)
 	if !ok {
-		return nil
+		return fmt.Errorf("failed to cast vault version number to string")
```
MikelAlejoBR: Small nitpick — would suggest using `errors.New` here, since you're not using any verbs in the string

MikelAlejoBR (approved):

lpichler:
@MikelAlejoBR merge please if you are OK with it 👍

MikelAlejoBR:
I can't, I don't have enough permissions to merge PRs :(

dao/authentication_dao.go
lindgrenj6: alrighty, i usually just do fmt since its already imported. but all good haha updated.

MikelAlejoBR: That's actually a good point

PR #89, opened 2022-01-28
refactor: run Miniredis in a non-default port

MikelAlejoBR:
Apparently Miniredis is causing some issues with people who are running a full Redis instance locally. It seems that turning off the Redis instance makes the tests pass again, so changing the port on which Miniredis runs should do the trick.

Instead of using the "RunT" function of Miniredis and requesting a random port, a hardcoded one is used instead. This is to avoid having the one in a million chance of getting the default Redis port as the random port. Hopefully this will avoid head scratching to the developer who has to experience tests failing once, and then passing without any issue in the next run.

### DIFF
--- a/redis/marketplace_token_test.go
+++ b/redis/marketplace_token_test.go
@@ -14,6 +14,7 @@ import (
 )
 
 var tokenCacher = &MarketplaceTokenCacher{TenantID: 5}
+var redisAddr = "localhost:45884"
 
 // setUpFakeToken sets up a test token ready to be used.
 func setUpFakeToken() *marketplace.BearerToken {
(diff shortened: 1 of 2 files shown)


redis/marketplace_token_test.go
```diff
@@ -28,7 +29,8 @@ func setUpFakeToken() *marketplace.BearerToken {
 
 // TestGetTokenBadTenant tests that when given a bad or nonexistent tenant, an expected error is returned.
 func TestGetTokenBadTenant(t *testing.T) {
-	mr, err := miniredis.Run()
+	mr := miniredis.NewMiniRedis()
```
lindgrenj6: So this is probably a weird request - but could we spin this up in the `TestMain` package that way it's there for all the tests? Or is there a reason we're spinning up a new one every time?

MikelAlejoBR: Yes, that is a good idea. I initially tried to do that, but I faced some problems I thought they were caused by Miniredis being run at the package level. But it seems that it was something else. Implemented the start/stop at the package level. Thanks!

lpichler (approved):

lindgrenj6 (approved):

PR #90, opened 2022-01-29
Fix /app_meta_data endpoint to resemble the original application

lindgrenj6:
Just some things I found while working on the seeding.

1. Don't need create/update/delete methods on the DAO since we're seeding it and its static information
2. Locked down the list view to be only the `AppMetaData` type
3. Removed availabilitystatus/pause fields since those don't exist on the model


lpichler (approved):
LGTM, thanks

MikelAlejoBR (approved):

PR #91, opened 2022-01-30
[RHCLOUD-17055] Add Middleware for raising event-stream events

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-17055

### So this is the start of raising events on create/update/delete events. 

I implemented it as a middleware component that runs _after_ the handler runs. So basically we just have to add the middleware to the request stack + set the `resource` field for each event in the handler. I commented it a lot so let me know if you have questions.

This is probably easiest reviewed in commit order. 

#### General Flow
1. request goes through stack, gets authenticated, 
2. raiseEvent middleware gets called, passed down the stack
3. request completes (whether error or not) and returns down the stack
4. event middleware looks and sees if there is a resource to raise and does so if needed
5. the request is rendered to the user while the kafka raise happens in the background

I'm debating adding a `go` directive around the raise event part - since we don't bubble that to the user ever. It'll just fail silently and warn us in the logs.

---

\# TODO:
- [x] Add middleware to all request stacks
- [x] change to raise event if since we aren't _always_ supposed to raise events. **EDIT**: added support so we can force skip in the handler
- [x] Add `c.Set("resource", xxx.ToResponse())` in all handlers
- [x] tests!

### DIFF
--- a/middleware/event_stream.go
+++ b/middleware/event_stream.go
@@ -60,7 +60,7 @@ func raiseEvent(eventType string) echo.MiddlewareFunc {
 			// in the handler for this to work.
 			resource := c.Get("resource")
 			if resource == nil {
-				l.Log.Warnf("failed to pull event resource from context - skipping raise event")
+				l.Log.Infof("failed to pull event resource from context - skipping raise event")
 				return nil
 			}
 
--- a/middleware/event_stream.go
+++ b/middleware/event_stream.go
@@ -1,0 +1,131 @@
+package middleware
+
+import (
+	"encoding/json"
+
+	"github.com/RedHatInsights/sources-api-go/internal/events"
+	"github.com/RedHatInsights/sources-api-go/kafka"
+	l "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/RedHatInsights/sources-api-go/util"
+	"github.com/labstack/echo/v4"
+)
+
+// middleware functions for each event type - adding this to the middleware
+// stack will raise an event based on the `resource` field set in the context
+// during the handler
+
+var (
+	RaiseSourceCreateEvent  = raiseEvent("Source.create")
+	RaiseSourceUpdateEvent  = raiseEvent("Source.update")
+	RaiseSourceDestroyEvent = raiseEvent("Source.destroy")
+
+	RaiseApplicationCreateEvent  = raiseEvent("Application.create")
+	RaiseApplicationUpdateEvent  = raiseEvent("Application.update")
+	RaiseApplicationDestroyEvent = raiseEvent("Application.destroy")
+
+	RaiseEndpointCreateEvent  = raiseEvent("Endpoint.create")
+	RaiseEndpointUpdateEvent  = raiseEvent("Endpoint.update")
+	RaiseEndpointDestroyEvent = raiseEvent("Endpoint.destroy")
+
+	RaiseAuthenticationCreateEvent  = raiseEvent("Authentication.create")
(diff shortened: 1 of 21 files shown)


lpichler:
probably it is https://issues.redhat.com/browse/RHCLOUD-17055

MikelAlejoBR (approved):
Looks good to me! I must admit that the "keep in mind that the `Raise` middlewares should always go last" part makes me feel a little bit itchy — just because I see myself adding a new middleware and screwing up —, but other than that the PR looks good!

lindgrenj6:
Well - we don't necessarily need it to be last either. It just has to be after
authentication/authorization imo. I just added it as last for now :) 

-- 
Jacob Lindgren

Red Hat
***@***.***     M: 402-841-6173

MikelAlejoBR:
Understood, thank you!

middleware/event_stream.go
```diff
@@ -0,0 +1,131 @@
...
+
+			msg, err := json.Marshal(resource)
+			if err != nil {
+				return err
+			}
+
+			// TODO: make this async? Run this in a goroutine that way the
+			// request isn't effectively blocked by kafka .
+			headers := append(getRequestHeaders(c), kafka.Header{Key: "event_type", Value: []byte(eventType)})
+			err = producer.RaiseEvent(eventType, msg, headers)
```
lpichler, line 85: in ruby we are sending also "Records.update" message  with [raise_event_for_update](https://github.com/RedHatInsights/sources-api/blob/0324aaf01a0d8d7949f148dacf0489c4320c9911/app/controllers/api/v1/mixins/update_mixin.rb#L10)

do we need to include it here also ? (it so there can be used method `RaiseEventForUpdate`)

routes.go
```diff
@@ -55,17 +55,17 @@ func setupRoutes(e *echo.Echo) {
 	// Applications
 	v3.GET("/applications", ApplicationList, tenancyWithListMiddleware...)
 	v3.GET("/applications/:id", ApplicationGet, middleware.Tenancy)
-	v3.POST("/applications", ApplicationCreate, permissionMiddleware...)
-	v3.PATCH("/applications/:id", ApplicationEdit, permissionMiddleware...)
-	v3.DELETE("/applications/:id", ApplicationDelete, permissionMiddleware...)
+	v3.POST("/applications", ApplicationCreate, append(permissionMiddleware, middleware.RaiseApplicationCreateEvent)...)
+	v3.PATCH("/applications/:id", ApplicationEdit, append(permissionMiddleware, middleware.RaiseApplicationUpdateEvent)...)
```
lpichler, line 59: it looks that we are sending `.create` event regard to superkey  https://github.com/RedHatInsights/sources-api/blob/3f77590b0a48a4fe122c08cfb9716d7be0976bd0/app/controllers/api/v3x1/applications_controller.rb#L32-L34 

Do we still  need it ?

and

there is  `extra` param added: https://github.com/RedHatInsights/sources-api/blob/3f77590b0a48a4fe122c08cfb9716d7be0976bd0/app/controllers/api/v3x1/applications_controller.rb#L37

lpichler:
@lindgrenj6 I had some comment that some things are different in ruby app otherwise it looks good.

middleware/event_stream.go
lindgrenj6: I will probably add them in follow-up, I'm going to wire up the bulk_create message produce events too and will do it then.

routes.go
lindgrenj6: For superkey I am doing that all in #100, it is in progress.

lpichler (approved):

PR #92, opened 2022-02-01
[RHCLOUD-17798] Update method ToResponse() in models - time fields format when are/are not empty

petracihalova:
https://issues.redhat.com/browse/RHCLOUD-17798

format time fields such as CreatedAt, PausedAt, LastCheckedAt ... when the time field is / is not empty
and ToResponse () updates in all models

----
I updated ToResponse() method in a similar way to the ToEvent () method, but now we have 3 structs for AvailabilityStatus (AvailabilityStatus, AvailabilityStatusEvent, AvailabilityStatusResponse) and last 2 are almost same, maybe we can use only 1 instead of these 2 and save this new one in model/common.go ... but not sure about name ... any ideas?


model/application.go
```diff
@@ -57,17 +57,22 @@ func (app *Application) ToResponse() *ApplicationResponse {
...
-		CreatedAt:               app.CreatedAt,
-		UpdatedAt:               app.UpdatedAt,
-		Pause:                   app.Pause,
-		AvailabilityStatusError: app.AvailabilityStatusError,
-		Extra:                   app.Extra,
-		SourceID:                sourceId,
-		ApplicationTypeID:       appTypeId,
+		AvailabilityStatusResponse: asResponse,
+		ID:                         id,
+		CreatedAt:                  util.DateTimeToRecordFormat(app.CreatedAt),
```
MikelAlejoBR: I think the `CreatedAt` columns are automatically filled with a value upon creation, so I guess that we could leave the `CreatedAt` members as non-pointers

model/common.go
```diff
@@ -16,3 +11,13 @@ type AvailabilityStatus struct {
...
+
+type PauseResponse struct {
+	PausedAt *string `json:"paused_at,omitempty"`
+}
+
+type AvailabilityStatusResponse struct {
+	AvailabilityStatus *string `json:"availability_status,omitempty"`
+	LastCheckedAt      *string `json:"last_checked_at,omitempty"`
+	LastAvailableAt    *string `json:"last_available_at,omitempty"`
+}
```
MikelAlejoBR: I was wondering if we could simply change the `AvailabilityStatus` struct above and change the inner members to pointers instead? That way if a `LastCheckedAt` column is `null` in the database, GORM would set it as `nil` here. This way we wouldn't need helper functions to transform the default values to `nil`

model/application.go
lindgrenj6: Correct, [gorm handles it](https://gorm.io/docs/conventions.html#CreatedAt)

model/application_authentication.go
```diff
@@ -49,8 +49,8 @@ func (aa *ApplicationAuthentication) ToResponse() *ApplicationAuthenticationResp
 	return &ApplicationAuthenticationResponse{
 		ID:                id,
 		AuthenticationUID: aa.AuthenticationUID,
-		CreatedAt:         aa.CreatedAt,
-		UpdatedAt:         aa.UpdatedAt,
+		CreatedAt:         util.DateTimeToRecordFormat(aa.CreatedAt),
+		UpdatedAt:         util.DateTimeToRecordFormat(aa.UpdatedAt),
```
lindgrenj6: So for all of the timestamps that come back in HTTP responses it needs to be `time.RFC1399`, see a v1.0 api response here:
```
$> curl -s -u insights-qa:***** 'https://console.stage.redhat.com/api/sources/v1.0/applications/87919' |jq  
{
  "application_type_id": "3",
  "availability_status": "unavailable",
  "availability_status_error": "Could not enable cloud metering. Error code CG1002. AWS account ID must be unique, but another cloud account with the same AWS account ID already exists.",
  "created_at": "2021-12-08T13:38:39Z",
  "extra": {},
  "id": "87919",
  "source_id": "96814",
  "updated_at": "2021-12-08T13:38:52Z",
  "tenant": "6089719"
}
```

By default we're using `RFC1339Nano` which is fine, but we probably just want to use regular RFC1339.

So I would just make another helper funciton `util.DateTimeToRFC1339(time.Time)`.

List of constants we have here: https://pkg.go.dev/time#pkg-constants

model/application_authentication_http.go
```diff
@@ -1,13 +1,9 @@
...
-	"time"
-)
-
 type ApplicationAuthenticationResponse struct {
-	ID        string    `json:"id"`
-	CreatedAt time.Time `json:"created_at"`
-	UpdatedAt time.Time `json:"updated_at"`
+	ID        string  `json:"id"`
+	CreatedAt *string `json:"created_at"`
+	UpdatedAt *string `json:"updated_at"`
```
lindgrenj6: Like mikel said i think created/updated at are always going to be there due to gorm so they don't have to be pointers.

model/common.go
lindgrenj6: I agree with this, if we made them all pointers then they wouldn't get stored in the database when they aren't there which might solve all of our problems.

petracihalova: question ... when inner members will be pointers (*time.Time), we still need to format them using layout RFC3339 (as was in comment for CreatedAt and UpdatedAt fields)? if so, I cannot use new function util.DateTimeToRFC3339() because CreatedAt and UpdatedAt are not pointers ... so I need to create new function or it is possible to update this existing one to handle pointers and non-pointers values?

model/application_authentication.go
petracihalova: function created in util/parser.go

model/application_authentication_http.go
petracihalova: done

model/application.go
petracihalova: done

model/application.go
```diff
@@ -57,17 +57,22 @@ func (app *Application) ToResponse() *ApplicationResponse {
 	id := strconv.FormatInt(app.ID, 10)
 	sourceId := strconv.FormatInt(app.SourceID, 10)
 	appTypeId := strconv.FormatInt(app.ApplicationTypeID, 10)
+	asResponse := AvailabilityStatusResponse{
+		AvailabilityStatus: util.StringValueOrNil(app.AvailabilityStatus.AvailabilityStatus),
+		LastCheckedAt:      util.DateTimeToRecordFormat(app.LastCheckedAt),
+		LastAvailableAt:    util.DateTimeToRecordFormat(app.LastAvailableAt),
```
lindgrenj6: These 2 need changed to RFC3339.

model/common.go
lindgrenj6: You should be able to make `util.DateTimeToRFC3339()` accept an interface as a param instead of a `time.Time` struct, and do the right thing based on the type. Just do a nil check and return nil if the time is nil.

Code Example here: https://go.dev/play/p/g1ZZU5I8wlS

model/endpoint.go
```diff
@@ -64,23 +64,28 @@ func (endpoint *Endpoint) ToEvent() *EndpointEvent {
 func (endpoint *Endpoint) ToResponse() *EndpointResponse {
 	id := strconv.FormatInt(endpoint.ID, 10)
 	sourceId := strconv.FormatInt(endpoint.SourceID, 10)
+	asResponse := AvailabilityStatusResponse{
+		AvailabilityStatus: util.StringValueOrNil(endpoint.AvailabilityStatus.AvailabilityStatus),
+		LastCheckedAt:      util.DateTimeToRecordFormat(endpoint.LastCheckedAt),
+		LastAvailableAt:    util.DateTimeToRecordFormat(endpoint.LastAvailableAt),
```
lindgrenj6: Update these to RFC3339 :pray: too

model/source.go
```diff
@@ -70,19 +70,24 @@ func (src *Source) ToEvent() *SourceEvent {
 func (src *Source) ToResponse() *SourceResponse {
 	id := strconv.FormatInt(src.ID, 10)
 	stid := strconv.FormatInt(src.SourceTypeID, 10)
+	asResponse := AvailabilityStatusResponse{
+		AvailabilityStatus: util.StringValueOrNil(src.AvailabilityStatus.AvailabilityStatus),
+		LastCheckedAt:      util.DateTimeToRecordFormat(src.LastCheckedAt),
+		LastAvailableAt:    util.DateTimeToRecordFormat(src.LastAvailableAt),
```
lindgrenj6: And last one here.

util/parser.go
```diff
@@ -76,3 +76,11 @@ func FormatTimeToString(inputTime time.Time, format string) string {
 func DateTimeToRecordFormat(inputTime time.Time) *string {
 	return StringValueOrNil(FormatTimeToString(inputTime, RecordDateTimeFormat))
 }
+
+func DateTimeToRFC3339(inputTime time.Time) string {
+	if inputTime.IsZero() {
+		return ""
+	}
+
```
lindgrenj6: I like this check - add it to the updated function - if it is zero return nil.

lindgrenj6 (commented):
Getting closer, just a few places to update + refactor the RFC3339 helper.

util/parser.go
```diff
@@ -76,3 +76,11 @@ func FormatTimeToString(inputTime time.Time, format string) string {
...
+	return inputTime.Format(time.RFC3339)
+}
```
MikelAlejoBR: I would replace the body of this function to the following:

```go
return FormatTimeToString(inputTime, time.RFC3339)
```

Essentially it is duplicating what `FormatTimeToString` is doing, and since we already have that function, I would just use it!

MikelAlejoBR (changes requested):

util/parser.go
```diff
@@ -73,6 +76,23 @@ func FormatTimeToString(inputTime time.Time, format string) string {
...
+		}
+		timeField = *v
+
+	case time.Time:
+		timeField = v
+
+	default:
+		l.Log.Errorf("Unexpected data type: expected *time.Time or time.Time, got %s", v)
+		return ""
+	}
```
MikelAlejoBR: I would like to hear the opinion of the rest of the team, but I personally would not give `DateTimeToRFC3339` the responsibility of having to parse the input. I think it would be better to extract this part into another function, and pass the validated value to `DateTimeTo...`:

```go
validTime, err := ValidateDateTime(input) // it doesn't have to be named like this
if err != nil { return err }

return &m.Whatever{
    // ...
    CreatedAt: &DateTimeToRFC3339(validTime)
    // ...
}
```

util/parser.go
```diff
@@ -73,6 +76,23 @@ func FormatTimeToString(inputTime time.Time, format string) string {
 	return inputTime.Format(format)
 }
 
-func DateTimeToRecordFormat(inputTime time.Time) *string {
```
lindgrenj6, line 76: We can't remove this function either - this one was wrapping a specific format (in this case the format represented by the `RecordDateTimeFormat` at the top of the file), I was moer implying we should have a similar function that does something like that...but with RFC3339

util/parser.go
lindgrenj6: I guess I'm confused - all this function is doing is transforming an interface from many types down to a time.Time that we can format. What would we pull out?

util/parser.go
```diff
@@ -73,6 +76,23 @@ func FormatTimeToString(inputTime time.Time, format string) string {
...
-	return StringValueOrNil(FormatTimeToString(inputTime, RecordDateTimeFormat))
+func DateTimeToRFC3339(inputTime interface{}) string {
+	var timeField time.Time
+	switch v := inputTime.(type) {
+
+	case *time.Time:
+		if reflect.ValueOf(inputTime).IsNil() {
```
lindgrenj6: Luckily you don't have to use reflection here, you can just have a `case nil:` since an interface{} can be the nil type.

model/source.go
```diff
@@ -45,15 +45,15 @@ type Source struct {
 
 func (src *Source) ToEvent() interface{} {
 	asEvent := AvailabilityStatusEvent{AvailabilityStatus: util.StringValueOrNil(src.AvailabilityStatus.AvailabilityStatus),
-		LastAvailableAt: util.DateTimeToRecordFormat(src.LastAvailableAt),
-		LastCheckedAt:   util.DateTimeToRecordFormat(src.LastCheckedAt)}
```
lindgrenj6: Sooooo for the `ToEvent()` methods we need to use the Record Format like before, basically we have different HTTP date formats compared to event-stream events. Some finnicky stuff that rails definitely did that makes 0 sense looking back. 

Maybe we'll standardize them eventually, but should make them match the rails api for now.

util/parser.go
petracihalova: I was thinking, that something like this you mean when you wrote in comment above "You should be able to make util.DateTimeToRFC3339() accept an interface as a param instead of a time.Time struct, and do the right thing based on the type. Just do a nil check and return nil if the time is nil."

model/source.go
petracihalova: ok so I'll undo the changes

lindgrenj6: Just on the calls in the `<Struct>.ToResponse()` methods

util/parser.go
MikelAlejoBR: I forgot to add the `DateTimeToRFC3339` function in the code sample :man_facepalming: . My point was that, in my opinion, the `DateTimeToRFC3339`'s responsibility should only be "I take a time, and I convert its format to RFC 3339", not "I take something, try to see if its a valid `time.Time`, and if it is, convert its format to RFC 3339". But never mind, I was just asking what you two thought about it :+1:

petracihalova:
I removed last commit and this new commit contains only change of format function for fields `LastAvailableAt` and `LastCheckedAt` and small refactor of `DateTimeToRFC3339()`
TODO: change of inner members of `AvailabilityStatus` struct to pointers and related changes

or maybe we can let it as it is now, because when I change inner fields in `AvailabilityStatus `struct to pointers, this affects not only `ToResponse() `but `ToEvent() `methods too, and there we use for format date fields function `util.DateTimeToRecordFormat() `which expects `time.Time` and not pointers + then we don't need to change `DateTimeToRFC3339()` to expect interface{} 

what do you think? @lindgrenj6 @MikelAlejoBR @lpichler

model/application.go
petracihalova: done

model/endpoint.go
petracihalova: done

MikelAlejoBR:
I would personally vote for the pointers option, especially if we are talking about a response, because that way the client doesn't receive an invalid datetime. I mean, instead of receiving something like `last_checked_at: "0001-01-01T00:00:00"` they wouldn't even receive the `last_checked_at` key, which reduces noise.

petracihalova:
> I would personally vote for the pointers option, especially if we are talking about a response, because that way the client doesn't receive an invalid datetime. I mean, instead of receiving something like `last_checked_at: "0001-01-01T00:00:00"` they wouldn't even receive the `last_checked_at` key, which reduces noise.

but that's how it works ... now (without pointers) the format function returns empty string and because we have in structs "omitempty" these empty fields are not in response

example: 
response before this change -` /api/sources/v3.1/applications/2`
```
{
    "availability_status": "available",
    "last_checked_at": "0001-01-01T00:00:00Z",
    "last_available_at": "0001-01-01T00:00:00Z",
    "paused_at": "0001-01-01T00:00:00Z",
    "id": "2",
    "created_at": "2021-11-05T08:39:12.06716Z",
    "updated_at": "2021-11-05T08:39:12.06716Z",
    "extra": {},
    "source_id": "3",
    "application_type_id": "2"
}
```

after this change
```
{
    "availability_status": "available",
    "id": "2",
    "created_at": "2021-11-05T08:39:12Z",
    "updated_at": "2021-11-05T08:39:12Z",
    "extra": {},
    "source_id": "3",
    "application_type_id": "2"
}
```

petracihalova:
I'm not saying we don't need to change the inner members to pointers ... but I think - because it will affects many places in code because it is related with ToResponse() and ToEvent() methods - it can be solved as new task in new PR

MikelAlejoBR:
> but that's how it works ... now (without pointers) the format function returns empty string and because we have in structs "omitempty" these empty fields are not in response

Aaaah yeah, that's true! I forgot about the `omitempty` bit.

> I'm not saying we don't need to change the inner members to pointers ... but I think - because it will affects many places in code because it is related with ToResponse() and ToEvent() methods - it can be solved as new task in new PR

Well, certainly I wouldn't mind having another PR so we can "start fresh"...

petracihalova:
merge conflicts solved

lindgrenj6:
Yeah - I agree. Going through this has showed there are a few things I think we should change mostly related to using struct embeds like `AvailabilityStatus`, `Pause`, etc. I think they're unnecessary and just add complication.

In the next PR can we remove those and just have regular fields? It'll make everything so much easier.

lindgrenj6 (approved):

petracihalova:
> Yeah - I agree. Going through this has showed there are a few things I think we should change mostly related to using struct embeds like `AvailabilityStatus`, `Pause`, etc. I think they're unnecessary and just add complication.
> 
> In the next PR can we remove those and just have regular fields? It'll make everything so much easier.

@lindgrenj6 does it mean that e.g. instead of 

```
type ApplicationResponse struct {
	AvailabilityStatus
	Pause

	ID        string    `json:"id"`
	CreatedAt time.Time `json:"created_at"`
	UpdatedAt time.Time `json:"updated_at"`
        .....
}
```

we want to have

```
type ApplicationResponse struct {
	AvailabilityStatus string    `json:"availability_status,omitempty"`
	LastCheckedAt      time.Time `json:"last_checked_at,omitempty"`
	LastAvailableAt    time.Time `json:"last_available_at,omitempty"`
	
        PausedAt time.Time `json:"paused_at,omitempty"`

	ID        string    `json:"id"`
	CreatedAt time.Time `json:"created_at"`
	UpdatedAt time.Time `json:"updated_at"`
        .....
}
```

lindgrenj6:
Yes - thats correct. The struct embeds just make everything harder to use. 

We'd pull them out of the "concrete" structs too (`Application`, `Source`, etc). But like I said that could probably be done in a follow-up instead of this one.

petracihalova:
> Yes - thats correct. The struct embeds just make everything harder to use.
> 
> We'd pull them out of the "concrete" structs too (`Application`, `Source`, etc). But like I said that could probably be done in a follow-up instead of this one.

yes ... I was asking because now we have this in our heads and now we "know" ... and I will prepare new ticket with description and assign it to myself ... so we can merge this PR now

PR #93, opened 2022-02-01
[RHCLOUD-17832] Internal endpoint for source listing

MikelAlejoBR:
Creates an internal endpoint to be able to list the sources along with their tenants. The code is very similar to any other "List" code, except for:

* The internal function doesn't need a DAO with a `tenantId`, since we're not filtering by that for this endpoint.
* The query just selects the columns that are needed for the response.
* The response only includes the `id`, `availability_status` and `tenant` fields; the ones that `sources-monitor-go` uses.

## Links
[RHCLOUD-17832](https://issues.redhat.com/browse/RHCLOUD-17832)


dao/source_dao.go
```diff
@@ -61,6 +61,24 @@ func (s *SourceDaoImpl) List(limit, offset int, filters []util.Filter) ([]m.Sour
 	return sources, count, result.Error
 }
 
+func (s *SourceDaoImpl) ListInternal(limit, offset int, filters []util.Filter) ([]m.Source, int64, error) {
+	query := DB.Table("sources").Select("id", "availability_status", "tenant_id")
```
lpichler: ~1. Why only 3 fields ?~
I guess it is because of source-monitor as I saw bellow.

2. We prefer (at least as we are using it everywhere) `DB.Model( &m.Source{})` instead `DB.Table..`

private/openapi-3-v2.0.yaml
```diff
@@ -0,0 +1,212 @@
+openapi: 3.0.0
```
lpichler: why we use yaml here as we are using jsons so far(in ruby project, in public) ?

source_handlers_test.go
```diff
@@ -275,6 +276,85 @@ func TestSourceListSatellite(t *testing.T) {
...
+
+	if out.Meta.Offset != 0 {
+		t.Error("offset not set correctly")
+	}
+
+	if len(out.Data) != len(fixtures.TestSourceData) {
+		t.Error("not enough objects passed back from DB")
+	}
+
+	for i, src := range out.Data {
```
lpichler: maybe do we need to sort those `out.Data`  for sure ? function `SortByStringValueOnKey("id", out.Data)` could be used for it. - I don't whether that there is defined order for `out.Data`

model/source.go
```diff
@@ -86,3 +86,17 @@ func (src *Source) ToResponse() *SourceResponse {
...
+
+// ToInternalResponse returns only the fields that "sources-monitor-go" requires.
+func (src *Source) ToInternalResponse() *SourceInternalResponse {
+	id := strconv.FormatInt(src.ID, 10)
+	tenantId := strconv.FormatInt(src.TenantID, 10)
+
+	source := &SourceInternalResponse{
+		Id:                 &id,
+		AvailabilityStatus: &src.AvailabilityStatus.AvailabilityStatus,
+		TenantId:           &tenantId,
```
lpichler: from request on ruby app - it looks that we are exposing `Tenant.ExternalTenant` for attribute "tenant" in response .

dao/source_dao.go
lindgrenj6: Yeah - since sources-monitor only requires those three fields and this was previously undocumented we're going to lock it down to only the 3 necessary fields, we can always expose more if needbe!

source_handlers.go
```diff
@@ -325,3 +325,37 @@ func SourceCheckAvailability(c echo.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	// When listing sources via cert-auth we want to lock them down to only the
+	// satellite source type.
+	if c.Get("cert-auth") != nil {
+		satelliteId := strconv.Itoa(int(dao.Static.GetSourceTypeId("satellite")))
+		filters = append(filters, util.Filter{Name: "source_type_id", Value: []string{satelliteId}})
+	}
```
lindgrenj6: This check probably isn't necessary for the internal endpoint, we can remove it!

dao/source_dao.go
MikelAlejoBR: I changed it to `DB.Model(&m.Source{})` but kept the `.Select` part, so we don't fetch more columns than necessary! Thanks for pointing it out.

private/openapi-3-v2.0.yaml
MikelAlejoBR: As I was struggling with editing the spec in JSON, I thought "well, we might as well switch to YAML since it's less painful". I didn't remember that we need to have the spec in JSON format because we're serving it on an endpoint, and for the sake of consistency, I'm transforming the private spec back too to JSON.

source_handlers.go
MikelAlejoBR: All right, thanks!

model/source.go
MikelAlejoBR: You are right. I tried it by setting up the Rails app on the local environment and we're returning the external tenant. I changed that.

source_handlers_test.go
MikelAlejoBR: I'm sorry: I've tried to find the function you told me about, but I am not able to find it :eyes: 

Anyway: in this case Gorm seems to be respecting the order we set in the fixtures, so we shouldn't have any problem with the tests. Although we could create a `TODO` for this to be safer. What do you think?

source_handlers.go
```diff
@@ -325,3 +325,30 @@ func SourceCheckAvailability(c echo.Context) error {
 
 	return c.JSON(http.StatusAccepted, map[string]interface{}{})
 }
+
+func InternalSourceList(c echo.Context) error {
```
lindgrenj6: Might I suggest moving this handlerFunc to a different file? Maybe could have an `internal_handlers.go` file - that way we can separate out the 2 internal endpoints there.  

What do you think?

lindgrenj6 (approved):
Looks good, just one tiny not-important thought I had while reviewing it.

source_handlers.go
MikelAlejoBR: I like the idea. Applying it now.

lpichler (approved):
LGTM,
@MikelAlejoBR thanks!

PR #95, opened 2022-02-01
Use go:embed to embed the seed yaml files

lindgrenj6:
So after merging seeding we are seeing a `CrashLoopBackOff` in stage - related to the seed files not being present in the container. This is due to the fact I didn't copy them into the actual container after building.

Rather than copy them into the container and add another layer I decided to use the [embed](https://pkg.go.dev/embed) package which embeds files into the binary itself, it is pretty cool. We just read from it like a normal filesystem but we don't have to worry about which directory we're in etc. The tests are more effective now as well since the tests are checking to see that the files that were on the FS match the ones that were embedded + seeded into the database.

example logs from the pod.:
```
{"@timestamp":"2022-02-01T16:31:46.779Z","@version":1,"app":"source-api-go","caller":"github.com/RedHatInsights/sources-api-go/dao.seedDatabaseFromDirectory","filename":"/build/dao/seeding.go:23","hostname":"sources-api-go-svc-7d49cb6ff6-mndvw","labels":{"app":"source-api-go"},"level":"info","message":"Seeding SourceType Table","tags":["source-api-go"]}
{"@timestamp":"2022-02-01T16:31:46.78Z","@version":1,"app":"source-api-go","caller":"github.com/RedHatInsights/sources-api-go/dao.Init","filename":"/build/dao/db.go:63","hostname":"sources-api-go-svc-7d49cb6ff6-mndvw","labels":{"app":"source-api-go"},"level":"fatal","message":"Failed to seed db: open ./dao/seeds/source_types.yml: no such file or directory","tags":["source-api-go"]}
```

### DIFF
--- a/dao/seeding.go
+++ b/dao/seeding.go
@@ -1,6 +1,7 @@
 package dao
 
 import (
+	"embed"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (approved):

lpichler (approved):
LGTM, it is good to know about embed 👍

PR #96, opened 2022-02-01
Use loop with counter in ToResponse conversion

lpichler:
This converts loops like this:

```go
for i, a := range applications {
  out[i] = *a.ToResponse()
}
```

to 

```go
for i := 0; i < len(applications); i++ {
  out[i] = applications[i].ToResponse()
}

```

`for i, a := range` - fields in `a` struct has same address and in this case we stores pointers in `out[I]` and when new value is assigned inside loop to `a` it is reflected in all elements in `out`. 

Issue is reproducible in endpoint `/api/sources/v3.1/sources` when names are same  (I am not sure why only `name`) but this change prevents this behaviour.

### DIFF
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -58,8 +58,8 @@ func ApplicationList(c echo.Context) error {
 	c.Logger().Infof("tenant: %v", *applicationDB.Tenant())
 
 	out := make([]interface{}, len(applications))
-	for i, a := range applications {
-		out[i] = *a.ToResponse()
+	for i := 0; i < len(applications); i++ {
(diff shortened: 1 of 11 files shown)


lpichler:
cc @MikelAlejoBR

lindgrenj6 (approved):

MikelAlejoBR:
Yes, I had a fun day yesterday trying to debug what was going on, and why my tests were failing for the internal endpoint. After debugging the database queries, fixtures and following every pointer I was using, it seemed to be an issue with the `for ... range` loop. It was a fun thing to learn indeed...

Thanks for the PR!

PR #97, opened 2022-02-02
[RHCLOUD-17832] Internal authentication endpoint

MikelAlejoBR:
Creates a new internal "Authentications" endpoint which exposes the password of the authentication.

* It also removes the `password` field from the public `AuthenticationResponse` objects.

## Links

[RHCLOUD-17832](https://issues.redhat.com/browse/RHCLOUD-17832)

### DIFF
--- a/internal_handlers.go
+++ b/internal_handlers.go
@@ -8,6 +8,21 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
+// InternalAuthenticationGet fetches one authentication and returns it with the password exposed. Internal use only.
+func InternalAuthenticationGet(c echo.Context) error {
+	authDao, err := getAuthenticationDao(c)
+	if err != nil {
+		return err
+	}
+
+	auth, err := authDao.GetById(c.Param("uuid"))
+	if err != nil {
+		return c.JSON(http.StatusNotFound, util.ErrorDoc(err.Error(), "404"))
+	}
+
+	return c.JSON(http.StatusOK, auth.ToInternalResponse())
+}
+
 // InternalSourceList lists all the sources in a compact format —since the client that will use it,
 // "sources-monitor-go" only requires a small set of fields—.
 func InternalSourceList(c echo.Context) error {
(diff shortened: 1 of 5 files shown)


lindgrenj6 (approved):

internal_handlers.go
```diff
@@ -8,6 +8,21 @@ import (
...
 )
 
+// InternalAuthenticationGet fetches one authentication and returns it with the password exposed. Internal use only.
+func InternalAuthenticationGet(c echo.Context) error {
+	authDao, err := getAuthenticationDao(c)
+	if err != nil {
+		return err
+	}
+
+	auth, err := authDao.GetById(c.Param("uuid"))
```
lindgrenj6, line 18: Sorry I already approved but forgot to comment, we need this functionality hidden behind that query parameter `expose_encrypted_attributes[]=password` basically do something similar to the way i did in the filtering middleware.

Heres an example request I make from the superkey worker: https://github.com/RedHatInsights/sources-superkey-worker/blob/d076299a7d68563aa563256c37c8d846412f44d3/sources/api_client.go#L58-L63

lindgrenj6: Hm. actually maybe since it wasn't documented we can change it. Let me chew on it a bit more.

lpichler: Does this require any other change ?

MikelAlejoBR:
I added a small patch to it so the rewrite doesn't break the current implementations that are passing the query parameter. Nonetheless, what's the purpose of the query parameter?

lindgrenj6:
> I added a small patch to it so the rewrite doesn't break the current implementations that are passing the query parameter. Nonetheless, what's the purpose of the query parameter?

Basically if we had multiple encrypted fields we could expose them selectively. For us there is only one so it makes less sense. Basically just an extra layer of security by obscurity for us afaik.

lindgrenj6 (approved):

MikelAlejoBR:
Thank you for the explanation! :smiley:

PR #98, opened 2022-02-02
Add authentications to records.update event in status listener

lpichler:
- get account number from status message's header
- this PR adds method `(a *AuthenticationDaoImpl) BulkMessage` to build `records.update` bulk messages emitted after `Authentication` status messages
-  on others (For source, endpoint ,..) bulk messages are populated "authentications" relations

## Links
Pulled out from https://github.com/RedHatInsights/sources-api-go/pull/76  and this is  6/7 PR

### DIFF
--- a/dao/common.go
+++ b/dao/common.go
@@ -73,11 +73,10 @@ func BulkMessageFromSource(source *m.Source, authentication *m.Authentication) (
 		return nil, err
 	}
 
-	authentications := make([]m.AuthenticationEvent, len(authenticationsByResource))
-
-	for i, authenticationByResource := range authenticationsByResource {
-		authenticationByResource.Tenant = source.Tenant
-		authentications[i] = *authenticationByResource.ToEvent()
+	authentications := make([]interface{}, len(authenticationsByResource))
+	for i := 0; i < len(authenticationsByResource); i++ {
+		authenticationsByResource[i].Tenant = source.Tenant
+		authentications[i] = authenticationsByResource[i].ToEvent()
 	}
 
 	bulkMessage["authentications"] = authentications
--- a/dao/common.go
+++ b/dao/common.go
@@ -6,6 +6,11 @@ import (
 	m "github.com/RedHatInsights/sources-api-go/model"
 )
 
+const (
+	defaultLimit  = 100
+	defaultOffset = 0
+)
+
 func GetFromResourceType(resourceType string) (*m.EventModelDao, error) {
 	var resource m.EventModelDao
 	switch resourceType {
@@ -24,7 +29,15 @@ func GetFromResourceType(resourceType string) (*m.EventModelDao, error) {
 	return &resource, nil
 }
 
-func BulkMessageFromSource(source *m.Source) (map[string]interface{}, error) {
+/*
+	Method generates bulk message for Source record.
+	authentication - specify resource (ResourceID and ResourceType) of
+                     which authentications are fetched to BulkMessage
+                   - specify application_authentications in BulkMessage otherwise
+                     application_authentications are obtained from authentications UIDs
+					 in BulkMessage
+*/
+func BulkMessageFromSource(source *m.Source, authentication *m.Authentication) (map[string]interface{}, error) {
 	result := DB.
 		Preload("Tenant").
 		Preload("Applications.Tenant").
@@ -52,8 +65,22 @@ func BulkMessageFromSource(source *m.Source) (map[string]interface{}, error) {
 
 	bulkMessage["applications"] = applications
 
-	bulkMessage["authentications"] = []m.Authentication{}
 	bulkMessage["application_authentications"] = []m.ApplicationAuthenticationEvent{}
+
+	authDao := &AuthenticationDaoImpl{TenantID: &source.TenantID}
+	authenticationsByResource, err := authDao.AuthenticationsByResource(authentication)
+	if err != nil {
+		return nil, err
+	}
+
+	authentications := make([]m.AuthenticationEvent, len(authenticationsByResource))
+
+	for i, authenticationByResource := range authenticationsByResource {
+		authenticationByResource.Tenant = source.Tenant
+		authentications[i] = *authenticationByResource.ToEvent()
+	}
+
+	bulkMessage["authentications"] = authentications
 
 	return bulkMessage, nil
 }
--- a/dao/common.go
+++ b/dao/common.go
@@ -7,8 +7,8 @@ import (
 )
 
 const (
-	defaultLimit  = 100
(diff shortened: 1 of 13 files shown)


dao/common.go
```diff
@@ -52,8 +65,22 @@ func BulkMessageFromSource(source *m.Source) (map[string]interface{}, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	authentications := make([]m.AuthenticationEvent, len(authenticationsByResource))
+
+	for i, authenticationByResource := range authenticationsByResource {
+		authenticationByResource.Tenant = source.Tenant
+		authentications[i] = *authenticationByResource.ToEvent()
+	}
```
MikelAlejoBR: I recall having some issues with `for .. range` loops in these cases which were fixed in https://github.com/RedHatInsights/sources-api-go/pull/96. It seems like a very similar situation, but I might be wrong.

util/xrh_header.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+
+	"github.com/RedHatInsights/sources-api-go/kafka"
+	"github.com/redhatinsights/platform-go-middlewares/identity"
+)
+
+func ParseXRHIDHeader(inputIdentity string) (XRHIdentity *identity.XRHID, err error) {
```
MikelAlejoBR: I would suggest using local variables instead of named return values: https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_avoid_named_return_values

dao/common.go
```diff
@@ -6,6 +6,11 @@ import (
 	m "github.com/RedHatInsights/sources-api-go/model"
 )
 
+const (
+	defaultLimit  = 100
+	defaultOffset = 0
```
lindgrenj6: Constants should be in `UPPER_SNAKE_CASE`.

dao/common.go
lindgrenj6: You are right - `authenticationByResource` in this case is a copy of `authenticationsByResource[i]`. Probalby need to just change the loop to a "regular" for loop or just a `for i := range authenticationsByResoruce` and use the index.

dao/tenant_dao.go
```diff
@@ -19,3 +19,13 @@ func GetOrCreateTenantID(accountNumber string) (*int64, error) {
 
 	return &tenant.Id, result.Error
 }
+
+func TenantBy(accountNumber string) (*m.Tenant, error) {
```
lindgrenj6: Rather than using the default parameter name as an extension of the function (not all of us use jetbrains :stuck_out_tongue_closed_eyes: ) can we rename this `TenantByAccountNumber`

util/xrh_header.go
```diff
@@ -0,0 +1,46 @@
...
+
+	err = json.Unmarshal(decodedIdentity, &XRHIdentity)
+	if err != nil {
+		return nil, fmt.Errorf("x-rh-identity header does not contain valid JSON")
+	}
+
+	return XRHIdentity, nil
+}
+
+func AccountNumberFrom(headers []kafka.Header) (string, error) {
```
lindgrenj6: Same comment here - can we rename this `AccountNumberFromHeaders`

statuslistener/statuslistener_test.go
```diff
@@ -138,6 +139,23 @@ func FetchDataFor(resourceType string, resourceID string, forBulkMessage bool) (
...
+		authentication := &m.Authentication{ResourceID: application.ID,
+			ResourceType:               "Application",
+			ApplicationAuthentications: []m.ApplicationAuthentication{},
+		}
+
+		authDao := &dao.AuthenticationDaoImpl{TenantID: &application.TenantID}
+		authenticationsByResource, err := authDao.AuthenticationsByResource(authentication)
+		if err != nil {
+			return err, nil
+		}
```
lindgrenj6: I probably should have seen this - why are we not returning err last? If we're going to return an error type we should always return it last.

util/xrh_header.go
lindgrenj6: I agree with this - my initial thought of this was that the `XRHIdentity` was at the package scope - it was really confusing.

dao/common.go
lpichler: 🤦  I changed it

util/xrh_header.go
lpichler: I agree, named params removed

dao/tenant_dao.go
lpichler: :) I renamed and I will keep suggested pattern next time

statuslistener/statuslistener_test.go
lpichler: method where is `return` even doesn't return error, I will use panic instead as it is code related to tests.

lpichler:
@lindgrenj6 @MikelAlejoBR I added all suggested changes, thanks for good 👀 and review!

MikelAlejoBR (approved):

util/xrh_header.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("x-rh-identity header does not contain valid JSON")
+	}
+
+	return XRHIdentity, nil
+}
+
+func AccountNumberFromHeaders(headers []kafka.Header) (string, error) {
+	XRHAccountNumberKey := "x-rh-sources-account-number"
+	XRHIdentityKey := "x-rh-identity"
```
lindgrenj6: I would move these out of the function scope if you're not wanting to just use them inline.

Maybe we can extract them to a constant and use them from the middleware in a follow-up (thats where they get parsed in the first place)

lindgrenj6 (approved):
Just one teeny little suggestion (so we don't allocate those strings on the stack every time), otherwise looks good!

util/xrh_header.go
lpichler: I moved them from function scope, thanks 

middleware: What's reason to extracting them to middleware ? - is it because we can use them also in middleware ? 
Anyway let's do it in follow up PR.

lindgrenj6: Yes - `middleware/headers.go` is where they get parsed into the context. Thinking about moving that to its own package `middleware/headers/`.  

Either way we can do it in a followup PR.

lindgrenj6 (approved):

lindgrenj6:
The spec failure is the same one as always - I think its just flakey. Merging.

lpichler:
> The spec failure is the same one as always - I think its just flakey. Merging.

yea I know about it is my to-do list. Thanks

PR #100, opened 2022-02-02
Add SuperKey Request Application flow

lindgrenj6:
Implementing the steps for superkey. Which basically involves:

1. After application creation, if the source type is `account_authorization` then we send the superkey create request on kafka
2. After application deletion, send the superkey delete request similar to 1

Pulling in the superkey struct types for this - so they're just defined over in the superkey repo, if we send one marshaled we know it will work over there. Super nice!

\# TODO:
- [x] helper function for getting superkey authentication
- [x] create request
- [x] delete request
- [x] application.extra `_superkey` special handling (populate `SuperKeyData` field on Application)
- [x] superkey destroy middleware
- [x] superkey destroy job


middleware/superkey_destroy.go
```diff
@@ -0,0 +1,87 @@
...
+		if !ok {
+			return fmt.Errorf("failed to pull tenant from request")
+		}
+
+		id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+		if err != nil {
+			return err
+		}
+
+		s := dao.GetSourceDao(&id)
```
MikelAlejoBR: I think this should be `&tenantId`, not the source's `&id`

middleware/superkey_destroy.go
```diff
@@ -0,0 +1,87 @@
...
+		if !ok {
+			return fmt.Errorf("failed to pull tenant from request")
+		}
+
+		id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+		if err != nil {
+			return err
+		}
+
+		a := dao.GetApplicationDao(&id)
```
MikelAlejoBR: I think this should be `&tenantId` too

jobs/superkey.go
```diff
@@ -0,0 +1,105 @@
...
+// Lists all applications for a source, sends the destroy request for each, then
+// enqueues deletion of itself.
+
+// the sub-jobs enqueue deletion of their respective resources.
+func (sk SuperkeyDestroyJob) sendForSource(id int64) error {
+	l.Log.Infof("Sending SuperKey Delete request for source %v", sk.Id)
+
+	a := dao.GetApplicationDao(&sk.Tenant)
+
+	apps, _, err := a.SubCollectionList(&m.Source{ID: id}, 100, 0, make([]util.Filter, 0))
```
MikelAlejoBR, line 59: I just want to double check that 100 is a reasonable limit for this. I am expecting a "yes, it is" response, but I wanted to ask nonetheless :)

MikelAlejoBR (approved):

jobs/superkey.go
lindgrenj6: Yep, most of our sources only have 1 application, the sub-applications (cost, swatch) only support one connection per credential. SO realistically the max is 2.

middleware/superkey_destroy.go
lindgrenj6: Classic "bug introduced by rebase" behavior here :facepalm: I totally did this on accident the first time, then rebased bringing in the DAO changes, and re-introduced the bug. 

The good news is that it would have shown up really quick lol

jobs/superkey.go
MikelAlejoBR: Cool, thank you!

lindgrenj6:
@lpichler could I get some :eyes: on this? It's been sitting for a while now.

service/superkey.go
```diff
@@ -39,7 +38,7 @@ func SendSuperKeyCreateRequest(identity string, application *m.Application) erro
 	}
 
 	req := superkey.CreateRequest{
-		IdentityHeader:  identity,
+		// IdentityHeader:  identity, <- header
```
lpichler: Should this be comment ?

lindgrenj6: It looks like that gets changed in a later commit - it's not present in the current state.

lpichler (approved):
It looks great! I am sorry for delay!

lindgrenj6:
@lpichler no worries, we're all busy. I'll rebase bulk create next and we can hopefully get that one in too.

PR #104, opened 2022-02-07
Change EventType to come from helper function rather than static strings in Middleware

lindgrenj6:
This is a bit better than `append`-ing everywhere like I initially had.

### DIFF
--- a/middleware/event_stream.go
+++ b/middleware/event_stream.go
@@ -10,32 +10,6 @@ import (
 	"github.com/labstack/echo/v4"
 )
 
-// middleware functions for each event type - adding this to the middleware
-// stack will raise an event based on the `resource` field set in the context
-// during the handler
-
-var (
-	RaiseSourceCreateEvent  = raiseEvent("Source.create")
-	RaiseSourceUpdateEvent  = raiseEvent("Source.update")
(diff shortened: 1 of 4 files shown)


MikelAlejoBR (approved):
LGTM!

middleware/event_stream.go
```diff
@@ -47,49 +21,64 @@ var producer = events.EventStreamProducer{Sender: &events.EventStreamSender{}}
...
-			// event.
-			if c.Get("skip_raise") != nil {
-				l.Log.Infof("skipping raise event per skip_raise set on context")
+		eventType, ok := c.Get("event_type").(string)
+		if !ok {
+			l.Log.Warnf("Failed to cast event_type to string - exiting")
+			return nil
+		}
+
+		if c.Get("event_override") != nil {
```
lpichler, line 54: Is there specific usage of `event_override` ?  (or in future)

lpichler (approved):

middleware/event_stream.go
lindgrenj6: Yes - it's only used for superkey right now. on the initial create we do not send the create message since the source isn't ready yet (thew orker has to handle it), but on update we override it and send the create message since the ARN got posted back from the worker.

PR #105, opened 2022-02-08
[RHCLOUD-17958] pause and resume a source

MikelAlejoBR:
Implements the following endpoints:

* `/sources/:source_id/pause`
* `/sources/:source_id/unpause`
* `/applications/:application_id/pause`
* `/applications/:application_id/unpause`

 ## Links

[[RHCLOUD-17958]](https://issues.redhat.com/browse/RHCLOUD-17958)

### DIFF
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -244,9 +244,14 @@ func ApplicationPause(c echo.Context) error {
 		return err
 	}
 
-	application, err := applicationDao.Pause(applicationId)
+	err = applicationDao.Pause(applicationId)
 	if err != nil {
 		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	application, err := applicationDao.GetById(&applicationId)
+	if err != nil {
+		return err
 	}
 
 	// Get the Kafka headers we will need to be forwarding.
@@ -273,9 +278,14 @@ func ApplicationResume(c echo.Context) error {
 		return err
 	}
 
-	application, err := applicationDao.Resume(applicationId)
+	err = applicationDao.Resume(applicationId)
 	if err != nil {
 		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	application, err := applicationDao.GetById(&applicationId)
+	if err != nil {
+		return err
 	}
 
 	// Get the Kafka headers we will need to be forwarding.
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -265,7 +265,7 @@ func ApplicationPause(c echo.Context) error {
 func ApplicationResume(c echo.Context) error {
 	applicationId, err := strconv.ParseInt(c.Param("id"), 10, 64)
 	if err != nil {
-		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+		return util.NewErrBadRequest(err)
 	}
 
 	applicationDao, err := getApplicationDao(c)
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -236,7 +236,7 @@ func SourceListApplications(c echo.Context) error {
 func ApplicationPause(c echo.Context) error {
 	applicationId, err := strconv.ParseInt(c.Param("id"), 10, 64)
 	if err != nil {
-		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+		return util.NewErrBadRequest(err)
 	}
 
 	applicationDao, err := getApplicationDao(c)
@@ -246,7 +246,7 @@ func ApplicationPause(c echo.Context) error {
 
 	err = applicationDao.Pause(applicationId)
 	if err != nil {
-		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+		return util.NewErrBadRequest(err)
 	}
 
 	application, err := applicationDao.GetById(&applicationId)
@@ -280,7 +280,7 @@ func ApplicationResume(c echo.Context) error {
 
 	err = applicationDao.Resume(applicationId)
 	if err != nil {
-		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+		return util.NewErrBadRequest(err)
 	}
 
 	application, err := applicationDao.GetById(&applicationId)
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -231,3 +231,61 @@ func SourceListApplications(c echo.Context) error {
 
 	return c.JSON(http.StatusOK, util.CollectionResponse(out, c.Request(), int(count), limit, offset))
 }
+
+// ApplicationPause pauses a given application by setting its "paused_at" column to "now()".
+func ApplicationPause(c echo.Context) error {
+	applicationId, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	applicationDao, err := getApplicationDao(c)
+	if err != nil {
+		return err
+	}
+
+	application, err := applicationDao.Pause(applicationId)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	// Get the Kafka headers we will need to be forwarding.
+	kafkaHeaders := service.ForwadableHeaders(c)
+
+	// Raise the resume event for the source.
+	err = service.RaiseEvent("Application.Pause", application, kafkaHeaders)
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(http.StatusNoContent, nil)
+}
+
+// ApplicationResume resumes a given application by setting its "paused_at" column to "NULL".
+func ApplicationResume(c echo.Context) error {
+	applicationId, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	applicationDao, err := getApplicationDao(c)
+	if err != nil {
+		return err
+	}
+
+	application, err := applicationDao.Resume(applicationId)
+	if err != nil {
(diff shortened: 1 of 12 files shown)


dao/application_dao.go
```diff
@@ -124,3 +125,21 @@ func (a *ApplicationDaoImpl) ToEventJSON(resource util.Resource) ([]byte, error)
 
 	return data, err
 }
+
+func (a *ApplicationDaoImpl) PauseAll(sourceId int64) error {
+	result := DB.Debug().
+		Model(&m.Application{}).
+		Where("source_id = ?", sourceId).
```
lindgrenj6: Probably want to tack a `.Where("tenant_id = ?", a.TenantID)` just to be safe.

source_handlers.go
```diff
@@ -331,3 +331,65 @@ func SourceCheckAvailability(c echo.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	err = sourceDao.Pause(sourceId)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	return c.JSON(http.StatusNoContent, nil)
```
lindgrenj6, line 413: So during any `Pause` request one side effect is that we need to raise events. I foresee a service class that basically goes through and for source:
1. calls Pause on all the applications for the source
2. raises the event for each

for application:
1. calls pause on the application
2. raises the event

old code here: https://github.com/RedHatInsights/sources-api/blob/7b8737277a1b7fcf0be2d58feb579272b66e7257/app/controllers/api/v3x1/sources_controller.rb#L21-L24


We could have a file in service that basically has 2 functions, `PauseSource` and `PauseApplication` where the PauseSource just calls PauseApplication for each source's applications. 

fwiw - most of the time we pause applications only and not sources.

dao/application_dao.go
MikelAlejoBR: Yeah, good idea!

source_handlers.go
MikelAlejoBR: I thought about doing something reusable with those two functions you are talking about, but I ruled that possibility out due to the N+1 issue. I don't know if we will have many applications related to a source, but it can get out of hand pretty easily.

Instead I was thinking on maybe doing the following:

1. Update all the applications for a source.
2. Update the source itself.
3. Fetch the applications and loop through raising events for every single one of them.

I could wrap the first two operations in a transaction so that if something fails, no fake pause notifications are raised.

Same thing for the applications:

1. Update them.
2. Fetch them.
3. Raise events.

What do you think of this approach? This way we could be done with 3 to 4 queries max.

MikelAlejoBR: Just to clarify, what I'm thinking of is the following thing:

```go
sourceId := 12345
source, applicationList, err := sourceDao.Pause(sourceId) // this pauses the related applications to the source and the source itself in a transaction

if err != nil { return err }

raiseEvent("Source.Pause", source)

for _, app := range applicationList {
    raiseEvent("Application.Pause", app)
} 
```

Or even better:

```go
sourceId := 12345
err := sourceDao.Pause(sourceId) // this pauses the related applications to the source and the source itself in a transaction

if err != nil { return err }

source, err := sourceDao.GetWithRelatedApplications(sourceId) // DB.Model(&Source{}).Preload(&Application{})

if err != nil { return err }

raiseEvent("Source.Pause", source)

for _, app := range source.Applications {
    raiseEvent("Application.Pause", app)
} 
```

-----

And for the applications:

```go
appId := 54321
application, err := applicationDao.Pause(appId) // this pauses the application


if err != nil { return err }

raiseEvent("Application.Pause", application)
```

lindgrenj6: Yeah that sounds good to me. We can probably pull the actual `raiseEvent` piece into a helper function (it's still just a middleware right now).

application_handlers.go
```diff
@@ -233,3 +233,61 @@ func SourceListApplications(c echo.Context) error {
 
 	return c.JSON(http.StatusOK, util.CollectionResponse(out, c.Request(), int(count), limit, offset))
 }
+
+// ApplicationPause pauses a given application by setting its "paused_at" column to "now()".
+func ApplicationPause(c echo.Context) error {
+	applicationId, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
```
petracihalova: Here you can return util.NewErrBadRequest(err). Same below and in source_handler.go. HandleErrors() in middleware recognise err type and prepare JSON with correct status and message.

application_handlers.go
```diff
@@ -233,3 +233,61 @@ func SourceListApplications(c echo.Context) error {
...
+	}
+
+	applicationDao, err := getApplicationDao(c)
+	if err != nil {
+		return err
+	}
+
+	application, err := applicationDao.Pause(applicationId)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
```
petracihalova: Sorry if I am wrong. When applicationDao.Pause() returns Not Found error, then I think here it will be replaced by Bad Request. Same situation in ApplicationResume(). Same in source_handler.go.

source_handlers.go
MikelAlejoBR: I'm resolving this since I've just implemented the changes.

petracihalova (changes requested):

application_handlers.go
MikelAlejoBR: Thank you, nice catch!

application_handlers.go
MikelAlejoBR: Yeah, I initially wrote these functions which only returned `err`, but I was in doubt. After your comment I'm going to go back to that design. Thank you!

lindgrenj6 (approved):
LGTM!

application_handlers.go
petracihalova: not fixed yet 😁

petracihalova: not fixed yet here and on other places

MikelAlejoBR: Whoops sorry, yeah, you are right. I've just pushed the changes to use the utility `NewBadRequestErr`!

petracihalova:
LGTM 😁. I am only curious ... in code I can see response status code is 202 for pause and unpause? and in openapi json we have 204 for pause and 202 for unpause

petracihalova (approved):

MikelAlejoBR:
Nice catch! I'll send a PR to the sources-api repo to fix that one, plus a more descriptive comment for pausing the sources —since pausing a source also pauses its applications, but the docs don't mention that—. 

Thanks!

dao/interfaces.go
```diff
@@ -24,6 +24,12 @@ type SourceDao interface {
 	BulkMessage(resource util.Resource) (map[string]interface{}, error)
 	FetchAndUpdateBy(resource util.Resource, updateAttributes map[string]interface{}) error
 	ToEventJSON(resource util.Resource) ([]byte, error)
+	// Pause pauses the given source and all its dependant applications. It returns the source with all the dependant
+	// applications preloaded, and an error if the source doesn't exist.
+	Pause(id int64) (*m.Source, error)
+	// Resume resumes the given source and all its dependant applications. It returns the source with all the dependant
+	// applications preloaded, and an error if the source doesn't exist.
+	Resume(id int64) (*m.Source, error)
```
lpichler: I don't know whether word `Resume` is more confusing here as we have event unpause - (and thanks to that it is not consistent) - it is obvious that unpause is opposite action of Pause and in word Resume it is less obvious.

application_handlers.go
```diff
@@ -231,3 +231,61 @@ func SourceListApplications(c echo.Context) error {
...
+	application, err := applicationDao.Resume(applicationId)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	// Get the Kafka headers we will need to be forwarding.
+	kafkaHeaders := service.ForwadableHeaders(c)
+
+	// Raise the resume event for the source.
+	err = service.RaiseEvent("Application.Unpause", application, kafkaHeaders)
```
lpichler, line 295: in rails app in `XXX.Pause/Unpause` is unpause/pause in lowercase but I don't know if it does matter. cc @lindgrenj6

dao/interfaces.go
lpichler: + we have also requests called with `...:/id/unpause`

lpichler:
I know I am late but is it possible to include in tests that `RaiseEvent` is called with proper arguments ? I am just asking but it would be great to have coverage also here. Calling `RaiseEvent` causes error which is not returned(only sent to logs)

lpichler:
as we have now pausing - we should also disallow updating for paused resources except to few attributes - see more https://github.com/RedHatInsights/sources-api/pull/421 (there is also JIRA ticket) - I don't know whether there is JIRA GO ticket for it.

application_handlers.go
lindgrenj6: It does matter - fixing here: https://github.com/RedHatInsights/sources-api-go/pull/149

dao/interfaces.go
lindgrenj6: Yeah I can't remember now since it's been so long but I think initially they wanted pause/resume but it changed to pause/unpause on the endpoints.

MikelAlejoBR: It is true that for the sake of consistency I should have kept it as "Unpause", but since I was only changing the internal naming, I thought it was more clear to use the word "resume", since it seems to be the opposite of pausing something. However I'll open a PR refactoring that.

application_handlers.go
MikelAlejoBR: Great catch! :+1:

MikelAlejoBR:
> I know I am late but is it possible to include in tests that `RaiseEvent` is called with proper arguments ? I am just asking but it would be great to have coverage also here. Calling `RaiseEvent` causes error which is not returned(only sent to logs)

All right, I'll add some tests in the refactor to check that.


> as we have now pausing - we should also disallow updating for paused resources except to few attributes - see more [RedHatInsights/sources-api#421](https://github.com/RedHatInsights/sources-api/pull/421) (there is also JIRA ticket) - I don't know whether there is JIRA GO ticket for it.

I'll create one and add it in the refactor too.

Thanks!

lpichler:
Great, thanks @MikelAlejoBR !!!

PR #111, opened 2022-02-10
[RHCLOUD-17967] Implement Red Hat Connector Connections

MikelAlejoBR:
Implements the endpoints for the Red Hat Connector connections. The implemented endpoints are the following ones:

- GET "/rhc_connections
- GET "/rhc_connections/:id"
- POST "/rhc_connections"
- PATCH "/rhc_connections/:id"
- DELETE "/rhc_connections/:id"

And the subcollections:

- GET "/sources/:source_id/rhc_connections"
- GET "/rhc_connections/:id/sources"

The DAOs need to join a couple of tables to be able to fetch all the sources associated to the connections. I couldn't find how to fetch the generated column which contains the related ids, so I retorted to manually mapping the rows to the corresponding structs.

Other than that I wrote a lot of tests, which is why the number of additions in this PR is ridiculously big.

## Links

[[RHCLOUD-17967]](https://issues.redhat.com/browse/RHCLOUD-17967)


model/event_models.go
```diff
@@ -105,3 +105,14 @@ type SourceEvent struct {
 	SourceTypeID        *int64  `json:"source_type_id"`
 	Tenant              *string `json:"tenant"`
 }
+
+type RhcConnectionEvent struct {
+	RhcId *string        `json:"uuid"`
```
lindgrenj6: So I think we're going to json-this as `rhd_id`.

model/rhc_connection_http.go
```diff
@@ -0,0 +1,35 @@
...
+
+// RhcConnectionUpdateRequest represents a request coming from the outside to update a Red Hat Connector connection.
+type RhcConnectionUpdateRequest struct {
+	Extra datatypes.JSON `json:"extra"`
+}
+
+// RhcConnectionListResponse doesn't include the source ID, since it would be very expensive to get all the sources for
+// all the RhcConnections that a user might want to list.
+type RhcConnectionListResponse struct {
+	Uuid  *string        `json:"uuid"`
```
lindgrenj6: Sooo this will be `rhc_id`.

dao/rhc_connection_dao.go
```diff
@@ -0,0 +1,192 @@
...
+
+	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/util"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+type RhcConnectionDaoImpl struct{}
+
+func (s *RhcConnectionDaoImpl) List(limit, offset int, filters []util.Filter) ([]m.RhcConnection, int64, error) {
```
lindgrenj6, line 18: So for this one it is a bit more complicated than this - we don't have tenancy on the table itself but we don't want to expose _every_ rhc_connection to every tenant out there.

It gets a bit complicated but we basically need to fetch the sources for that tenant and list the rhc_connections for each of those.

The good news is that most people are most likely going to be doing `/sources/:id/rhc_connections` so we won't have to do this listing very often and filtering is less important since it'll most likely be a 1:1 relationship between sources/rhc_connections other than the very few exceptions.

---

The SQL is probably going to look something like this:

```sql
-- List RHC Connections for a Source ID
SELECT *
FROM rhc_connections rc
         INNER JOIN source_rhc_connections srhc ON rc.id = srhc.rhc_connection_id
WHERE srhc.source_id = 4;

-- List RHC Connections for a Tenant
SELECT rc.*
FROM sources s
         INNER JOIN source_rhc_connections srhc ON s.id = srhc.source_id
         INNER JOIN rhc_connections rc ON rc.id = srhc.rhc_connection_id
WHERE s.tenant_id = 1;
```

...translated to GORM:
```go
res := make(map[string]interface{})
result := dao.DB.Table("rhc_connections").
  Joins("INNER JOIN source_rhc_connections ON rhc_connections.id = source_rhc_connections.rhc_connection_id").
  Where("source_rhc_connections.source_id = ?", 4).
  Scan(&res)

if result.Error != nil {
  fmt.Printf("result.Error: %v\n", result.Error)
}

// couldn't get tihs one working. it complains there is no FROM clause for rhc_connections
res2 := make([]map[string]interface{}, 0)
result = dao.DB.Table("sources").Select("rhc_connections.*").
  Joins("INNER JOIN source_rhc_connections ON rhc_connections.id = sources.id").
  Joins("INNER JOIN rhc_connections ON rhc_connections.id = source_rhc_connections.rhc_connection_id").
  Where("sources.tenant_id = ?", 1).
  Scan(&res2)

if result.Error != nil {
  fmt.Printf("result.Error: %v\n", result.Error)
}

// works fine with the raw query.
err := dao.DB.Raw(`SELECT rc.*
           FROM sources s
            INNER JOIN source_rhc_connections srhc ON s.id = srhc.source_id
            INNER JOIN rhc_connections rc ON rc.id = srhc.rhc_connection_id
           WHERE s.tenant_id = 1;`).Scan(&res2).Error
if err != nil {
  fmt.Printf("err: %v\n", err)
}
```

dao/rhc_connection_dao.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (s *RhcConnectionDaoImpl) Update(rhcConnection *m.RhcConnection) error {
+	err := DB.Debug().
+		Updates(rhcConnection).
+		Error
+	return err
+}
+
+func (s *RhcConnectionDaoImpl) Delete(id *int64) error {
```
lindgrenj6: For this one if we added a fkey would it cascade delete? If so that would be something I'd be interested in adding to the db.

source_handlers.go
```diff
@@ -331,3 +331,50 @@ func SourceCheckAvailability(c echo.Context) error {
...
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	limit, offset, err := getLimitAndOffset(c)
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
+	}
+
+	// Check if the given source exists.
+	sourceDao, _ := getSourceDao(c)
```
lindgrenj6: Not sure we want to ignore the error and then check the previous error again :P

lindgrenj6 (commented):
Looking good - coming along nicely. Just a few comments and the usage of `.FirstOrCreate` like we discussed in PM.

dao/rhc_connection_dao.go
lindgrenj6: Actually one more thing - I see you used the many2many thing - we may be able to preload the associations!

`DB.Model(&m.Source{}).Preload("RhcConnections").Where("tenant_id = ?", s.TenantID).Scan(...)` would work maybe.

lpichler:
@MikelAlejoBR oh this PR is really big - is this meant as final PR or are you going to pull out changes to separated PRs ? 
or is this because of speed up process ?

dao/rhc_connection_dao.go
lindgrenj6: Also another thought: if we added tenancy to the join table we wouldn't have to do the double join on the sources table. What do you think of that?

MikelAlejoBR:
> @MikelAlejoBR oh this PR is really big - is this meant as final PR or are you going to pull out changes to separated PRs ? or is this because of speed up process ?

Unfortunately it is big :disappointed: . Most of the added lines belong to tests, though. However I wasn't planning on separating this on multiple PRs, although if you think it'll be easier to review it I can do it once I remove the WIP tag.

Thanks!

dao/rhc_connection_dao.go
MikelAlejoBR: It would. Here's a POC which is easy to play with:

```sql
-- Rhc Connections table --
CREATE TABLE "rhc_connection" (
    "id" BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
    "rhc_id" TEXT NOT NULL UNIQUE,
    "name" TEXT
);

-- Sources table --
CREATE TABLE "source"(
    "id" BIGINT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
    "name" TEXT,
    "tenant_id" BIGINT
);

-- Join table --
CREATE TABLE "rhc_connection_sources" (
    "rhc_id" BIGINT NOT NULL,
    "source_id" BIGINT NOT NULL,
    PRIMARY KEY ("rhc_id", "source_id"),
    CONSTRAINT "fk_rhc_connection_id" FOREIGN KEY ("rhc_id") REFERENCES "rhc_connection"("id") ON DELETE CASCADE,
    CONSTRAINT "fk_source_id" FOREIGN KEY ("source_id") REFERENCES "source"("id") ON DELETE CASCADE
);

-- Insert some data into the three tables --
INSERT INTO
    "rhc_connection"("rhc_id", "name")
VALUES
    ('unique-string', 'my rhc connection');

INSERT INTO
    "source"("name", "tenant_id")
VALUES
    ('my super source', 5);

INSERT INTO
    "rhc_connection_sources" ("rhc_id", "source_id")
VALUES
    (1, 1);

```

model/rhc_connection_http.go
MikelAlejoBR: Changed :+1:

model/event_models.go
MikelAlejoBR: This too :+1:

source_handlers.go
MikelAlejoBR: Good catch! Solved.

dao/rhc_connection_dao.go
MikelAlejoBR: > Actually one more thing - I see you used the many2many thing - we may be able to preload the associations!

I was avoiding to use `.Preload` because GORM just throws as many `SELECT * FROM`s as `.Preload` calls you may have, to then match everything in memory[citation needed].

> Also another thought: if we added tenancy to the join table we wouldn't have to do the double join on the sources table. 

I foresee complications when syncing data among the tables. First, we would need to create a new row in the join table every time a new source is created. That would imply leaving the `rhc_connection_id` column as `NULL`. Then, every time we create a new connection, we would have to update that row to include the `rhc_connection_id`.

I guess that could work, but I'm debating on whether to use the following and call it a day:

```sql
SELECT "rhc_connections".*, STRING_AGG(CAST ("src"."id" AS text), ',')
FROM "rhc_connections"
INNER JOIN "source_rhc_connections" "jt" ON "rhc_connections"."id" = "jt"."rhc_connection_id"
INNER JOIN "sources" "src" ON "jt"."source_id" = "src.id" WHERE "src"."tenant_id" = ?
GROUP BY "rhc_connections"."id" LIMIT 100
```

Essentially this query brings all the connections and the aggregated source ids for a given tenant. This will make me have to pseudo manually scan the rows:

```go
rows, err := /* blablabla query */.Rows()
if err != nil { return err }

for rows.Next() {
    var rhcConnection m.RhcConnection
    var sourceIds map[string]interface{}

    err = DB.Scan(rows, &rhcConnection)
    if err != nil { return err }

    err = DB.Scan(rows, &sourceIds)
    if err != nil { return err }
    
    // for each source ID create a new source object and append it to the rhcConnection.Sources slice
}
```

It is somewhat manual but I kind of like it. In order to produce the above query the following Go code is needed:

```go
	query := DB.
		Debug().
		Model(&m.RhcConnection{}).
		Select(`"rhc_connections".*, STRING_AGG(CAST ("src"."id" AS TEXT), ',') AS "source_ids"`).
		Joins(`INNER JOIN "source_rhc_connections" "jt" ON "rhc_connections"."id" = "jt"."rhc_connection_id"`).
		Joins(`INNER JOIN "sources" "src" ON "jt"."source_id" = "src"."id"`).
		Where(`"src"."tenant_id" = ?`, s.TenantID).
		Group(`"rhc_connections"."id"`).
		Limit(limit).
		Offset(offset)
```

What do you think?

dao/rhc_connection_dao.go
lindgrenj6: https://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/add_foreign_key

Yep will translate the FK into this, going to be fun.

dao/rhc_connection_dao.go
lindgrenj6: >First, we would need to create a new row in the join table every time a new source is created. 

Yeah that is pretty standard, and once we add the foreign key it will be fine. We don't need to leave any records NULL, at least I don't see where/why. For resource creation it's always going to be:
1. `DB.FirstOrCreate(&m.RhcConnection{RhcID: "asdfasdfaf")`
2. `DB.Create(&m.SourceRhcConnection{SourceID: <srcid from payload>, RhcConnectionID: <rhcid>})`

Then for listing we'll have to get fancy with the SQL - but if we add tenancy to the join talbe we won't have to double-join things in anymore at least. It'll be something like:
```sql
SELECT DISTINCT rhc_connections.*
FROM source_rhc_connections 
INNER JOIN rhc_connections ON rhc_connections.id = source_rhc_connections.rhc_connection_id
WHERE tenant_id = 1
```
This should be pretty easy to emulate with GORM at least.

MikelAlejoBR: > Yeah that is pretty standard, and once we add the foreign key it will be fine. We don't need to leave any records NULL, at least I don't see where/why. For resource creation it's always going to be:

You are right: when creating the connection we already will be able to create the row in the join table, since all the required ids will come in the request —source id, tenant id and the connection id that we will grab from the freshly created one—.

All right, suggestion implemented.

dao/rhc_connection_dao.go
MikelAlejoBR: Implemented in the DAOs.

MikelAlejoBR:
This is ready for review. I'm sorry about the number of additions, but a big amount of them are related to tests. If you want me to split this PR or do something which can help you review it please do let me know. Thank you.

dao/rhc_connection_dao.go
```diff
@@ -0,0 +1,236 @@
...
+			FirstOrCreate(&rhcConnection).
+			Error
+
+		// Is it a new connection or is it just an association?
+		if rhcConnection.ID == 0 {
+			err = tx.Debug().
+				Omit(clause.Associations).
+				Create(&rhcConnection).
+				Error
+		}
```
lindgrenj6: This shouldn't be needed - the ID will get populated from the `.FirstOrCreate()` call.
`.FirstOrCreate` basically does 2 queries, a SELECT and an INSERT. The insert runs iff the record is not found, see here: 

```
{"@timestamp":"2022-02-15T09:52:05.398Z"..."message":"SELECT * FROM \"rhc_connections\" WHERE rhc_id = 'thing2' ORDER BY \"rhc_connections\".\"id\" LIMIT 1","rows":0,"tags":["source-api-go"]}
{"@timestamp":"2022-02-15T09:52:05.408Z"..."message":"INSERT INTO \"rhc_connections\" (\"rhc_id\",\"extra\",\"availability_status\",\"last_checked_at\",\"last_available_at\",\"availability_status_error\",\"created_at\",\"updated_at\") VALUES ('thing2',NULL,'','0000-00-00 00:00:00','0000-00-00 00:00:00','','2022-02-15 09:52:05.398','2022-02-15 09:52:05.398') RETURNING \"id\"","rows":1,"tags":["source-api-go"]}
```

Those were ran when I did this in my little dummy handler: 
```go
e.GET("/", func(c echo.Context) error {
	thing := m.RhcConnection{RhcId: "thing2"}
	result := dao.DB.Model(&thing).
		Where("rhc_id = ?", thing.RhcId).
		FirstOrCreate(&thing)
	if result.Error != nil {
		fmt.Println(result)
	}
	return c.JSON(http.StatusOK, thing)
})
```

From here on, when I hit that endpoint there is only one query ran: 
```
{"@timestamp":"2022-02-15T09:52:07.794Z"...,"message":"SELECT * FROM \"rhc_connections\" WHERE rhc_id = 'thing2' ORDER BY \"rhc_connections\".\"id\" LIMIT 1","rows":1,"tags":["source-api-go"]}
```

dao/rhc_connection_dao.go
```diff
@@ -0,0 +1,236 @@
...
+
+		if err != nil {
+			return err
+		}
+
+		// Try to insert an association. If it exists the database will complain.
+		association := m.SourceRhcConnection{
```
lindgrenj6: Small nit - can we rename this var to `sourceRhcConnection` or something - the association key word threw me off on first look.

dao/source_dao.go
```diff
@@ -182,3 +182,30 @@ func (s *SourceDaoImpl) ToEventJSON(resource util.Resource) ([]byte, error) {
 
 	return data, err
 }
+
+func (s *SourceDaoImpl) GetRelatedRhcConnectionsToId(sourceId *int64, limit, offset int, filters []util.Filter) ([]m.RhcConnection, int64, error) {
```
lindgrenj6: I think this method belongs on the RHC Connection DAO since we're fetching RHC Connection resources. 

I'm trying to start a pattern - so I would recommend moving this over the RHC Connection DAO and name the method `ListForSource`, the same way I named the `AuthenticationDAO` subresource methods.

model/rhc_connection.go
```diff
@@ -0,0 +1,64 @@
...
+	"strconv"
+	"time"
+
+	"github.com/RedHatInsights/sources-api-go/util"
+	"gorm.io/datatypes"
+)
+
+type RhcConnection struct {
+	ID    int64 `gorm:"primaryKey" json:"id"`
+	RhcId string
```
lindgrenj6: I'd add some `json` tags here just in case we need to marshal it someday.

model/rhc_connection_http.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// RhcConnectionUpdateRequest represents a request coming from the outside to update a Red Hat Connector connection.
+type RhcConnectionUpdateRequest struct {
+	Extra datatypes.JSON `json:"extra"`
+}
+
+type RhcConnectionResponse struct {
+	Id    *string        `json:"id"`
+	Uuid  *string        `json:"rhc_id"`
```
lindgrenj6: Just need to make these consistent, always using `RhcID` and marhsalling as `rhc_id`

rhc_connection_handlers.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return c.JSON(http.StatusBadRequest, util.ErrorDoc("invalid source id", "400"))
+	}
+
+	rhcConnection := &model.RhcConnection{
+		RhcId: input.RhcId,
+		Extra: input.Extra,
+		AvailabilityStatus: model.AvailabilityStatus{
+			AvailabilityStatus: input.AvailabilityStatus,
+		},
```
lindgrenj6: Probably don't want to allow availability status on create - that is going to be an update-only field.

model/event_models.go
lindgrenj6: We're doing to return the field as `rhc_id` instead of `uuid`.

lindgrenj6 (commented):
Just some changes for consistency - plus moving around some stuff. Other than that I think we're mostly good.

dao/rhc_connection_dao.go
MikelAlejoBR: Sure, done!

model/rhc_connection.go
MikelAlejoBR: Done.

model/event_models.go
MikelAlejoBR: Changed. I also included the `ID` field that was missing.

model/rhc_connection_http.go
MikelAlejoBR: Changed!

dao/source_dao.go
MikelAlejoBR: All right, moved and renamed!

rhc_connection_handlers.go
MikelAlejoBR: All right, I've removed that field from the creation. Regarding being an update-only field: you mean that we will update it on the back, right? We won't allow clients to update that field.

dao/rhc_connection_dao.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+func (s *RhcConnectionDaoImpl) Create(rhcConnection *m.RhcConnection) (*m.RhcConnection, error) {
+	// If the source doesn't exist we cannot create the RhcConnection, since it needs to be linked to at least one
+	// source.
+	var sourceExists bool
+	err := DB.Debug().
+		Model(&m.Source{}).
+		Select(`1`).
+		Where(`id = ?`, rhcConnection.Sources[0].ID).
```
lindgrenj6, line 128: Sorry I missed this yesterday - we need a tenancy check otherwise someone could attach to any source!

dao/rhc_connection_dao.go
```diff
@@ -0,0 +1,255 @@
...
+	rhcConnection := &m.RhcConnection{ID: *id}
+
+	err := DB.Debug().
+		Where("id = ?", id).
+		First(&rhcConnection).
+		Error
+
+	if err != nil {
+		return util.NewErrNotFound("rhcConnection")
+	}
```
lindgrenj6: Hmmm...this got em thinking of a refactor we can do eventually. Rather than fetching the entirety of the record and checking I think we can use (read as: abuse) the postgres RETURNING keyword to do it all in one. 

EDIT: yep found the docs! It's ugly but easy enough to understand: https://gorm.io/docs/delete.html#Returning-Data-From-Deleted-Rows

We can do that in a refactor though, that way we can change it everywhere. It will be nice to change out all of the `Delete` functions across the app to use it.

dao/rhc_connection_dao.go
```diff
@@ -0,0 +1,255 @@
...
+	// The foreign key in the join table takes care of deleting the associated row.
+	err = DB.Debug().
+		Where(`id = ?`, *id).
+		Delete(&m.RhcConnection{}).
+		Error
+
+	return err
+}
+
+func (s *RhcConnectionDaoImpl) GetRelatedSourcesToId(rhcConnectionId *int64, limit, offset int, filters []util.Filter) ([]m.Source, int64, error) {
```
lindgrenj6: Once again I'm mad I missed this yesterday, I caught one but not the other...This should probably be on the `SourceDAO` since it is returning an array of sources, something like `SourceDAO#ListForRhcConnection`.

model/rhc_connection.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (r *RhcConnection) ToResponse() *RhcConnectionResponse {
+	id := strconv.FormatInt(r.ID, 10)
+
+	var sourceIds []string
+	if len(r.Sources) > 0 {
+		for _, src := range r.Sources {
+			sourceIds = append(sourceIds, strconv.FormatInt(src.ID, 10))
+		}
+	}
```
lindgrenj6: For this I'd do something like:

```suggestion
	sourceIds := make([]string, len(r.Sources))
	for i, src := range r.Sources {
		sourceIds[i] = strconv.FormatInt(src.ID, 10)
	}
```

1. Allocating exactly the size you need in the array!
2. Looping over without checking bounds (the loop wouldn't run unless there is a source, and there is almost guaranteed there will be one)

model/rhc_connection_http.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"gorm.io/datatypes"
+)
+
+// RhcConnectionCreateRequest represents a request coming from the outside to create a Red Hat Connector connection.
+type RhcConnectionCreateRequest struct {
+	RhcId    string         `json:"rhc_id"`
+	Extra    datatypes.JSON `json:"extra"`
+	SourceId string         `json:"source_id"`
```
lindgrenj6: Might need to make this one `interface{}` that way we can accept ints and strings for this field. Kind of like the `SourceIDRaw` field on Endpoint/Application

rhc_connection_handlers.go
```diff
@@ -0,0 +1,220 @@
...
+	if err != nil {
+		return err
+	}
+
+	connection, err := rhcConnectionDao.Create(rhcConnection)
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(http.StatusCreated, connection.ToResponse())
```
lindgrenj6, line 113: We most likely need a call to `setEventStreamResource(c, connection)` here, make sure it implements the `.ToEvent()` method to satisfy the `model.Event` interface. 

This would be done on the create/update/delete endpoints. It looks like the middleware is already included at least!

lindgrenj6 (commented):
Just a few more things - a few I should have seen yesterday but this is always how code review goes right? :stuck_out_tongue:

dao/rhc_connection_dao.go
MikelAlejoBR: Great catch! Added a test as well to avoid regressions.

dao/rhc_connection_dao.go
MikelAlejoBR: I've refactored the `.Delete()` function so it returns the deleted object, because I had missed it.

---

I also was thinking on creating an issue to implement the `.Exists() (bool, err)` function in the DAOs, which only would do a `SELECT 1` instead of a  `SELECT *`, since there are handlers which only need to know whether the related entity exists or not.

Anyway, I created a story for the refactor so we don't forget: https://issues.redhat.com/browse/RHCLOUD-18138

dao/rhc_connection_dao.go
MikelAlejoBR: Moved!

model/rhc_connection.go
MikelAlejoBR: Oooof yes. I confused the `len` and `cap` arguments in the `make` call, and I was being hit by some weird "index out of bounds" errors. I was doing the following:

`make([]string, 0, len(r.Sources)`

Thus allocating a zero length array. After that, by trying to access it with the `sourceIds[i]` statement I was having that error. That's why I was using a more "manual" loop.

I have refactored it now.

model/rhc_connection_http.go
MikelAlejoBR: Good catch! Refactored.

rhc_connection_handlers.go
MikelAlejoBR: Done!

service/rhc_connection_validation.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/RedHatInsights/sources-api-go/util"
+)
+
+// ValidateRhcConnectionRequest validates that the incoming input is valid.
+func ValidateRhcConnectionRequest(req *model.RhcConnectionCreateRequest) error {
+	if req.RhcId == "" {
+		return errors.New("the Red Hat Connector Connection's id is invalid")
+	}
+
+	sourceId, err := util.InterfaceToInt64(req.SourceId)
```
lindgrenj6: I think this needs to be `util.InterfaceToInt64(req.SourceIdRaw)`

lindgrenj6 (approved):
One last change, then we're good.

service/rhc_connection_validation.go
MikelAlejoBR: Nice catch. Refactored.

lpichler (approved):

PR #115, opened 2022-02-14
[RHCLOUD-18031] BadRequest error handling in source handler

petracihalova:
[RHCLOUD-18031](https://issues.redhat.com/browse/RHCLOUD-18031)

PR #116, opened 2022-02-14
[RHCLOUD-18032] BadRequest error handling in application type handler

petracihalova:
[RHCLOUD-18032](https://issues.redhat.com/browse/RHCLOUD-18032)

### DIFF
--- a/dao/application_type_dao.go
+++ b/dao/application_type_dao.go
@@ -31,7 +31,7 @@ func (a *ApplicationTypeDaoImpl) SubCollectionList(primaryCollection interface{}
 	// limiting + running the actual query.
 	result := query.Limit(limit).Offset(offset).Find(&applicationTypes)
 	if result.Error != nil {
-		return nil, 0, util.NewErrBadRequest(result.Error.Error())
+		return nil, 0, util.NewErrBadRequest(result.Error)
 	}
 
 	return applicationTypes, count, nil
@@ -45,7 +45,7 @@ func (a *ApplicationTypeDaoImpl) List(limit, offset int, filters []util.Filter) 
 
 	query, err := applyFilters(query, filters)
 	if err != nil {
-		return nil, 0, util.NewErrBadRequest(err.Error())
+		return nil, 0, util.NewErrBadRequest(err)
 	}
 
 	// getting the total count (filters included) for pagination
@@ -55,7 +55,7 @@ func (a *ApplicationTypeDaoImpl) List(limit, offset int, filters []util.Filter) 
 	// limiting + running the actual query.
 	result := query.Limit(limit).Offset(offset).Find(&appTypes)
 	if result.Error != nil {
-		return nil, 0, util.NewErrBadRequest(result.Error.Error())
+		return nil, 0, util.NewErrBadRequest(result.Error)
 	}
 
 	return appTypes, count, nil
--- a/dao/application_type_dao.go
+++ b/dao/application_type_dao.go
@@ -30,8 +30,11 @@ func (a *ApplicationTypeDaoImpl) SubCollectionList(primaryCollection interface{}
 
 	// limiting + running the actual query.
 	result := query.Limit(limit).Offset(offset).Find(&applicationTypes)
+	if result.Error != nil {
+		return nil, 0, util.NewErrBadRequest(result.Error.Error())
+	}
 
-	return applicationTypes, count, result.Error
+	return applicationTypes, count, nil
 }
 
 func (a *ApplicationTypeDaoImpl) List(limit, offset int, filters []util.Filter) ([]m.ApplicationType, int64, error) {
@@ -42,7 +45,7 @@ func (a *ApplicationTypeDaoImpl) List(limit, offset int, filters []util.Filter) 
 
 	query, err := applyFilters(query, filters)
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, util.NewErrBadRequest(err.Error())
 	}
 
 	// getting the total count (filters included) for pagination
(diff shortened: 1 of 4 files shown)


dao/application_type_dao.go
```diff
@@ -30,8 +30,11 @@ func (a *ApplicationTypeDaoImpl) SubCollectionList(primaryCollection interface{}
 
 	// limiting + running the actual query.
 	result := query.Limit(limit).Offset(offset).Find(&applicationTypes)
+	if result.Error != nil {
+		return nil, 0, util.NewErrBadRequest(result.Error.Error())
+	}
```
MikelAlejoBR: I would suggest the following:

```diff
- result := query.Limit(limit).Offset(offset).Find(&applicationTypes)
+ err := query.Limit(limit).Offset(offset).Find(&applicationTypes).Error
```

This way you don't have to chain the `result.Error.Error()` members and functions.

dao/application_type_dao.go
```diff
@@ -51,8 +54,11 @@ func (a *ApplicationTypeDaoImpl) List(limit, offset int, filters []util.Filter)
 
 	// limiting + running the actual query.
 	result := query.Limit(limit).Offset(offset).Find(&appTypes)
+	if result.Error != nil {
+		return nil, 0, util.NewErrBadRequest(result.Error.Error())
+	}
```
MikelAlejoBR: Same thing here. Actually, wouldn't this be an internal server error? If I recall correctly, the filters are validated at the handler level, where a `400` is returned if something unexpected is sent by the client. But once those filters reach the DAO level, they're supposed to be valid, so if an error occurs it should be a 5xx, shouldn't it?

MikelAlejoBR (approved):
Looks good to me! Just a minor comment.

dao/application_type_dao.go
petracihalova: Well, maybe it's for more discussion ... because I did the same here I did in the previous PR for application handlers [(here)](https://github.com/RedHatInsights/sources-api-go/pull/110)

and now I noticed that the line `query, err = applyFilters(query, filters)` is missing in application_type_dao.go

so all these requests return same data:
```
/api/sources/v3.1/sources/2/application_types
/api/sources/v3.1/sources/2/application_types?filter[asdfasdf]=asdfasfasf
/api/sources/v3.1/sources/2/application_types?filter[id]=10
```
and after your comment I realised after adding `query, err = applyFilters(query, filters)` in method SubCollectionList() I am able to simulate several different errors and I think some can be 4xx and some 5xx

1) when I rename table in DB = something wrong with DB ... this is I think 5xx error
```
{
    "errors": [
        {
            "detail": "ERROR: relation \"application_types\" does not exist (SQLSTATE 42P01)",
            "status": "400"
        }
    ]
}
```
2) when I use wrong filter` api/sources/v3.1/sources/2/application_types?filter[id]=aaa` ... this is I think 4xx error
```
{
    "errors": [
        {
            "detail": "ERROR: invalid input syntax for type bigint: \"aaa\" (SQLSTATE 22P02)",
            "status": "400"
        }
    ]
}
```

3) or` api/sources/v3.1/sources/{{id}}/application_types?filter[bbb]=aaa`
```
{
    "errors": [
        {
            "detail": "ERROR: column application_types.bbb does not exist (SQLSTATE 42703)",
            "status": "400"
        }
    ]
}
```

so I dont know how to solve this ... because we can get all of the above on line 56
@lindgrenj6?

lindgrenj6: This one is interesting, the first example (a table rename) _would_ in fact be a 500, but would most likely be caught in our integration tests luckily. So I wouldn't worry about catching that sort of error. I think in the near future when we flesh out the `/health` endpoint a bit more to hit the database it'll be really nice to have that response be something telling if the db can be hit, maybe even names of tables, etc. 

The other two though are indeed 400s. I would just leave them as 400s for now. We aren't validating the filters (just parsing them) in the middleware.

petracihalova: so I can leave it as it is now? maybe I can do the change suggested by Mikel in the first comment
`+ err := query.Limit(limit).Offset(offset).Find(&applicationTypes).Error`

lindgrenj6: That's up to you - I see the point but it's all the same to me.

lindgrenj6 (commented):
I actually have a suggestion - can we make the helper functions take an interface so we can pass both strings and errors? that way we don't have to call `err.Error()` everywhere.

Like so:
```diff
diff --git a/util/error_document.go b/util/error_document.go
index 6fa906d..33a5c86 100644
--- a/util/error_document.go
+++ b/util/error_document.go
@@ -57,6 +57,13 @@ func (e ErrBadRequest) Is(err error) bool {
 	return reflect.TypeOf(err) == reflect.TypeOf(e)
 }
 
-func NewErrBadRequest(t string) error {
-	return ErrBadRequest{Message: t}
+func NewErrBadRequest(t interface{}) error {
+	switch t := t.(type) {
+	case string:
+		return ErrBadRequest{Message: t}
+	case error:
+		return ErrBadRequest{Message: t.Error()}
+	default:
+		panic("bad interface type for bad request: " + reflect.ValueOf(t).String())
+	}
 }
```

petracihalova:
NewErrBadRequest() updated ... after confirmation and merge I prepare needed changes for source handler (https://github.com/RedHatInsights/sources-api-go/pull/115) and for application handler (https://github.com/RedHatInsights/sources-api-go/pull/110)

MikelAlejoBR (approved):

lindgrenj6 (approved):

PR #117, opened 2022-02-16
Add ApplicationAuthentications to BulkMessageFromSource

lpichler:
- pull out from https://github.com/RedHatInsights/sources-api-go/pull/76(but code is mostly changed) as last thing of  https://issues.redhat.com/browse/RHCLOUD-17059

This adds `ApplicationAuthentications` relations to bulk message which is build here in [UpdateMessage](https://github.com/RedHatInsights/sources-api-go/blob/cbd8755cc54d9701e8e363a40f939bed91a25c8c/internal/events/event_stream_producer.go) method 

This bulk message is send to Kafka topic `platform.sources.event-stream`  by [raise event](https://github.com/RedHatInsights/sources-api-go/blob/cbd8755cc54d9701e8e363a40f939bed91a25c8c/internal/events/event_stream_producer.go) function.

Bulk message is built [here](https://github.com/RedHatInsights/sources-api/blob/19cf1aa2f6f2ba7f0ce8c17dc8a4662cfabc368d/app/models/concerns/event_concern.rb) in ruby project.

### DIFF
--- a/dao/application_authentication_dao.go
+++ b/dao/application_authentication_dao.go
@@ -19,9 +19,9 @@ func (a *ApplicationAuthenticationDaoImpl) ApplicationAuthenticationsByApplicati
 		applicationIDs = append(applicationIDs, value.ID)
 	}
 
-	result := DB.Preload("Tenant").Where("application_id IN ?", applicationIDs).Find(&applicationAuthentications)
-	if result.Error != nil {
-		return nil, result.Error
+	err := DB.Preload("Tenant").Where("application_id IN ?", applicationIDs).Find(&applicationAuthentications).Error
+	if err != nil {
+		return nil, err
 	}
 
 	return applicationAuthentications, nil
--- a/dao/application_authentication_dao.go
+++ b/dao/application_authentication_dao.go
@@ -11,6 +11,46 @@ type ApplicationAuthenticationDaoImpl struct {
 	TenantID *int64
 }
 
+func (a *ApplicationAuthenticationDaoImpl) ApplicationAuthenticationsByApplications(applications []m.Application) ([]m.ApplicationAuthentication, error) {
+	var applicationAuthentications []m.ApplicationAuthentication
+
+	applicationIDs := make([]int64, 0)
+	for _, value := range applications {
+		applicationIDs = append(applicationIDs, value.ID)
+	}
+
+	result := DB.Preload("Tenant").Where("application_id IN ?", applicationIDs).Find(&applicationAuthentications)
+	if result.Error != nil {
+		return nil, result.Error
+	}
+
+	return applicationAuthentications, nil
+}
+
+func (a *ApplicationAuthenticationDaoImpl) ApplicationAuthenticationsByAuthentications(authentications []m.Authentication) ([]m.ApplicationAuthentication, error) {
+	var applicationAuthentications []m.ApplicationAuthentication
+
+	authenticationUIDs := make([]string, 0)
+	for _, value := range authentications {
+		authenticationUIDs = append(authenticationUIDs, value.ID)
+	}
+
+	result := DB.Preload("Tenant").Where("authentication_uid IN ?", authenticationUIDs).Find(&applicationAuthentications)
+	if result.Error != nil {
+		return nil, result.Error
+	}
+
+	return applicationAuthentications, nil
+}
+
+func (a *ApplicationAuthenticationDaoImpl) ApplicationAuthenticationsByResource(resourceType string, applications []m.Application, authentications []m.Authentication) ([]m.ApplicationAuthentication, error) {
+	if resourceType == "Source" {
+		return a.ApplicationAuthenticationsByApplications(applications)
+	}
+
+	return a.ApplicationAuthenticationsByAuthentications(authentications)
+}
+
 func (a *ApplicationAuthenticationDaoImpl) List(limit int, offset int, filters []util.Filter) ([]m.ApplicationAuthentication, int64, error) {
 	appAuths := make([]m.ApplicationAuthentication, 0, limit)
 	query := DB.Debug().Model(&m.ApplicationAuthentication{}).
--- a/dao/common.go
+++ b/dao/common.go
@@ -65,8 +65,6 @@ func BulkMessageFromSource(source *m.Source, authentication *m.Authentication) (
 
 	bulkMessage["applications"] = applications
 
-	bulkMessage["application_authentications"] = []m.ApplicationAuthenticationEvent{}
-
 	authDao := &AuthenticationDaoImpl{TenantID: &source.TenantID}
 	authenticationsByResource, err := authDao.AuthenticationsByResource(authentication)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


dao/application_authentication_dao.go
```diff
@@ -11,6 +11,46 @@ type ApplicationAuthenticationDaoImpl struct {
 	TenantID *int64
 }
 
+func (a *ApplicationAuthenticationDaoImpl) ApplicationAuthenticationsByApplications(applications []m.Application) ([]m.ApplicationAuthentication, error) {
```
MikelAlejoBR, line 14: What do you think about renaming this to `ByApplications`? Sine we're in the `ApplicationAuthenticationDaoImpl` I guess we can assume that the query is to fetch appauths. Same about the other two functions, which I guess they could be renamed to `ByAuthentications` and `ByResource` accordingly.

dao/application_authentication_dao.go
```diff
@@ -11,6 +11,46 @@ type ApplicationAuthenticationDaoImpl struct {
...
+
+	applicationIDs := make([]int64, 0)
+	for _, value := range applications {
+		applicationIDs = append(applicationIDs, value.ID)
+	}
+
+	result := DB.Preload("Tenant").Where("application_id IN ?", applicationIDs).Find(&applicationAuthentications)
+	if result.Error != nil {
+		return nil, result.Error
+	}
```
MikelAlejoBR: You can append `.Error` after the last call so that it directly returns the error. That way you can simply do a `if err != nil { return nil, err`.

statuslistener/statuslistener_test.go
```diff
@@ -248,6 +275,29 @@ func TransformDateFieldsInJSONForBulkMessage(resourceType string, resourceID str
 
 	contentMap["endpoints"] = endpoints
 
+	var applicationAuthentications []interface{}
+
+	applicationAuthenticationsBulkMessage, success := bulkMessage["application_authentications"].([]m.ApplicationAuthentication)
+
+	if success {
```
MikelAlejoBR: I would suggest using what's called "the comma ok" notation, which is a convention for these cases: https://go.dev/doc/effective_go#maps . The same thing applies for the type checkings below.

dao/common.go
```diff
@@ -79,6 +77,20 @@ func BulkMessageFromSource(source *m.Source, authentication *m.Authentication) (
...
 	}
 
+	applicationAuthenticationDao := &ApplicationAuthenticationDaoImpl{TenantID: &source.TenantID}
+	applicationAuthenticationsFromResource, err := applicationAuthenticationDao.ApplicationAuthenticationsByResource(authentication.ResourceType, source.Applications, authenticationsByResource)
+
+	if err != nil {
+		return nil, err
+	}
+
+	applicationAuthentications := make([]interface{}, len(applicationAuthenticationsFromResource))
```
MikelAlejoBR, line 87: What about using `[]*m.ApplicationAuthenticationEvent` here instead of the `[]interface{}`? That way its easier to know at a glance what the slice holds.

MikelAlejoBR (commented):
Just a few minor comments. Other than that, it looks good to me!

dao/common.go
lindgrenj6: `.ToEvent()` needs to return `interface{}` to satisfy the `model.Event` interface. Since we're just marshaling and throwing it onto kafka the type is less important.

MikelAlejoBR: All right, thank you. In this case it doesn't satisfy that `model.Event` interface, though, so I would probably vote for refactoring that in a follow-up PR.

dao/application_authentication_dao.go
lindgrenj6: The word `map`/`get` comes to mind too. `GetFromApplications` and/or `MapFromApplications`.

lpichler: naming is hard and in this case it looks that origin suggestion seems to me still more clear (I am ok with changing `By` -> `From` as it is probably correct from English language )

so we have signature:

```go
func (a *ApplicationAuthenticationDaoImpl) ApplicationAuthenticationsByApplications(applications []m.Application)
```

**Removing `ApplicationAuthentications` because type of receiver is "same"**
but receiver is not same type it just contains `ApplicationAuthentication` and even it is in singular form - so it is different type and there is also confusion about plurals. 

**Include Get or Map in name of method**

**Get**
I don't prefer prefix `Get` because this word in here doesn't say anything in addition there is no added value (example):  

```
aa := a.ApplicationAuthentications(..)
```
vs

```
aa := a.GetApplicationAuthentications(..)
```

it is obvious that in both cases returned values are` ApplicationAuthentications`

**Map**
 
Map is good word in method naming but here 
It feels that `Map` doesn't cover exactly the meaning - maps are types in GO (end even there are not used it this method ) and 
 in this method is mapping is not happening there is just query


**By -> From *
I am ok with this.

NOTE: personally I would remove also `...Applications` because it is stated in first parameter but I agreed with Jacob that it should stay here because it is more obvious what is method about.

Naming is hard and important - so thanks for your inputs - it is important how other perceives it - we definitely need all to agree on this - so we can use this pattern across whole application.

let me know what do you think about it now after my comments, thanks!

MikelAlejoBR: > **Removing `ApplicationAuthentications` because type of receiver is "same"**
> but receiver is not same type it just contains `ApplicationAuthentication` and even it is in singular form - so it is different type and there is also confusion about plurals.

I was referring to the fact that the function's signature already gives a clue about what we will be returning: `([]m.ApplicationAuthentication, error)`. So by taking a look at the signature we know that an array of "ApplicationAuthentication"s will be returned. And if we comment the function then it'd be even more clear.


It is true though that the struct could be named anything, like `type DifferentNameDaoImpl struct` and make it implement the "ApplicationAuthenticationDao" interface, which would make using the `List()`, `Create()`, `GetBy` functions unclear: what would we be listing? DifferentNames?

However this is an hypothetical case, which I guess we wouldn't allow in a code review.

---

Regarding standardizing a convention for naming methods, I do really like how Spring names the operations in their CRUD repositories\[1\]:

```java
    <S extends T> S save(S entity);
    T findOne(ID primaryKey);
    Iterable<T> findAll();
    Long count();
    void delete(T entity);
    boolean exists(ID primaryKey);
```

Plus the signatures for functions like the one we're discussing about would be like `FindAllByApplications(...)`, which maks it clear that it is going to find all the AppAuths by the applications you provide.

Though I'm feeling that this PR shouldn't be the place to discuss this.

\[1\]: https://docs.spring.io/spring-data/data-commons/docs/1.6.1.RELEASE/reference/html/repositories.html

---

So, my final vote is: let's leave it like it is, since discussing the naming of the functions isn't the goal of this PR. I'm sorry for diverting the discussion.

lindgrenj6:
Yeah IMO we can make this a ticket in the backlog or something - if we want
to make everything standardized (they mostly are, but there are a few
outliers like we're seeing) we can do a refactor there in the future.
--
Jacob Lindgren

Red Hat

dao/application_authentication_dao.go
lpichler: thanks @MikelAlejoBR for your insights , I understand now and I like `FindAll` pattern but let's leave it for follow up PRs.

lpichler:
thanks for review, I added changes according to @MikelAlejoBR's comments. (except about naming methods )

MikelAlejoBR (approved):

lindgrenj6 (approved):

PR #118, opened 2022-02-16
refactor: pull "RaiseEvent" function from the middleware closure

MikelAlejoBR:
The reason for pulling it from the closure is that we might want to use the "RaiseEvent" function in a place where multiple events must be raised, and therefore if it is embedded in the closure that would be difficult to do.

### DIFF
--- a/middleware/event_stream.go
+++ b/middleware/event_stream.go
@@ -1,20 +1,14 @@
 package middleware
 
 import (
-	"encoding/json"
-
-	"github.com/RedHatInsights/sources-api-go/internal/events"
-	"github.com/RedHatInsights/sources-api-go/kafka"
 	l "github.com/RedHatInsights/sources-api-go/logger"
-	"github.com/RedHatInsights/sources-api-go/util"
+	"github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/service"
 	"github.com/labstack/echo/v4"
 )
 
-// producer instance used to send messages - default just an empty instance of
-// the struct.
-var producer = events.EventStreamProducer{Sender: &events.EventStreamSender{}}
-
-// RaiseEventMiddleware calls the "RaiseEvent" function once the previous handler has succeeded.
+// RaiseEventMiddleware calls the "RaiseEvent" function once the previous handler has succeeded. It grabs the resource
+// and the event type from the context.
 func RaiseEventMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
 	return func(c echo.Context) error {
 		// first call the handler function (or the next middlware)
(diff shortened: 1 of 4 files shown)


lindgrenj6 (commented):
Instead of just yoinking the event function and leaving it in the `middleware` package what do you think about moving it to `service` and make it take a couple parameters.

Something like `service.RaiseEvent(eventType string, thing model.Event)`, then the middleware would just call that function the params parsed from the context (any overrides, the message, the event itself). This way we would have separation of context-aware sending and one-off sending. 

Unless you were thinking of just setting the overrides in the handlers? but that wouldn't work for the source pause event I guess where you'd need to send N application events and 1 source event.

MikelAlejoBR:
I was thinking on doing `c.Set` for the source, then raising the event, and then `c.Set` on each loop when iterating the list of applications. But I like your idea more: having a function on the service package and then making the middleware call that with the context resource.

MikelAlejoBR:
I've moved the `RaiseEvent` function to the service package, but it still needs the context, because even if I removed the "skip" and "override" code bits, we must forward the identity headers anyway :/

middleware/event_stream.go
```diff
@@ -51,70 +31,8 @@ func RaiseEvent(next echo.HandlerFunc) echo.HandlerFunc {
...
-		// request isn't effectively blocked by kafka .
-		headers := append(getRequestHeaders(c), kafka.Header{Key: "event_type", Value: []byte(eventType)})
-		err = producer.RaiseEvent(eventType, msg, headers)
-		if err != nil {
-			l.Log.Warnf("failed to raise event to kafka: %v", err)
-			return nil
-		}
-
-		return nil
+		return service.RaiseEvent(c, eventType, resource)
```
lindgrenj6: Maybe I should have thought about it more. I know we're going to need to pass the context for the headers or just have another helper method that takes the context as an argument.

I was more implying we should leave the `event_override` and other stuff based on the context here in the middleware. Basically all that `service.RaiseEvent()` would do is marshal the resource + post the message. 

---

Maybe we can change the `getRequestHeaders(c)` function to something like `ForwardableHeaders(c)` so it takes in the echo context and returns the slice of headres - that way the new RaiseEvent function would have a signature like:

```go
func RaiseEvent(eventType string, resource m.Resource, headers []kafka.Header) error

// called via
service.RaiseEvent("Source.pause", source, service.ForwardableHeaders(c))
```

The RaiseEvent function would just add the `event_type` header and produce the message.

---

Does that make sense?

MikelAlejoBR: Absolutely. I implemented those changes. Thanks for the feedback!

middleware/event_stream.go
```diff
@@ -1,46 +1,26 @@
...
 		}
 
-		// specifically skip raising an event if this is set - usually when
-		// a create action happened but we do not want to re-raise the
-		// event.
-		if c.Get("skip_raise") != nil {
-			l.Log.Infof("skipping raise event per skip_raise set on context")
-			return nil
-		}
-
```
lindgrenj6: So this part needs to stay - I am surprised my tests didn't catch it.

middleware/event_stream.go
```diff
@@ -51,70 +31,10 @@ func RaiseEvent(next echo.HandlerFunc) echo.HandlerFunc {
...
-			event, ok := c.Get("event_override").(string)
-			if !ok {
-				l.Log.Warnf("Failed to cast event_override from request - ditching post to kafka")
-				return nil
-			}
-
-			l.Log.Infof("Using overridden event_type %v instead of %v", c.Get("event_override"), eventType)
-			eventType = event
-		}
-
```
lindgrenj6: This should stay too. I don't have a test around this, oops!

lindgrenj6 (commented):
Just 2 little pieces that need to be added back into the middleware - other than that looks good to me!

middleware/event_stream.go
MikelAlejoBR: I think it is because of the following:

```go
	f := raiseMiddleware(func(c echo.Context) error {
		c.Set("skip_raise", struct{}{})
		return c.NoContent(http.StatusNoContent)
	})
```

It seems to be returning the `StatusNoContent` all the time, which causes the tests to pass no matter what.

middleware/event_stream.go
```diff
@@ -1,27 +1,15 @@
...
-	Function that takes an event-type as a string and then returns a middleware
-	function that raises the specified event if the handler operation succeeds.
-
-	It bails out if there isn't a `resource` field on the context which should
-	be a model.ToEvent() call in the handler.
-*/
-func RaiseEvent(next echo.HandlerFunc) echo.HandlerFunc {
+// RaiseEventMiddleware calls the "RaiseEvent" function once the previous handler has succeeded. It grabs the resource
+// and the event type from the context.
+func RaiseEventMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
```
lindgrenj6: One last teeny thing - this can now just be `RaiseEvent` again.

MikelAlejoBR: Renamed!

lindgrenj6 (approved):

PR #124, opened 2022-02-21
Add RHC Connection Availability Check

lindgrenj6:
This is going to sit for a while until #111 is merged, the org_id <-> acct # service is deployed, and getting time to finish it.

Pretty simple process though - loop through all of the join records and request availability from RHC for each of them. It gets returned from the POST request right in the HTTP response, nice!

### DIFF
--- a/dao/rhc_connection_dao_test.go
+++ b/dao/rhc_connection_dao_test.go
@@ -1,507 +1,507 @@
 package dao
 
-import (
-	"bytes"
-	"errors"
-	"reflect"
-	"strings"
-	"testing"
+// import (
+// 	"bytes"
+// 	"errors"
+// 	"reflect"
+// 	"strings"
+// 	"testing"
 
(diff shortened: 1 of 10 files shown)


lindgrenj6:
Well - it looks like a bunch of RHC Connection DAO tests are failing now that I added the relationships on the struct(s). It may be due to letting gorm generate the relations I'm wondering.

lindgrenj6:
So it turns out we need this functionality yesterday - I commented out all of the failing tests just because they're valid, but adding the GORM relations on the structs just broke most of these. GORM is creating foreign keys etc that make the tests fail. 

For now I would just like to get the RHC availability chekcing code on stage to the satellite folks can validate it.

lindgrenj6:
It looks like an IQE failure, `AttributeError: 'IQEVaultReference' object has no attribute 'openapi_types'`, so probably safe to ignore.

MikelAlejoBR (approved):

lpichler (approved):

PR #127, opened 2022-02-23
[RHCLOUD-18182] Database DAO for authentications

MikelAlejoBR:
This PR:

- Creates a new config option, "VaultOn", which gets set to `true` if the `VAULT_ADDR` environment variable has been set.
- Creates a database DAO to query the authentications table.
- Adapts the response depending on whether Vault is online or not. The only difference in the model is the `ID` field, which will either be filled with Vault's ID or the database's auto generated ID.

## Links

[[RHCLOUD-18182]](https://issues.redhat.com/browse/RHCLOUD-18182)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -43,7 +43,7 @@ type SourcesApiConfig struct {
 	StatusListener            bool
 	MigrationsSetup           bool
 	MigrationsReset           bool
-	VaultOn                   bool
+	SecretStore               string
 }
 
 // Get - returns the config parsed from runtime vars
@@ -114,8 +114,12 @@ func Get() *SourcesApiConfig {
 	options.SetDefault("MarketplaceHost", os.Getenv("MARKETPLACE_HOST"))
 	options.SetDefault("SlowSQLThreshold", 2) //seconds
 	options.SetDefault("BypassRbac", os.Getenv("BYPASS_RBAC") == "true")
-	// We assume that if the "VAULT_ADDR" environment variable is set, Vault is online
-	options.Set("VaultOn", os.Getenv("VAULT_ADDR") != "")
+	// The secret store defaults to the database in case an empty or an incorrect value are provided.
+	secretStore := os.Getenv("SECRET_STORE")
+	if secretStore != "database" && secretStore != "vault" {
+		secretStore = "database"
+	}
+	options.SetDefault("SecretStore", secretStore)
 
 	// Parse any Flags (using our own flag set to not conflict with the global flag)
 	fs := flag.NewFlagSet("runtime", flag.ContinueOnError)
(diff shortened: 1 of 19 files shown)


lindgrenj6:
Ok I'd like to get this moving - if the tests are the problem lets just remove them so we can get this merged. That way we're one step closer to doing the cutover.

config/config.go
```diff
@@ -105,6 +106,8 @@ func Get() *SourcesApiConfig {
 	options.SetDefault("MarketplaceHost", os.Getenv("MARKETPLACE_HOST"))
 	options.SetDefault("SlowSQLThreshold", 2) //seconds
 	options.SetDefault("BypassRbac", os.Getenv("BYPASS_RBAC") == "true")
+	// We assume that if the "VAULT_ADDR" environment variable is set, Vault is online
+	options.Set("VaultOn", os.Getenv("VAULT_ADDR") != "")
```
lindgrenj6: I'm not completely in love with this - **but we can probably leave it for now, it's just the naming**.

---

But I'm wondering if instead we could have an environment variable that was `"SECRET_STORE"` (or something) and then have convenience methods on the config struct

OR just have config package level functions that look and see if the store is configured to be vault or not. e.g.

```go
func VaultOn() bool { 
	return config.SecretStore == "vault"
}
```

We'd probably default to Postgres as well - which is what we're doing here.

model/authentication.go
```diff
@@ -153,3 +162,22 @@ func (auth *Authentication) UpdateBy(attributes map[string]interface{}) error {
 func (auth *Authentication) Path() string {
 	return fmt.Sprintf("secret/data/%d/%s_%v_%s", auth.TenantID, auth.ResourceType, auth.ResourceID, auth.ID)
 }
+
+// mapIdExtraFields returns the ID and the Extra fields ready to be assigned to the response model, depending on
+func mapIdExtraFields(auth *Authentication) (string, map[string]interface{}) {
```
lindgrenj6, line 167: Chewing on this a bit. I'm wondering if we should make a struct that just contains the "changeable depending on the secret store" fields. In this case it would just be an id + extra but maybe would expand in the future (for instance `version` wouldn't be present I dont think).

lindgrenj6 (commented):
I have a couple very minor things - but otherwise looks good can't wait to get this merged. 

I will need to test this with /bulk_create but it should "just work" after updating the ApplicationAuthentication logic in the code.

config/config.go
MikelAlejoBR: Yeah, I like this idea more. I'm changing it!

model/authentication.go
MikelAlejoBR: I'm not sure I follow. What do you mean? Creating an embeddable struct that handles those fields?

lindgrenj6: I'm changing my  mind the more I think about it. I was just trying to think of a better way than just passing a tuple back and forth where we would have to know that this function basically just decides on which fields are the id/extra based on the secret store.

We can change it later if we want to.

MikelAlejoBR: Yep, I don't like that function either, to be honest. Feels really ugly.

lpichler (approved):
LGTM 👍

lindgrenj6 (approved):

PR #129, opened 2022-02-23
Fix StatusListener Test

lindgrenj6:
`reflect.DeepEqual` goes through and does checks the same way ruby does it - checking keys/values.

This should prevent this test from failing randomly when the json decides to marshal in a different order.


lindgrenj6:
Closing as I read this wrong. 

I think the issue is that the bulk messages are map[string]interfaces{} and those interfaces can contain pointers - it fails if the pointers aren't pointing at the same object. The regular single `ResourceJson` tests pass consistently though.

PR #131, opened 2022-02-24
[RHCLOUD-18192] fix: rows are closed error when listing RhcConnections

MikelAlejoBR:
The result of the "Next()" function —which checks if there are any more rows to process from the result set— was being ignored, therefore when there were no results when listing the Rhc Connections, the functions that came afterwards threw the "Rows are closed" error.

It also adds a test to avoid regressions.

## Links

[[RHCLOUD-18192]](https://issues.redhat.com/browse/RHCLOUD-18192)

### DIFF
--- a/dao/rhc_connection_dao_test.go
+++ b/dao/rhc_connection_dao_test.go
@@ -322,3 +322,48 @@ func TestRhcConnectionListForSources(t *testing.T) {
 
 	DoneWithFixtures(RHC_CONNECTION_SCHEMA)
 }
+
+// TestRhcConnectionRowsClosed is a regression test for https://issues.redhat.com/browse/RHCLOUD-18192. It tests that
+// when there are no rows to process from a result set, a proper response is returned instead of a "rows are closed"
+// error.
+func TestRhcConnectionRowsClosed(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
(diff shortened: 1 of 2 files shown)


dao/rhc_connection_dao_test.go
```diff
@@ -322,3 +322,44 @@ func TestRhcConnectionListForSources(t *testing.T) {
...
+	CreateFixtures(RHC_CONNECTION_SCHEMA)
+
+	dbRhcConnections := make([]model.RhcConnection, 0)
+	err := DB.Debug().
+		Model(&model.RhcConnection{}).
+		Find(&dbRhcConnections).
+		Error
+
+	if err != nil {
+		t.Errorf(`want nil error, gokt "%s"`, err)
```
lpichler: gokt -> got

MikelAlejoBR: Nice catch, thank you! :smile_cat: :+1:

lpichler: actually it looks that it is word from Dutch :)

MikelAlejoBR:
I'm going to kindly ask Petra to review it and I think we can merge this one, since it's not a big change :)

lpichler (approved):

petracihalova (approved):

PR #132, opened 2022-02-24
[RHCLOUD-18212] Refactor the way DAOs are created

MikelAlejoBR:
The reason behind this change is to decouple the instantiation of the DAOs from the code that needs them, in order to gain more flexibility when switching DAO implementations or when testing. Consider this example:

```go
// dao_file.go
type Dao interface {
	FindByAccountNumber(string accountNumber) (Something, error)
}

type dbDaoImpl struct {}

func (d* dbDaoImpl) FindByAccountNumber(string accountNumber) (Something, error) {
	// call db and return whatever
}

type vaultDaoImpl struct {}

func (d* dbDaoImpl) FindByAccountNumber(string accountNumber) (Something, error) {
	// call Vault and return whatever
}

// my_file_that_uses_the_dao.go
func myFunctionThatNeedsADao(string accountNumber) error {
	dao := dbDaoImpl{} // now this function is coupled to the dbDao, and we
	                   // cannot change to the vaultDao implementation.

	something, err := dao.FindByAccountNumber(&accountNumber)

	// ...

	return nil
}
```
In the above case the implementation is very coupled to the function that uses it, so... what could we do if we wanted to use the `VaultDaoImpl` instead? :thinking: 

```go
// dao_file.go
var GetDao func() Dao // returns the interface!

var getDefaultDao func() Dao {
	if config.VaultIsOnline {
		return &vaultDaoImpl{}
	} else {
		return &dbDaoImpl{}
	}
}

// init will set the function for obtaining the dao to the default one we have
// coded. If you want to use a different one, to return a mock dao for example,
// you only have to set another function to the exported GetDao variable!
func init() {
	GetDao = getDefaultDao
}

// my_file_that_uses_the_dao.go
func myFunctionThatNeedsADao(string accountNumber) error {
	// In this case we don't care which DAO implementation is underlying, since
	// we only care that we will be returned "something".
	dao := GetDao()
	something, err := dao.FindByAccountNumber(&accountNumber)

	// ...

	return nil
}
```

This is one way of doing it. We can override the `GetDao` function whenever we want if we need a different DAO.

---

So, this PR:

- Extracts the "get tenant from context" part of the code that was duplicated in every handler to a helper function.
- Creates a `var GetDao*` function for every DAO, and initializes it with a default DAO provider on the `init()` function, so that every other package has only one way of using DAOs, and so that those packages work against interfaces and not final implementations.
- Makes the `*Impl` structs unexported so that nobody can instantiate a DAO directly.

## Links

[[RHCLOUD-18212]](https://issues.redhat.com/browse/RHCLOUD-18212)

### DIFF
--- a/meta_data_handlers.go
+++ b/meta_data_handlers.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"net/http"
 	"strconv"
 
@@ -15,19 +14,16 @@ import (
 var getMetaDataDao func(c echo.Context) (dao.MetaDataDao, error)
 
 func getMetaDataDaoWithTenant(c echo.Context) (dao.MetaDataDao, error) {
-	var tenantID int64
-	var ok bool
-	tenantVal := c.Get("tenantID")
+	tenantId, err := getTenantFromEchoContext(c)
 
-	// if we set the tenant on this request - include it. otherwise do not.
-	if tenantVal != nil {
-		if tenantID, ok = tenantVal.(int64); !ok {
-			return nil, fmt.Errorf("failed to pull tenant from request")
-		}
+	if err != nil {
+		return nil, err
+	}
 
-		return &dao.MetaDataDaoImpl{TenantID: &tenantID}, nil
-	} else {
+	if tenantId == 0 && err == nil {
 		return &dao.MetaDataDaoImpl{}, nil
+	} else {
+		return &dao.MetaDataDaoImpl{TenantID: &tenantId}, nil
 	}
 }
 
--- a/meta_data_handlers.go
+++ b/meta_data_handlers.go
@@ -21,9 +21,9 @@ func getMetaDataDaoWithTenant(c echo.Context) (dao.MetaDataDao, error) {
 	}
 
 	if tenantId == 0 && err == nil {
-		return &dao.MetaDataDaoImpl{}, nil
+		return dao.GetMetaDataDao(nil), nil
 	} else {
-		return &dao.MetaDataDaoImpl{TenantID: &tenantId}, nil
+		return dao.GetMetaDataDao(&tenantId), nil
 	}
 }
 
--- a/application_authentication_handlers.go
+++ b/application_authentication_handlers.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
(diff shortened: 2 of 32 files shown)


lindgrenj6 (approved):
Awesome, thanks for getting this out so quick. This will enforce how we isntantiate db dao's all over the app! I'll have to refactor a bunch in my other PRs haha

meta_data_handlers.go
```diff
@@ -15,19 +14,16 @@ import (
...
-		if tenantID, ok = tenantVal.(int64); !ok {
-			return nil, fmt.Errorf("failed to pull tenant from request")
-		}
+	if err != nil {
+		return nil, err
+	}
 
-		return &dao.MetaDataDaoImpl{TenantID: &tenantID}, nil
-	} else {
+	if tenantId == 0 && err == nil {
```
lpichler, line 23: it looks that `err == nil` is the only value possible here as `err != nil` is covered on line 19.

I guess  `err == nil` is not needed here.

lpichler:
it looks that commit with `SourceTypeDaoImpl`  is missing, is that on purpose ?

MikelAlejoBR:
> Awesome, thanks for getting this out so quick. This will enforce how we isntantiate db dao's all over the app! I'll have to refactor a bunch in my other PRs haha

Instead of merging this PR, I don't mind waiting and then rebasing main and doing it myself. The same for any other big complicated PR you might be working on @lpichler . I don't think it is fair for me to burden you with some unrelated changes to your work, so I'm totally fine with waiting! :smile: :+1: 

> it looks that commit with `SourceTypeDaoImpl` is missing, is that on purpose ?

Great catch! :fishing_pole_and_fish: I missed that one :hand_over_mouth: 

---

Also: I changed `common.go` so that it uses also the DAO providers. I know it lives in the same "dao" package and that we could cheat there and use the final implementations, but I think we're better off if we use the providers here too: that way once we implement the database authentication DAO the `common.go` code will work with both the Vault and the DB implementations! :smile:

petracihalova (approved):
Looks great. I really appreciate the way you split changes into commits 👍🏻 .

MikelAlejoBR:
Thank you! And thanks for the review :+1:

PR #134, opened 2022-02-24
Add Basic Authentication Validation + Authentication event-stream raising

lindgrenj6:
This was mostly done for https://github.com/RedHatInsights/sources-superkey-worker/pull/48 since I need the resource ID to be a string || integer for the rails app, also more flexibility doesn't hurt. 

---

I also turned on the event-stream raising for authentications since we satisfy that interface now.

### DIFF
--- a/authentication_handlers.go
+++ b/authentication_handlers.go
@@ -98,8 +98,7 @@ func AuthenticationCreate(c echo.Context) error {
 		return c.JSON(http.StatusBadRequest, util.ErrorDoc(err.Error(), "400"))
 	}
 
-	// TODO: once ToEvent() is added for authentication un-comment this.
-	// setEventStreamResource(c, auth)
+	setEventStreamResource(c, auth)
 	return c.JSON(http.StatusCreated, auth.ToResponse())
 }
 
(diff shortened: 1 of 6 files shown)


MikelAlejoBR (approved):

lpichler:
Do we need also other validations ?

1.availability_status -  https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb#L16 
2.only_one_superkey - https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb#L19 
3.both_username_and_password - https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb#L20

lindgrenj6:
> Do we need also other validations ?
> 
> 1.availability_status - [https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb#L16](https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb?rgh-link-date=2022-02-24T18%3A12%3A28Z#L16) 2.only_one_superkey - [https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb#L19](https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb?rgh-link-date=2022-02-24T18%3A12%3A28Z#L19) 3.both_username_and_password - [https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb#L20](https://github.com/RedHatInsights/sources-api/blob/master/app/models/authentication.rb?rgh-link-date=2022-02-24T18%3A12%3A28Z#L20)

1. No we don't need that for create requests, maybe for update requests in the future. We also do not display the availability status for authentications in the UI at all. It is just there in case we wanted to support it in the future eventually.

2 & 3. I'll do those in a follow-up - they aren't necessary except for QE tests. I mostly just need the "raw" varialbe support for the superkey refactor. I changed the title to "basic" validation since it isn't all of it haha

lpichler (approved):

PR #141, opened 2022-03-04
Add test to the "authFromVault" function

MikelAlejoBR:
I thought I could refactor that function so that it used `json.Marshal` instead of manually doing everything, but after writing the test I tried to replace the code and realized that Vault returns a `map[string]interface{}`, not a `[]byte`, so there's not much benefit in trying to unmarshal the secret differently to what we do now.

Since I already wrote the test, I thought "let's just keep it".

### DIFF
--- a/dao/authentication_dao_test.go
+++ b/dao/authentication_dao_test.go
@@ -4,12 +4,15 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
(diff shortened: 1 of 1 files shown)


lindgrenj6 (approved):
Sweet, thanks for these. If this format breaks it would be very bad so hopefully this would prevent that or at least alert us to any changes.

PR #142, opened 2022-03-07
[RHCLOUD-18037] Tests for authentication handler

petracihalova:
[RHCLOUD-18037](https://issues.redhat.com/browse/RHCLOUD-18037)

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/dao"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/database"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/fixtures"
+	"github.com/RedHatInsights/sources-api-go/internal/testutils/mocks"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/parser"
 	l "github.com/RedHatInsights/sources-api-go/logger"
 	"github.com/RedHatInsights/sources-api-go/middleware"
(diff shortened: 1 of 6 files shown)


main_test.go
```diff
@@ -45,6 +47,12 @@ func TestMain(t *testing.M) {
 		getMetaDataDao = getMetaDataDaoWithTenant
 		getRhcConnectionDao = getDefaultRhcConnectionDao
 		getApplicationAuthenticationDao = getApplicationAuthenticationDaoWithTenant
+		getAuthenticationDao = getAuthenticationDaoWithTenant
+
+		dao.Vault = &mocks.MockVault{}
```
MikelAlejoBR, line 52: If its an integration test Vault shouldn't be mocked, should be?

petracihalova: Well my idea was, that before my change there was only 1 test where we used this MockVault and it is test for status listener. So I wanted to do same thing in auth tests. There was option to add this mock into every particular test which needs Vault or move it in main_test.go.

MikelAlejoBR: You are right, yes. Let's keep it this way then.

---

What I was saying —which I really didn't notice even though I've been working on it—, is that we're mocking Vault on integration tests, which I guess we shouldn't be doing since they're integration tests. Since this isn't the purpose of this PR I've opened a ticket so we can discuss it there: https://issues.redhat.com/browse/RHCLOUD-18285

petracihalova: Yes, that makes sense. Let's discuss.

lindgrenj6 (approved):

petracihalova:
rebase + solved merge conflicts

lpichler (approved):

lindgrenj6:
Sorry Petra but I created more 😔

PR #147, opened 2022-03-08
Add the marketplace extra fields to the openapi spec

dehort:
I'm not sure if this is right or not.  Based on [1], it looks like the extra object can contain a "marketplace" entry.

[1] https://issues.redhat.com/browse/RHCLOUD-17515


MikelAlejoBR (approved):
That is correct, thank you for your PR!

lpichler (approved):
Thanks!

PR #148, opened 2022-03-08
Handle already-registered RHC Connection

lindgrenj6:
Shim ran into this while integrating - we were returning a 500 if the RHC Connection was registered to the Source. 

---

To fix this we just need to do a `.FirstOrCreate` on the join table record - if the `result.RowsAffected` is 0 then we know nothing was inserted and thus the source is already registered. On that case we return a bad request error rather than just a general 500 error.


MikelAlejoBR (approved):
Looking good! :+1:

lpichler (approved):

PR #151, opened 2022-03-09
Add alltest to makefile to run both unit + integration tests

lindgrenj6:
for @petracihalova  :) 

This make target runs `go test  ./...` and then `go test --integration ./...`


petracihalova (approved):
❤️

MikelAlejoBR (approved):

PR #152, opened 2022-03-09
ApplicationAuthentication Create+Delete

lindgrenj6:
Somehow these methods were missed - I think `PATCH` isn't needed - but will implement it if I get time this afternoon.

### DIFF
--- a/application_authentication_handlers_test.go
+++ b/application_authentication_handlers_test.go
@@ -1,12 +1,14 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"net/http"
 	"testing"
 
 	"github.com/RedHatInsights/sources-api-go/internal/testutils"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/fixtures"
+	"github.com/RedHatInsights/sources-api-go/internal/testutils/parser"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/request"
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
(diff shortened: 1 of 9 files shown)


application_authentication_handlers_test.go
```diff
@@ -179,3 +181,101 @@ func TestApplicationAuthenticationGetBadRequest(t *testing.T) {
...
+		"/api/sources/v3.1/application_authentications/1234523452542",
+		nil,
+		map[string]interface{}{
+			"tenantID": int64(1),
+		},
+	)
+	c.SetParamNames("id")
+	c.SetParamValues("1234523452542")
+
+	err := ApplicationAuthenticationDelete(c)
```
petracihalova: I am thinking 🤔  in this case we expect 404, so we can check more than that there is some error. For example same logic as is in TestApplicationAuthenticationGetNotFound(). 

```
	notFoundApplicationAuthenticationDelete := ErrorHandlingContext(ApplicationAuthenticationDelete)
	err := notFoundApplicationAuthenticationDelete(c)
	if err != nil {
		t.Error(err)
	}

	testutils.NotFoundTest(t, rec)
```

... and same for new bad request tests ... or  😁. I can update these test after merging in new PR

petracihalova (approved):

application_authentication_handlers_test.go
lindgrenj6: Thanks for catching that, I was completely on auto-pilot and just wrote them the old way :grin:. 

Updated to use the new helpers.

petracihalova (approved):

MikelAlejoBR (approved):

lpichler (approved):

PR #153, opened 2022-03-10
[RHCLOUD-18280] Source handler missing tests

petracihalova:
[RHCLOUD-18280](https://issues.redhat.com/browse/RHCLOUD-18280)

### DIFF
--- a/dao/authentication_dao.go
+++ b/dao/authentication_dao.go
@@ -106,6 +106,13 @@ func (a *authenticationDaoImpl) List(limit int, offset int, filters []util.Filte
 }
 
 func (a *authenticationDaoImpl) ListForSource(sourceID int64, _, _ int, _ []util.Filter) ([]m.Authentication, int64, error) {
+	// Check if sourceID exists
+	src := &m.Source{ID: sourceID}
+	result := DB.First(src)
+	if result.Error != nil {
+		return nil, 0, util.NewErrNotFound("source")
+	}
+
 	keys, err := a.listKeys()
 	if err != nil {
 		return nil, 0, err
--- a/dao/mock_daos.go
+++ b/dao/mock_daos.go
@@ -94,6 +94,7 @@ func (src *MockSourceDao) GetById(id *int64) (*m.Source, error) {
 }
 
 func (src *MockSourceDao) Create(s *m.Source) error {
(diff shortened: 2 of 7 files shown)


dao/authentication_dao.go
```diff
@@ -124,6 +124,10 @@ func (a *authenticationDaoImpl) ListForSource(sourceID int64, _, _ int, _ []util
 		}
 	}
 
+	if len(out) == 0 {
+		return nil, 0, util.NewErrNotFound("source")
+	}
+
```
MikelAlejoBR: I think that this code block is not needed, since we are in a `List*` function. When you are querying for a collection in a REST API, having the empty collection "[]" as the indicator that there are no elements is a valid response. There is no need to return a 404 in that case.

petracihalova: Yes, you're right. My idea was to handle scenario when source doesn't exists. Because endpoint is `GET /api/sources/v3.1/sources/{id}/authentications`. I prepared second version, where I try to find out if source exists or not and then continue to get list of authentications.

MikelAlejoBR (approved):

dao/mock_daos.go
```diff
@@ -555,8 +555,30 @@ func (m MockAuthenticationDao) GetById(id string) (*m.Authentication, error) {
...
 }
 
-func (m MockAuthenticationDao) ListForSource(sourceID int64, limit, offset int, filters []util.Filter) ([]m.Authentication, int64, error) {
-	panic("implement me")
+func (mAuth MockAuthenticationDao) ListForSource(sourceID int64, limit, offset int, filters []util.Filter) ([]m.Authentication, int64, error) {
+	sourceExists := false
+
+	for _, src := range fixtures.TestSourceData {
+		if src.ID == sourceID {
+			sourceExists = true
```
lpichler, line 566: I think you can return already here and we don't need var `sourceExists`.

lpichler:
I had just minor comment otherwise LGTM 👍

dao/mock_daos.go
petracihalova: 🧐. I dont think so ... I need raise Not Found Error if source doesn't exist

lpichler: I see!

lpichler (approved):

PR #154, opened 2022-03-10
[RHCLOUD-18328] Feature disallow update paused resources

MikelAlejoBR:
Disallows updating sources and applications if they're paused, returning a 400 in those cases.

## Link

[[RHCLOUD-18328]](https://issues.redhat.com/browse/RHCLOUD-18328)

### DIFF
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -141,11 +141,6 @@ func ApplicationEdit(c echo.Context) error {
 		return err
 	}
 
-	input := &m.ApplicationEditRequest{}
-	if err := c.Bind(input); err != nil {
-		return err
-	}
-
 	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
 	if err != nil {
 		return util.NewErrBadRequest(err)
@@ -156,8 +151,28 @@ func ApplicationEdit(c echo.Context) error {
 		return util.NewErrNotFound("application")
 	}
 
+	// Store the previous status before updating the application.
 	previousStatus := app.AvailabilityStatus
-	app.UpdateFromRequest(input)
+
+	if app.PausedAt != nil {
+		input := &m.ResourceEditPausedRequest{}
+		if err := c.Bind(input); err != nil {
+			return util.NewErrBadRequest(err)
+		}
+
+		err := app.UpdateFromRequestPaused(input)
+		if err != nil {
+			return util.NewErrBadRequest(err)
+		}
+	} else {
+		input := &m.ApplicationEditRequest{}
+		if err := c.Bind(input); err != nil {
+			return util.NewErrBadRequest(err)
+		}
+
+		app.UpdateFromRequest(input)
+	}
+
 	err = applicationDB.Update(app)
 	if err != nil {
 		return err
--- a/application_handlers_test.go
+++ b/application_handlers_test.go
@@ -9,7 +9,9 @@ import (
 	"net/http"
 	"reflect"
 	"strconv"
+	"strings"
 	"testing"
+	time "time"
 
 	"github.com/RedHatInsights/sources-api-go/dao"
 	"github.com/RedHatInsights/sources-api-go/internal/events"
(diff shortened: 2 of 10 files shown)


application_handlers_test.go
```diff
@@ -854,3 +856,114 @@ func applicationEventTestHelper(t *testing.T, c echo.Context, expectedEventType
...
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	if rec.Code != 400 {
```
petracihalova: Only because you have this different in previous test 😁  lets use here http.StatusBadRequest too.

application_handlers_test.go
```diff
@@ -854,3 +856,114 @@ func applicationEventTestHelper(t *testing.T, c echo.Context, expectedEventType
...
+
+	want := "can't modify a paused application. Unpause it first and try again"
+	got := rec.Body.String()
+	if !strings.Contains(got, want) {
+		t.Errorf(`unexpected body returned. Want "%s" contained in what we got "%s"`, want, got)
+	}
+}
+
+// TestApplicationEditPausedUnit tests that a "bad request" response is returned when a paused application is tried to
+// be updated. Runs on unit tests.
```
petracihalova: I am not sure about "Runs on unit tests". Does this mean that this test is not running with integration tests? I am asking because I cannot see here something like `testutils.SkipIfNotRunningIntegrationTests(t)`.

dao/mock_daos.go
```diff
@@ -157,6 +161,10 @@ func (s *MockSourceDao) Unpause(_ int64) error {
 	return nil
 }
 
+func (s *MockSourceDao) IsPaused(_ *int64) (bool, error) {
+	return false, nil
```
petracihalova: Maybe here should be return `s.IsPausedReturnValue, nil`?

source_handlers_test.go
```diff
@@ -1136,3 +1139,113 @@ func sourceEventTestHelper(t *testing.T, c echo.Context, expectedEventType strin
...
+	got := rec.Body.String()
+	if !strings.Contains(got, want) {
+		t.Errorf(`unexpected body returned. Want "%s" contained in what we got "%s"`, want, got)
+	}
+}
+
+// TestSourceEditPausedUnit tests that a "bad request" response is returned when a paused source is tried to
+// be updated. Runs on unit tests.
+func TestSourceEditPausedUnit(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
```
petracihalova: In comment is "Runs on unit tests" and then you skip if unit tests are running

petracihalova (changes requested):
Great job 👍🏻  I have only few small comments.

application_handlers_test.go
MikelAlejoBR: Nice catch, thank you!

application_handlers_test.go
MikelAlejoBR: I rephrased those comments and hopefully are more clear now.

dao/mock_daos.go
MikelAlejoBR: Nice catch :fish: !

source_handlers_test.go
MikelAlejoBR: You are right, I copied and pasted the test, and forgot to modify it :facepalm: now the unit tests runs on unit tests :laughing:

MikelAlejoBR:
Thanks for your review @petracihalova! :+1:

application_handlers_test.go
petracihalova: 🧐. ... so this test is running with unit (mocked) test and with integration test, but in both cases the test handle the job as unit (mocked) test?

MikelAlejoBR: Exactly, yes. I did this to avoid having something like this in the tests:

```go
var backupDao func(c echo.Context) (dao.SourceDao, error)
if parser.RunningIntegrationTests {
	// Modify the source so that the underlying code identifies it as "paused".
	err := dao.DB.Model(m.Source{}).Where("id = ?", 1).UpdateColumn("paused_at", time.Now()).Error
	if err != nil {
		t.Error(err)
	}
} else {
	// Get the specific SourceDao mock which simulates that the sources are paused.
	backupDao = getSourceDao
	getSourceDao = func(c echo.Context) (dao.SourceDao, error) {
		return &dao.MockSourceDao{IsPausedReturnValue: true}, nil
	}

}

badRequestSourceEdit := ErrorHandlingContext(SourceEdit)
err := badRequestSourceEdit(c)
if err != nil {
	t.Error(err)
}

if parser.RunningIntegrationTests {
	// Revert the changes so other tests don't have any problems.
	err = dao.DB.Model(m.Source{}).Where("id = ?", 1).UpdateColumn("paused_at", nil).Error
	if err != nil {
		t.Error(err)
	}
} else {
	// Go back to the previous DAO mock.
	getSourceDao = backupDao
}
```

Since all the integration tests begin with the `testutils.SkipIfNotRunningIntegrationTests(t)` I thought of following the same pattern at the expense of writing a "kind-of-duplicated" tests for this feature. Just for consistency.

petracihalova (approved):

application_handlers_test.go
petracihalova: ok thanks for the explanation, I was not sure if I understand correctly that this unit test is running when integration tests run

dao/application_dao.go
```diff
@@ -173,3 +173,19 @@ func (a *applicationDaoImpl) Unpause(id int64) error {
...
+func (a *applicationDaoImpl) IsPaused(id *int64) (bool, error) {
+	var isPaused bool
+
+	err := DB.Debug().
+		Model(&m.Application{}).
+		Select("1").
+		Where("id = ?", id).
+		Where("tenant_id = ?", a.TenantID).
+		Where("applications.paused_at IS NOT NULL").
+		Where("DATE(applications.paused_at) <> '0001-01-01'"). // time.Time's default value
```
lpichler: Where `0001-01-01` came from ? (I have it also in my local DB) Is it something related to go init date values ?

MikelAlejoBR: Yes, as the `Pause` structure's `paused_at` member is a `time.Time` and not a `*time.Time`, we cannot insert null `paused_at` values in the database using Gorm. I firstly tried to change it on this PR but when I saw it affected multiple files I went the conservative way and added another `Where` clause to make it work.

lindgrenj6:
I think we can use this as an excuse to remove the `Pause` struct embeds
all over and change `PausedAt` to a `*time.Time` :grin:
--
Jacob Lindgren

Red Hat

lindgrenj6:
(in a followup)
--
Jacob Lindgren

Red Hat

MikelAlejoBR:
Agree. I've opened a ticket so we don't forget about it: https://issues.redhat.com/browse/RHCLOUD-18371

lindgrenj6:
I swear there is a ticket for that in the backlog - but Jira won't let me
see it.
--
Jacob Lindgren

Red Hat

lpichler:
@MikelAlejoBR thanks for doing this!

we disallow paused resources but we allow to update some attributes for paused resources: `availability_status, availability_status_error, last_checked_at, last_available_at` (if available on resource)

NOTE: it could end in multipart response

Please check description for more info https://github.com/RedHatInsights/sources-api/pull/421

MikelAlejoBR:
Ooops! My bad. I thought pausing a resource was like "freezing" or "disabling" it, so that's why I assumed it would return a 400 and that's it.

It feels odd thought that we allow updating fields like the "availability status" and not the name... I would certainly see a client wanting to rename a source/application while they are paused. Why is this?

Thanks :)

lpichler:
you are right, @MikelAlejoBR you can read more background here: https://issues.redhat.com/browse/RHCLOUD-9850 I think that we need to update available status only because that need to send unavailable status when cloud account is disabled in cost side:

>  If we detect a customer's AWS access is no longer sufficient for our needs, we disable the cloud account object on our side, and we send an update to the sources-api application with the "unavailable" status. 

Hopefully it brings more light on it otherwise we need to ask @lindgrenj6 :)

lindgrenj6:
> Agree. I've opened a ticket so we don't forget about it: https://issues.redhat.com/browse/RHCLOUD-18371

I found it: https://issues.redhat.com/browse/RHCLOUD-18064

petracihalova:
Yes, it's in my queue ... I haven't made it yet

MikelAlejoBR:
> you are right, @MikelAlejoBR you can read more background here: https://issues.redhat.com/browse/RHCLOUD-9850 I think that we need to update available status only because that need to send unavailable status when cloud account is disabled in cost side:
> 
> > If we detect a customer's AWS access is no longer sufficient for our needs, we disable the cloud account object on our side, and we send an update to the sources-api application with the "unavailable" status.
> 
> Hopefully it brings more light on it otherwise we need to ask @lindgrenj6 :)

Thank you for the background Libor, now it's all clear! :) :+1: 

> > Agree. I've opened a ticket so we don't forget about it: https://issues.redhat.com/browse/RHCLOUD-18371
> 
> I found it: https://issues.redhat.com/browse/RHCLOUD-18064

Cool, I marked my ticket as a duplicate and transferred the comment about making the field a pointer to `time.Time`.

MikelAlejoBR:
I would like to ask for your opinion on my last commit. Essentially, I had to manually bind the incoming data from the request to a source, keep track of which data was updated and which one is bogus data, and then build the JSON manually to match the OpenApi spec.

This code isn't tidied up, since I could improve it —maybe— by adding a struct here and there and extracting functions as well. Also it lacks tests. However the reason for stopping here is that I didn't want to commit to investing more time until we discussed this matter a bit, because my code "feels ugly".

I don't know if it is because of my tunnel vision, but I don't see other way right now to return a multi status data including the updated keys, the ones that didn't get updated, etc.

Can you please take a look and tell me if you see any other way? Thank you very much in advance!

---

The other way would be much simpler, but it wouldn't exactly respect the OpenApi spec:

- Create a new `SourcePausedEditRequest` with the three allowed fields.
- Try to bind the input to that struct, and if it fails, return a 400 telling the client that the keys they provided are not allowed.
- Return a 200 with the updated object if the operation succeeded.

source_handlers.go
```diff
@@ -149,7 +149,57 @@ func SourceEdit(c echo.Context) error {
 		return err
 	}
 
-	s.UpdateFromRequest(input)
+	if isPaused {
+		binder := SourcePausedEditBinder{}
```
lindgrenj6: I agree this puts a weird taste in my mouth. We also wouldn't want to use a binder by itself here since we can theoretically only have one binder at the echo router level. I see this as more of a validation in a service class if we did go this route. 

----

fwiw I would rather we have a `SourcePauseEdit` struct and let the "regular" binder return an error, and return a 400 that way. I'm not sure the multi-response is necessary since 99.9% of the time the user is just updating availability status.

application_handlers.go
```diff
@@ -141,16 +141,25 @@ func ApplicationEdit(c echo.Context) error {
...
-	input := &m.ApplicationEditRequest{}
-	if err := c.Bind(input); err != nil {
-		return err
+	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	if err != nil {
+		return util.NewErrBadRequest(err)
 	}
 
-	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	isPaused, err := applicationDB.IsPaused(&id)
```
petracihalova: Here when err is not nil, I believe it is not 400, but 500, so we can return only err

petracihalova (changes requested):
LGTM I have only small comment

source_handlers.go
```diff
@@ -132,16 +132,25 @@ func SourceEdit(c echo.Context) error {
...
-	input := &m.SourceEditRequest{}
-	if err := c.Bind(input); err != nil {
-		return err
+	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	if err != nil {
+		return util.NewErrBadRequest(err)
 	}
 
-	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	isPaused, err := sourcesDB.IsPaused(&id)
```
petracihalova: same here ... the error occurs on server side

lpichler:
CI failure looks related to the PR

application_handlers.go
MikelAlejoBR: Thank you for the comments but I ended up removing the `IsPaused` functions since they're no longer necessary on this PR. I need to fetch the resource from the DB to be able to update it in any case, so I can check if the `PausedAt` member is nil directly, without having to go to the database again.

lpichler (approved):

lindgrenj6 (approved):

PR #155, opened 2022-03-10
Only Seed the database when running the API Pod

lindgrenj6:
I noticed this when testing the statuslistener on stage - we are seeding the database on startup on both the API pod AND the availability status listener. 

We only want to see the DB one time per rollout, and the API pod makes the most sense to handle that. 

---

This involved moving the flag parsing to the `config/` package - which makes sense due to the fact that we parse the config once and we can refer to that struct for everything from then on. 

---

I also moved the type cache initialization to the dao init function - makes more sense there.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -106,9 +106,14 @@ func Get() *SourcesApiConfig {
 	options.SetDefault("SlowSQLThreshold", 2) //seconds
 	options.SetDefault("BypassRbac", os.Getenv("BYPASS_RBAC") == "true")
 
-	// Parse any Flags
-	availabilityListener := flag.Bool("listener", false, "run availability status listener")
-	flag.Parse()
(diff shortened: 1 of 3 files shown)


MikelAlejoBR (approved):

lpichler:
I am sorry, conflict also here.

lpichler:
CI failure is related  - I guess  that it needs parse flags in `TestMain` function as we  have in `main_test.go` as is in `statuslistener` or `service` package.

lindgrenj6:
Ended up switching our "real" flags to use a `FlagSet`. That way they are completely separate from the global flag instance.

lindgrenj6:
Nevermind apparently that worked locally but not up in the runners. Weird.

lindgrenj6:
There we go. Ready for review.

lpichler (approved):
👍

PR #156, opened 2022-03-10
Handle Int + String resource_id's in the status listener

lindgrenj6:
A few small refactors on the way we parse statusmessages - I tested this on stage and it was erroring out on some payloads. It looks like we seamlessly handles integer and/or string status message resource_id's. 

This PR handles both of them. I also just moved the function that translates headers to a method on the message object - since that is where the state is.

### DIFF
--- a/internal/types/status_message.go
+++ b/internal/types/status_message.go
@@ -1,8 +1,9 @@
 package types
 
 type StatusMessage struct {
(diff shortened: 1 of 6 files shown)


MikelAlejoBR (approved):

lpichler (approved):
Thanks @lindgrenj6 for fixing this.

PR #158, opened 2022-03-11
Fix Availability Status Warn condition

lindgrenj6:
Saw this while testing - we were getting a warn for 204 and I realized I used `%` instead of `/`. This will eliminate the warn message since anything 2xx is not a problem.


MikelAlejoBR (approved):

lpichler (approved):

lpichler:
I like size of this PR :)

lindgrenj6:
`+1/-1` are the best kind!

PR #163, opened 2022-03-15
[RHCLOUD-18316] Source and application types for AppStudio

MikelAlejoBR:
Adds the following source types:

- Bitbucket
- Github
- Gitlab
- Quay
- Dockerhub

And the following app type:

- AppStudio

## Links

[[RHCLOUD-18316]](https://issues.redhat.com/browse/RHCLOUD-18316)

### DIFF
--- a/dao/seeds/application_types.yml
+++ b/dao/seeds/application_types.yml
@@ -1,4 +1,25 @@
 ---
+"/insights/platform/app-studio":
+  display_name: App Studio
+  dependent_applications: []
+  supported_source_types:
+    - bitbucket
+    - dockerhub
+    - github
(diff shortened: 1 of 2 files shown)


petracihalova (approved):

lindgrenj6:
~Initially I think we were more going to have 2 generic source types like "Git Repository" and "Container Registry", so we could keep it a bit more generic than having a source type for every SCM/Registry provider.~

Actually NVM, it might be better to have them separate. That way they don't have to store anything in the application.extra column to determine what type of git application it is.

lindgrenj6 (approved):

PR #165, opened 2022-03-16
[RHCLOUD-17965] Refactor delete returning

MikelAlejoBR:
This PR replaces the `Delete` methods from the DAOs so that they use SQL's `RETURNING` statement. That way we can spare the "does this resource exist?" query to the database.

**Note:** requires https://github.com/RedHatInsights/sources-api-go/pull/167 to be merged first.

## Links

[[RHCLOUD-17965]](https://issues.redhat.com/browse/RHCLOUD-17965)

### DIFF
--- a/dao/application_authentication_dao.go
+++ b/dao/application_authentication_dao.go
@@ -5,6 +5,7 @@ import (
 
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
+	"gorm.io/gorm/clause"
 )
 
 // GetApplicationAuthenticationDao is a function definition that can be replaced in runtime in case some other DAO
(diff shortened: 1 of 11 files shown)


MikelAlejoBR:
~I have been trying to use the "RETURNING" clause with Gorm, as specified in https://gorm.io/docs/delete.html#Returning-Data-From-Deleted-Rows, but it doesn't seem to work for me. I've tried specifying the individual columns just for the sake of testing, but it doesn't work either.~

~Also, as a test, I've tried using `Raw` and `Exec` statements along with `Scan`, to try to do it "manually" but then let Gorm do the mapping. With `Raw` it both doesn't get mapped and we don't get the `RowsAffected` metadata, although the query gets successfuly executed.~

~With `Exec` you can't run `Scan` since it seems to do a new `SELECT` query to get the just deleted resource —Gorm seems to interpret the `Exec` and `Scan` as two separate queries—, and it breaks if you just try to append the `.Scan()` code to the previous `Exec()` query.~

~I would vote for ruling out this refactor entirely since the only other way I see we could do this is by doing a `Raw()`, grabbing the rows somehow —which I haven't been able to neither— and mapping the resources manually, but it seems a lot of effort for very little gain.~

~What are you thoughts, @RedHatInsights/sources-committers?~

~**EDIT:** whoops, I mentioned the wrong group, sorry.~

After much debugging —thanks Jacob for the help— we've realized that we are using an outdated version of Gorm and its Postgres driver. I'll try updating both and see if I can make it work, since with the newer versions we're able to use the `RETURNING` helper!

lindgrenj6 (approved):
This looks good but may need a rebase I think. Will merge once rebased!

MikelAlejoBR:
Rebased!

lpichler (approved):

PR #167, opened 2022-03-17
Bump Gorm and Postgres versions

MikelAlejoBR:
Related to: https://github.com/RedHatInsights/sources-api-go/pull/165

The older version of Gorm or its Postgres driver has a bug which doesn't add the "RETURNING" statement to the queries even if you explicitly specify it with `clauses.Returning{}`. By bumping the Gorm and Postgres driver versions this issue is fixed.

There is a caveat, though: the version bump also affects to how `RowsAffected` gets updated when using `FirstOrCreate`. In the old version of the libraries, if `FirstOrCreate` found the object in the database, `RowsAffected` would be zero. However, in the newer versions, if the object gets created OR gets found, in both cases `RowsAffected` will be 1. So we can't rely on that anymore to tell if the `sourceRhcConnection` exists or not.

That was making one test fail, and that's why I suggest and vote for refactoring it.


petracihalova (approved):

dao/rhc_connection_dao.go
```diff
@@ -168,25 +167,38 @@ func (s *rhcConnectionDaoImpl) Create(rhcConnection *m.RhcConnection) (*m.RhcCon
...
+		if err != nil {
+			return err
 		}
 
-		if result.Error != nil {
-			return fmt.Errorf("cannot link red hat connection to source: %w", result.Error)
+		// If it exists, we let the client know. If it doesn't, we attempt to create it.
+		if relationExists {
+			return util.NewErrBadRequest("connection already exists")
+		} else {
```
lindgrenj6: I think per Go conventions we can just remove this `else`. Returning early handles the error fine.

lindgrenj6 (commented):
Just one tiny comment, otherwise LGTM!

dao/rhc_connection_dao.go
MikelAlejoBR: All right! Done :)

lindgrenj6 (approved):

PR #168, opened 2022-03-18
[RHCLOUD-18064] Remove availability status and pause structs in models and replace them by regular fields

petracihalova:
[RHCLOUD-18064](https://issues.redhat.com/browse/RHCLOUD-18064)

Removed pause and availability status structs in models and converted PauseAt fields from time.Time to pointer. Because we have now some time fields as time.Time and some as pointers, I created a new functions in parser.go. I didn't change the existing because then there should be an interface and I believe now it is more simple. But I am open to any suggestions.


MikelAlejoBR (commented):
I only want to make one comment:

The `last_checked_at` and `last_available_at` columns in the different tables of the database don't have a `NOT NULL` constraint, so my idea is that we can make them pointer too. That way when producing an output, thanks to the `omitempty` tags on the models, we wouldn't return those keys instead of returning the default values we are returning now.

Other than that, the PR looks good :+1: !

petracihalova:
> 
Not sure if I understand what you mean. Now we creating an output, these two fields are strings, so thanks to the `omitempty` when value is empty (empty string) then we don't return those keys.

MikelAlejoBR:
Ooooooh I keep forgetting that we return empty strings and not the default date. All right, then! :+1:

lindgrenj6:
I agree with @MikelAlejoBR - we should make the LastCheckedAt/LastAvailableAt fields in the "core" structs like Application/Source pointers so the empty value doesn't get populated into the DB, like so: 
image

The http response looks right - but theoretically I think those should be `null` in the DB rather than the zero-date.

petracihalova:
Added new commit - LastAvailableAt and LastCheckedAt fields changed into pointers

lindgrenj6 (approved):

MikelAlejoBR (approved):

lindgrenj6:
I really want to merge this - but it's going to break a lot of PRs mergability. Is everyone ok with that? Ready to rebase like crazy?

MikelAlejoBR:
> Ready to rebase like crazy?

story-of-my-life

petracihalova:
rebase done & conflicts solved

lindgrenj6 (approved):

lindgrenj6:
https://media.giphy.com/media/GwRBmXyEOvFtK/giphy.gif

lpichler (approved):

PR #169, opened 2022-03-18
small fix  in app auth handlers test

petracihalova:
without jira ticket
only tiny fix


MikelAlejoBR (approved):
Great catch! :fishing_pole_and_fish:

lpichler (approved):
Thanks!

PR #170, opened 2022-03-18
add new line to standard error output

petracihalova:
without JIRA ticket

I had interesting troubles in Golang with tests results. My test results tab didn't contain all tests.
e.g. here I ran 4 tests, but only 3 are visible:
image

and console output:
```
flag provided but not defined: -test.v
Usage of runtime:
  -listener
    	run availability status listener
error parsing flags: flag provided but not defined: -test.v=== RUN   TestInternalAuthenticationGet
    internal_handlers_test.go:36: Did not return 200
    internal_handlers_test.go:46: ghosts infected the return
=== RUN   TestInternalAuthenticationGetNotFound
--- PASS: TestInternalAuthenticationGetNotFound (0.00s)
```

focus on `-test.v=== RUN`
it seems that because here is not line indent, the Golang incorrectly shows the test results

same test after change:
image

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -112,7 +112,7 @@ func Get() *SourcesApiConfig {
 
 	err := fs.Parse(os.Args[1:])
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "error parsing flags: %v", err)
+		fmt.Fprintf(os.Stderr, "error parsing flags: %v\n", err)
 	}
 
 	options.SetDefault("StatusListener", *availabilityListener)


MikelAlejoBR (approved):
Nice one! I don't know why very recently the tests have started to complain about the "run" argument though :eyes: 

The good thing is that it doesn't affect the tests, at least.

lpichler (approved):

lpichler:
~~it complains when we give arguments to `go test`  which doesn't have flag definition (like `flag.Bool("integration", false, "run unit or integration tests")`) which not conformable when you would like to verbose tests like:~~

~~`go test ./... --integration -v`~~

~~I am not sure how to solve it yet.~~

PR #173, opened 2022-03-23
Add notification integration

lpichler:
This PR sends Kafka message (according to notification doc.) which represent notification when av. status on any resource was changed. There are two situation when status can happen - when PATCH request is send or when status message is processed in availability listener.

---
This PR adds notification emission when available status(Source, Application, Authentication, Endpoint) is changed, change are present in :
- PATCH requests:
   - updates on handlers
     - get previous status
     - get current status
     - set it to context - so it can be processed by notification middleware
  - `Notifier` middleware which call function for notification emission(`EmitAvailabilityStatusNotification`) from new `notifications` service package 
  - `EmailNotificationInfo` method added to models which covers attributes for email templates (sent via Kafka message)
  - added `notifications.go` in service package which covers sending notification in kafka message form (it is process by notification service)
- Availability Status Listener
  - get previous av. status from resource in status message 
  - change in signature of `FetchAndUpdateBy`  we need to return resource in order to populate  `EmitAvailabilityStatusNotification`  (together with previous av. status)

### Links
PR in notification backend:
https://github.com/RedHatInsights/notifications-backend/pull/1097
https://issues.redhat.com/browse/RHCLOUD-16269


lpichler:
replaced by https://github.com/RedHatInsights/sources-api-go/pull/204

PR #174, opened 2022-03-23
[RHCLOUD-15003] Remove old active record migrations table

MikelAlejoBR:
This PR adds a migration to remove the old ActiveRecord migrations table from the database, since we are not going to use it anymore.

## Links

https://issues.redhat.com/browse/RHCLOUD-15003

### DIFF
--- a/db/migrations/20220705100000_remove_old_migrations_table.go
+++ b/db/migrations/20220705100000_remove_old_migrations_table.go
@@ -1,0 +1,54 @@
+package migrations
+
+import (
+	"time"
+
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// RemoveOldMigrationsTable removes the ActiveRecord migrations table.
+func RemoveOldMigrationsTable() *gormigrate.Migration {
+	type schemaMigrations struct {
+		Version string `gorm:"primaryKey;type:varchar"`
+	}
+
+	type arInternalMetadata struct {
+		Key       string    `gorm:"primaryKey;type:varchar;not null"`
+		Value     string    `gorm:"type:varchar"`
+		CreatedAt time.Time `gorm:"type:timestamp"`
+		UpdatedAt time.Time `gorm:"type:timestamp"`
+	}
+
+	return &gormigrate.Migration{
+		ID: "20220705100000",
+		Migrate: func(db *gorm.DB) error {
+			logging.Log.Info(`Migration "remove old migrations table" started`)
+			defer logging.Log.Info(`Migration "remove old migrations table" ended`)
+
+			err := db.Transaction(func(tx *gorm.DB) error {
+				err := tx.Migrator().DropTable(&schemaMigrations{})
+				if err != nil {
+					return err
+				}
+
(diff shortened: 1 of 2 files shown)


petracihalova (approved):

lindgrenj6:
Adding a label here - `wait-to-merge` since we aren't going to merge these until we un-deploy the rails app.

db/migrations/20220222120300_remove_old_migrations_table.go
```diff
@@ -0,0 +1,54 @@
...
+				if err != nil {
+					return err
+				}
+
+				err = tx.Migrator().DropTable(&arInternalMetadata{})
+				if err != nil {
+					return err
+				}
+
+				return nil
```
lindgrenj6: I did have a thought here - can we add a `IF EXISTS` somehow? It might complain if the tables aren't there (e.g. local database lacks them)

MikelAlejoBR: There is no need for that — the initial migration checks for the existence of that table, and if it doesn't exist, it creates it. So this migration will always find a table to delete, unless you run only this migration in isolation and not along with the rest:

https://github.com/RedHatInsights/sources-api-go/blob/f89e42d9e2faaa0372be149b5804885bd8bb3431/db/migrations/sql/1_initial_schema.sql#L630-L664

lpichler:
this needs rebase - and @lindgrenj6 is it still wait-to-merge ?

lindgrenj6:
All depends if we're ready to drop the old tables. My gut instinct says yes
it's ready to go!
--
Jacob Lindgren

Red Hat

MikelAlejoBR:
All right, I've rebased the PR

lpichler (approved):
LGTM

lpichler (commented):
This PR needs rebase.

lpichler:
@MikelAlejoBR I guess we can remove label wait-to-merge.

MikelAlejoBR:
Sure thing! I've rebased the PR and:

* Make this migration run as the last one.
* Included a small fix which makes sure that the legacy `schema_migrations` table is created for new databases — dev, basically —.

lpichler:
I am sorry @MikelAlejoBR for requesting to rebase this again - I guess It looks that  I forgot to merge it :/

MikelAlejoBR:
No problem! This should be good to go now!

db/migrations/sql/1_initial_schema.sql
```diff
@@ -652,8 +652,8 @@ DO
 
 $$
 BEGIN
-    IF NOT "table_exists"('ar_internal_metadata') THEN
-        CREATE TABLE "schema_migrations" (
+    IF NOT "table_exists"('schema_migrations') THEN
```
lpichler: @MikelAlejoBR Should `1_initial_schema.sql` reflect original state as rails app had ? if so should we create `ar_internal_metadata` here ? (because in migration here we are deleting `ar_internal_metadata `)

MikelAlejoBR: I don't think so, because just above this code block we're checking for the `ar_internal_metadata` table. However I was checking here for `ar_internal_metadata` again instead of `schema_migrations`... Maybe this belongs to another PR

lpichler:
Application works fine but tests are failing :/

PR #176, opened 2022-03-23
Add Proxy of /graphql over to rails application

lindgrenj6:
Pretty simple, just proxying a `POST /graphql` request over to the rails side.

\# TODO:
- [x] initial proxying
- [x] store rails URL in ENV

PR #177, opened 2022-03-23
Skip marshaling OpenAPI JSON when reading from disk/embed

lindgrenj6:
Found out as in #176 we can use `c.JSONBlob(...)` instead of re-marshaling the JSON when we are reading from disk or a response etc. 

Basically we use this function only when we know the bytes are already JSON.

### DIFF
--- a/openapi_handlers.go
+++ b/openapi_handlers.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"embed"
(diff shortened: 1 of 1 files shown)


MikelAlejoBR (approved):
Nice finding! :+1:

lpichler (approved):

PR #180, opened 2022-03-24
[RHCLOUD-18600] feature: categorize source types in different categories

MikelAlejoBR:
Due to the considerable number of source types we currently support, the UI is getting a little bit confusing. The aim of adding a new "category" field is to make it easier for the UI team to better organize how the source types are displayed.

## Links

[[RHCLOUD-18600]](https://issues.redhat.com/browse/RHCLOUD-18600)

### DIFF
--- a/dao/seeding.go
+++ b/dao/seeding.go
@@ -88,6 +88,7 @@ func seedSourceTypes() error {
 		}
 
 		// mark the fields as updated
+		st.Category = values.Category
 		st.ProductName = values.ProductName
 		st.IconUrl = values.IconURL
 		st.Schema = schema
--- a/dao/seeding_types.go
+++ b/dao/seeding_types.go
@@ -9,6 +9,7 @@ type (
 )
 
 type sourceTypeSeed struct {
+	Category    string      `json:"category"`
 	ProductName string      `json:"product_name"`
 	Schema      interface{} `json:"schema"`
 	Vendor      string      `json:"vendor"`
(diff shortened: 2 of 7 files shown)


rvsia (approved):
Looks great to me! Thanks! 🏆 🌮

lindgrenj6:
Sooo there is one little thing here - the `category` field isn't going to be inserted into the DB because there isn't a column for it :disappointed:. 

We can probably make a migration in this PR (after #72) that adds it. Then update the models to support seeding that field. 

---
E: The other option would be to throw it in the schema somehow....but that seems hacky.

lpichler:
> Sooo there is one little thing here - the `category` field isn't going to be inserted into the DB because there isn't a column for it 😞.
> 
> We can probably make a migration in this PR (after #72) that adds it. Then update the models to support seeding that field.
> 
> E: The other option would be to throw it in the schema somehow....but that seems hacky.

@MikelAlejoBR I guess we can add the migration with adding `category` column.

MikelAlejoBR:
I've:

- Created a migration to add a new "Category" column and to update all the existing source types' categories.
- Updated the models to reflect this change.
- Updated the seeding so that it picks up the category.

It includes a couple of commits from https://github.com/RedHatInsights/sources-api-go/pull/228 so that the tests don't fail.

lindgrenj6 (approved):
This one looks good - might just need to be rebased to pick up the fixed version of the endpointdao.exists method.

lpichler (approved):

PR #181, opened 2022-03-24
Loop over all versions when creating routes on the echo router

lindgrenj6:
This is to support `(v1.0|v2.0|v3.0|v3.1)` for Sources. Currently just routing all to the same handler functions but we can split them out eventually if we want true versioning. Versioning on sources is kind of a joke since we share a boatload of behavior for basically all of the endpoints and the true only differences between them are added/removed endpoints.

### DIFF
--- a/openapi_handlers.go
+++ b/openapi_handlers.go
@@ -10,15 +10,13 @@ import (
 //go:embed public
 var publicDocs embed.FS
 
-func PublicOpenApiv31(c echo.Context) error {
-	bytes, err := readOpenApiJson("public/openapi-3-v3.1.json")
-	if err != nil {
-		return err
+func PublicOpenApi(version string) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		bytes, err := publicDocs.ReadFile("public/openapi-3-" + version + ".json")
+		if err != nil {
+			return err
+		}
+
+		return c.JSONBlob(http.StatusOK, bytes)
 	}
-
-	return c.JSONBlob(http.StatusOK, bytes)
-}
-
-func readOpenApiJson(filename string) ([]byte, error) {
-	return publicDocs.ReadFile(filename)
 }
--- a/routes.go
+++ b/routes.go
@@ -20,82 +20,85 @@ func setupRoutes(e *echo.Echo) {
 		return c.String(http.StatusOK, "OK")
 	})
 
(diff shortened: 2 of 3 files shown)


openapi_handlers.go
```diff
@@ -10,15 +10,13 @@ import (
...
-	bytes, err := readOpenApiJson("public/openapi-3-v3.1.json")
-	if err != nil {
-		return err
-	}
-
-	return c.JSONBlob(http.StatusOK, bytes)
-}
+func PublicOpenApi(version string) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		bytes, err := publicDocs.ReadFile("public/openapi-3-" + version + ".json")
```
syncrou, line 15: To throw in more context around this.  I believe that we should support REAL versioning. But for now we need to get the rewrite out.  

So I back this change for now ( as a crawl phase ) and would like to see real versioning handled in the future in follow up PRs.

routes.go
```diff
@@ -20,79 +20,82 @@ func setupRoutes(e *echo.Echo) {
...
-
-	// Red Hat Connector Connections
-	v3.GET("/rhc_connections", RhcConnectionList, tenancyWithListMiddleware...)
-	v3.GET("/rhc_connections/:id", RhcConnectionGetById, permissionMiddleware...)
-	v3.POST("/rhc_connections", RhcConnectionCreate, permissionMiddleware...)
-	v3.PATCH("/rhc_connections/:id", RhcConnectionUpdate, permissionMiddleware...)
-	v3.DELETE("/rhc_connections/:id", RhcConnectionDelete, permissionMiddleware...)
-	v3.GET("/rhc_connections/:id/sources", RhcConnectionSourcesList, permissionWithListMiddleware...)
+	apiVersions := []string{"v1.0", "v2.0", "v3.0", "v3.1"}
+	for i := range apiVersions {
```
MikelAlejoBR: I would like to suggest using a `for _, version` here since we don't use the index of the loop for anything special.

lindgrenj6: Sure, done.

lpichler:
@lindgrenj6 this needs rebase & solve conflict.

petracihalova (approved):

lindgrenj6:
rebased!
--
Jacob Lindgren

Red Hat

lpichler (approved):

PR #182, opened 2022-03-24
Clean up authentications when the subresource is deleted

lindgrenj6:
Now that we've sharded our data - we need to clean up the vault db whenever something is deleted. 

I did this in the dao layer that way it's mostly transparent from the handlers. 

---

NOTE: ~this is only necessary for the vault store. The foreign keys should handle it for authentications~ I'm wrong.

### DIFF
--- a/dao/authentication_dao.go
+++ b/dao/authentication_dao.go
@@ -362,13 +362,44 @@ func (a *authenticationDaoImpl) Delete(uid string) (*m.Authentication, error) {
 	for _, key := range keys {
 		if strings.HasSuffix(key, uid) {
 			path := fmt.Sprintf("secret/metadata/%d/%s", *a.TenantID, key)
-			sec, err := Vault.Delete(path)
+			// fetch the key (for returning)
+			sec, _ := a.getKey(key)
+			// ...then delete it
+			_, err := Vault.Delete(path)
 
-			return authFromVault(sec), err
+			return sec, err
 		}
 	}
 
 	return nil, util.NewErrNotFound("authentication")
+}
+
+// Deletes all authentications for a parent resource. Sources are handled
+// differently than other parent resources mostly due to the cascade effect from
+// a source delete
+func (a *authenticationDaoImpl) Cleanup(resourceType string, resourceID int64) error {
+	resourceType = util.Capitalize(resourceType)
+
+	auths, _, err := a.List(100, 0, []util.Filter{})
+	if err != nil {
+		return err
+	}
+
+	for _, auth := range auths {
+		if resourceType == "Source" && resourceID == auth.SourceID {
+			_, err := a.Delete(auth.ID)
+			if err != nil {
+				logging.Log.Warnf("Failed to delete authentication %v", auth.ID)
+			}
+		} else if auth.ResourceID == resourceID && auth.ResourceType == resourceType {
+			_, err := a.Delete(auth.ID)
+			if err != nil {
+				logging.Log.Warnf("Failed to delete authentication %v", auth.ID)
+			}
+		}
+	}
+
+	return nil
 }
 
 func (a *authenticationDaoImpl) Tenant() *int64 {
--- a/dao/application_dao.go
+++ b/dao/application_dao.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	l "github.com/RedHatInsights/sources-api-go/logger"
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
 	"gorm.io/gorm/clause"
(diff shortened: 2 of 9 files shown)


lindgrenj6:
It looks like it's failing on a test where the mock vault is throwing a NPE. Unable to reproduce this locally. I may work on the pr_check script to spin up a vault container to do our integration tests with.

dao/authentication_dao.go
```diff
@@ -315,15 +315,54 @@ func (a *authenticationDaoImpl) Delete(uid string) (*m.Authentication, error) {
...
+			if auths[i].ResourceID != resourceID || auths[i].ResourceType != resourceType {
+				continue
+			}
+
+			_, err := a.Delete(auths[i].ID)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
MikelAlejoBR: I would suggest using `for _, auth := range auths` to avoid using the `auths[i]` statement inside the loop. I say this because I don't see us doing anything special with the index.

MikelAlejoBR: The problem I see with returning early when trying to cleanup authentications is that we might leave dangling authentications in Vault if one of them fails to be deleted. I'd like to open up a discussion on this one:

I see that we have three options here:

1. Returning early and leaving dangling authentications on Vault. In this case I would suggest modifying the log statement on the `Delete` functions that call the `Cleanup` function to the following: 

```diff
- l.Log.Warnf("error cleaning up authentications: %v", err)
+ l.Log.Errorf(`error cleaning up authentications for application with id "%d": %s`, applicationId, err)
```

Or even:

```diff
- l.Log.Warnf("error cleaning up authentications: %v", err)
+ l.Log.Errorf(`[application_id: %d] error cleaning up authentications: %s`, applicationId, err)
```
Also, I would include a `infof` statement with either of the formats above stating that the resource with id "X" was deleted.

That way whenever we check Vault manually and spot some dangling authentications, we can always go back to the log history to see if the resource was deleted from the database, and if it is safe to delete the dangling authentications.

2. Delete the resource itself but if there is an error with an authentication, store an async job to retry the dangling authentications deletion on a scheduled basis. This has the drawback of having to persist these jobs to avoid losing them if for some reason the application crashes, but we make the process pseudo-automatic, unless the job fails. In that last case manual action would be required from our part.

3. Wrapping up the resource deletion + authentications cleanup in a transaction. This would leave the resource undeleted but in an inconsistent state. The idea here would be to "pause" and disable the resource, let the user know there was a problem deleting the resource, but leave the resource there in the list, only allowing to click on the "delete" button again, so that the whole process triggers and hopefully this time all the auths get deleted.

I guess the second and third options are too much work + too much stuff to test, so the first one seems the most viable one. Although I leave the other two in case new ideas arise.

This feels like a good case to add another event to the integration with the notifications team: whenever a manual action is required, the team receives an email indicating that we should take care of dangling authentications.

dao/authentication_dao.go
```diff
@@ -315,15 +315,54 @@ func (a *authenticationDaoImpl) Delete(uid string) (*m.Authentication, error) {
...
+	resourceType = util.Capitalize(resourceType)
+
+	auths, _, err := a.List(100, 0, []util.Filter{})
+	if err != nil {
+		return err
+	}
+
+	for i := range auths {
+		if resourceType == "Source" {
+			if auths[i].SourceID != resourceID {
```
lpichler: What do you think to rewrite those condition to positive way ? 
I guess it is more understandable, like:

```go
		if resourceType == "Source" && auths[i].SourceID == resourceID {
...
```

dao/authentication_dao.go
```diff
@@ -315,15 +315,54 @@ func (a *authenticationDaoImpl) Delete(uid string) (*m.Authentication, error) {
...
+				continue
+			}
+
+			_, err := a.Delete(auths[i].ID)
+			if err != nil {
+				return err
+			}
+		} else {
+			if auths[i].ResourceID != resourceID || auths[i].ResourceType != resourceType {
```
lpichler: maybe 

```go
... 
}else if auths[i].ResourceID == resourceID && auths[i].ResourceType == resourceType {
...
```

lpichler: also it would be possible to merge both branches, like:
```go
resourceCondition := auths[i].ResourceID == resourceID && auths[i].ResourceType == resourceType
sourceCondition  := resourceType == "Source" && auths[i].SourceID == resourceID

if resourceCondition || sourceCondition {
...
```
names are just suggestion but maybe It could improve readability.

dao/authentication_dao.go
lindgrenj6: I just added a warn log there when it fails. We can potentially add metrics around vault failures in the future once we get the vault backend. 

---

I also added an implementation for the db dao since the cascade on deleting a source will still leave authentications dangling. The DB DAO is much easier since its just one query.

dao/authentication_dao.go
lindgrenj6: I reworked them to be more readable. I also added a DB implementation since we do not have a vault backend yet.

lindgrenj6:
I also realized we aren't raising `Type.destroy` events properly for subresources - e.g. if we delete a source we only raise `Source.destroy` rather than the source + all subresourcess. 

Going to add some code that basically destroys all resources (with the delete functions returning what was destroyed) then raise all of the events.

MikelAlejoBR (approved):

PR #183, opened 2022-03-24
Add miq-password Compatible encryption/decryption of strings

lindgrenj6:
Relates to #127 

These are the helper functions we can use when someone has a password in their payload when using the database backed DAO. 

Basically when validating an `AuthenticationCreate` struct we will see if the password is present - and if it is encrypt it before shipping it off to the DB. The likewise will be true too - in the `Authentication#ToInternalResponse` we will return the decrypted version of the password.

### DIFF
--- a/util/encryption.go
+++ b/util/encryption.go
@@ -1,0 +1,132 @@
+package util
+
+import (
+	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
+	"encoding/base64"
+	"fmt"
+	"os"
+	"regexp"
+	"strings"
+)
+
+var (
+	// "the key" to encrypt/decrypt passwords with
+	key        = os.Getenv("ENCRYPTION_KEY")
+	keyPresent = false
+
+	// regex to pull the actual secret out of the miq-password style column
+	xtractRe = regexp.MustCompile(`v2:{(.+)}`)
+)
+
+// init for this file basically just decodes the ENCRYPTION_KEY env var to the
+// plain text equivalent
+func init() {
+	if key == "" {
+		return
+	}
+
+	// the key is base64 encoded in the ENV
+	decoded, err := base64.RawStdEncoding.DecodeString(string(key))
+	if err != nil {
+		panic(err)
+	}
+	key = string(decoded)
+	keyPresent = true
+}
+
+// Encrypts str into a miq-compatible password format using the encryption key
+// in the environment
+func Encrypt(str string) (string, error) {
+	if !keyPresent {
+		return "", fmt.Errorf("no encryption key present")
+	}
+
+	encoded, err := encode(str)
+	if err != nil {
+		return "", err
+	}
+
+	// base64 encode the encrypted secret for text-storage
+	return "v2:{" + base64.RawStdEncoding.EncodeToString([]byte(encoded)) + "}", nil
+}
+
+// Decrypts a miq-compatible password into a string
+func Decrypt(str string) (string, error) {
+	if !keyPresent {
+		return "", fmt.Errorf("no encryption key present")
+	}
+
+	if !xtractRe.Match([]byte(str)) {
+		return "", fmt.Errorf("string does not match format")
+	}
+
+	// extract the password from between the brackets
+	matches := xtractRe.FindAllStringSubmatch(str, 1)
+
+	// the password is base64 encoded
+	rawPass, err := base64.RawStdEncoding.DecodeString(matches[0][1])
+	if err != nil {
+		return "", err
+	}
+
+	return decode(string(rawPass))
+}
+
+func decode(pw string) (string, error) {
+	// create the block from the key
+	block, err := aes.NewCipher([]byte(key))
+	if err != nil {
+		return "", err
+	}
+
+	// no iv used, but need an iv that is the same length as the block size.
+	cbc256 := cipher.NewCBCDecrypter(block, make([]byte, block.BlockSize()))
+
+	// output size must be the same as input. using the same bytes for
+	// input/output is fine but can lead to trailing spaces which is why we trim
+	// it before returning
+	output := make([]byte, len(pw))
+	cbc256.CryptBlocks(output, []byte(pw))
+
+	// strip out the trailing zeros that come from the extra length due to the
+	// block size
+	return strings.Trim(string(output), "\x00"), nil
+}
+
+func encode(pw string) (string, error) {
+	// create the block from the key
+	block, err := aes.NewCipher([]byte(key))
+	if err != nil {
+		return "", err
+	}
+
+	// no iv used, but need an iv that is the same length as the block size.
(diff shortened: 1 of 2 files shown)


util/encryption.go
```diff
@@ -0,0 +1,132 @@
...
+	"strings"
+)
+
+var (
+	// "the key" to encrypt/decrypt passwords with
+	key        = os.Getenv("ENCRYPTION_KEY")
+	keyPresent = false
+
+	// regex to pull the actual secret out of the miq-password style column
+	xtractRe = regexp.MustCompile(`v2:{(.*)}`)
```
MikelAlejoBR: The following string would be valid with the defined regex: `v2:{()}`. Maybe that `.*` needs to be `.+`?

lindgrenj6: Well the `()` are acting as capture groups - but the point stands that `v2{}` would have matched. I added a test for it in the TestBadFormat test and changed it to `.+` to make it safer.

MikelAlejoBR: Oops, yeah, I forgot about the capturing groups :sweat_smile:

util/encryption.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func decode(pw string) (string, error) {
+	// create the block from the key
+	block, err := aes.NewCipher([]byte(key))
+	if err != nil {
+		return "", err
+	}
+
+	// no iv used, but need an iv that is the same length as the block size.
```
petracihalova, line 84: Hi 😁  please what does "iv" mean?

petracihalova (approved):
LGTM 👍🏻

util/encryption.go
lindgrenj6: Without going into _too_ much details the IV is kind of like a salt on the string for encryption.

The way it **differs** from a salt is this:
- IV keys should be unique per record (e.g. randomly generated, then stored alongside for decryption)
- they _aren't_ secret, it is safe to store them in plaintext right alongside the cipher.

To decrypt with an IV you just provide it sort of as a second key. In the miq-password gem it looks like they don't use IV's which is kind of a no-no, we're not using that extra layer of security. I'm sort-of glad we can move away from it eventually through vault or we could eventually change out our implementation to use an IV per-password.

More in-depth explanation here: https://stackoverflow.com/a/39412887

petracihalova: thanks

util/encryption.go
```diff
@@ -0,0 +1,132 @@
...
+// helper function to create a string that is the "proper" length to be
+// encrypted. the right block size is basically the length of the string +
+// however many bytes it takes to get up to a multiple of the blocksize
+//
+// in mathy terms: length = len(password) + (len(password) % blocks)
+//
+// e.g. a string length 4 with a blocksize of 8 would return the string with
+// four spaces at the end.
+func padString(text string, blockSize int) string {
+	padLength := blockSize - len(text)%blockSize
```
lpichler, line 137: I guess we need guard condition that `blockSize` has to be positive (+without zero)

util/encryption.go
```diff
@@ -0,0 +1,132 @@
...
+	padding := bytes.Repeat([]byte{byte(0)}, padLength)
```
lpichler, line 138: maybe we can `return text if padLength == len(text)`

util/encryption.go
```diff
@@ -0,0 +1,132 @@
...
+// init for this file basically just decodes the ENCRYPTION_KEY env var to the
+// plain text equivalent
+func init() {
+	if key == "" {
+		return
+	}
+
+	// the key is base64 encoded in the ENV
+	decoded, err := base64.RawStdEncoding.DecodeString(string(key))
+	if err != nil {
```
lpichler, line 32: it looks that it is not necessary to convert type of `key` here.

lpichler, line 32: `RawStdEncoding` omits padding characters and `StdEncoding` doesn't.
What about to cover both - padded and unpadded encodings ? 

```go
	
decoded, err := base64.StdEncoding.DecodeString("MTIzNDU2Nzg=")
fmt.Printf("Decode(MTIzNDU2Nzg=): StdEncoding: %v error: %v \n", string(decoded), err)

decodedA, err := base64.StdEncoding.DecodeString("MTIzNDU2Nzg")
fmt.Printf("Decode(MTIzNDU2Nzg): StdEncoding: %v error: %v \n", string(decodedA), err)

decodedRaw, err := base64.RawStdEncoding.DecodeString("MTIzNDU2Nzg=")
fmt.Printf("Decode(MTIzNDU2Nzg=): RawStdEncoding: %v error: %v \n", string(decodedRaw), err)

decodedRawA, err := base64.RawStdEncoding.DecodeString("MTIzNDU2Nzg")
fmt.Printf("Decode(MTIzNDU2Nzg): RawStdEncoding: %v error: %v \n", string(decodedRawA), err)

encoded := base64.StdEncoding.EncodeToString([]byte("12345678"))
fmt.Printf("StdEncoding(12345678): %v \n", encoded)
encodedRaw := base64.RawStdEncoding.EncodeToString([]byte("12345678"))
fmt.Printf("RawStdEncoding(12345678): %v \n", encodedRaw)
```

```
Decode(MTIzNDU2Nzg=): StdEncoding: 12345678 error: <nil> 
Decode(MTIzNDU2Nzg): StdEncoding: 123456 error: illegal base64 data at input byte 8 
Decode(MTIzNDU2Nzg=): RawStdEncoding: 123456 error: illegal base64 data at input byte 11 
Decode(MTIzNDU2Nzg): RawStdEncoding: 12345678 error: <nil> 
StdEncoding(12345678): MTIzNDU2Nzg= 
RawStdEncoding(12345678): MTIzNDU2Nzg 
```

lpichler:
Question: How is value for env variable `ENCRYPTION_KEY` generated in stage/prod envs ? Thanks!

util/encryption.go
lindgrenj6: leftover from my POC where the key was `[]byte`, whoops.

lindgrenj6:
> Question: How is value for env variable `ENCRYPTION_KEY` generated in stage/prod envs ? Thanks!

It comes from the appSRE vault. We have a secret in there that gets persisted in our namespaces named `sources-api-secrets` and `ENCRYPTION_KEY` gets mounted from there.

util/encryption.go
lindgrenj6: I added a panic test for it - since that should definitely fail as one is testing locally.

util/encryption.go
lindgrenj6: Done - working on a test for that yet though. Since strings are intern'd we can't really tell if it's the same string or not.

util/encryption.go
lindgrenj6: As I put in DM - main reasoning was that in miq-password they use strict encoding so I wanted to be backwards compatible by only supporting that way.

lpichler (approved):
LGTM 👍 thanks!

PR #184, opened 2022-03-27
Add support for subresource filtering operations

lindgrenj6:
This adds support for filtering sources or applications based on their types. This is used by GraphQL as well as some random requests that come from consoledot. 

example syntax:

```
GET /api/sources/v3.1/sources?filter[source_type][name]=amazon
```
would list the sources that have the amazon source type.

---

The biggest change was in the parsing - rather than using a fancy regex to try and parse whats between the brackets we just match the whitespace to ignore them and build out the struct appropriately. Comments explain what the input/output looks like.

### DIFF
--- a/dao/filtering.go
+++ b/dao/filtering.go
@@ -15,46 +15,59 @@ func applyFilters(query *gorm.DB, filters []util.Filter) (*gorm.DB, error) {
 			return nil, fmt.Errorf("failed to parse statement: %v", err)
 		}
 	}
+
 	var filterName string
 	for _, filter := range filters {
-		if query.Statement.Table != "" {
+		// subresource filtering!
+		if filter.Subresource != "" {
+			switch filter.Subresource {
+			case "source_type":
+				query = query.Joins("SourceType")
+				filterName = fmt.Sprintf("%v.%v", `"SourceType"`, filter.Name)
+			case "application_type":
+				query = query.Joins("ApplicationType")
+				filterName = fmt.Sprintf("%v.%v", `"ApplicationType"`, filter.Name)
+			default:
+				return nil, fmt.Errorf("invalid subresource type [%v]", filter.Subresource)
+			}
+		} else if query.Statement.Table != "" {
 			filterName = fmt.Sprintf("%v.%v", query.Statement.Table, filter.Name)
 		} else {
 			filterName = filter.Name
 		}
 
 		switch filter.Operation {
-		case "", "[eq]":
+		case "", "eq":
 			query = query.Where(fmt.Sprintf("%v = ?", filterName), filter.Value[0])
-		case "[not_eq]":
+		case "not_eq":
 			query = query.Where(fmt.Sprintf("%v != ?", filterName), filter.Value[0])
-		case "[gt]":
+		case "gt":
 			query = query.Where(fmt.Sprintf("%v > ?", filterName), filter.Value[0])
-		case "[gte]":
+		case "gte":
 			query = query.Where(fmt.Sprintf("%v >= ?", filterName), filter.Value[0])
-		case "[lt]":
+		case "lt":
 			query = query.Where(fmt.Sprintf("%v < ?", filterName), filter.Value[0])
-		case "[lte]":
+		case "lte":
 			query = query.Where(fmt.Sprintf("%v <= ?", filterName), filter.Value[0])
-		case "[nil]":
+		case "nil":
 			query = query.Where(fmt.Sprintf("%v IS NULL", filterName))
-		case "[not_nil]":
+		case "not_nil":
 			query = query.Where(fmt.Sprintf("%v IS NOT NULL", filterName))
-		case "[contains]":
+		case "contains":
 			query = query.Where(fmt.Sprintf("%v LIKE ?", filterName), fmt.Sprintf("%%%s%%", filter.Value[0]))
-		case "[starts_with]":
+		case "starts_with":
 			query = query.Where(fmt.Sprintf("%v LIKE ?", filterName), fmt.Sprintf("%s%%", filter.Value[0]))
-		case "[ends_with]":
+		case "ends_with":
 			query = query.Where(fmt.Sprintf("%v LIKE ?", filterName), fmt.Sprintf("%%%s", filter.Value[0]))
-		case "[eq_i]":
+		case "eq_i":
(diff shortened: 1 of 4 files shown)


MikelAlejoBR (approved):

petracihalova (approved):
LGTM 😁

lindgrenj6:
@lpichler when you get a minute - I could use this one for my graphQL implementation. I split it out to make it easier to review.

dao/filtering.go
```diff
@@ -15,46 +15,59 @@ func applyFilters(query *gorm.DB, filters []util.Filter) (*gorm.DB, error) {
...
 	for _, filter := range filters {
-		if query.Statement.Table != "" {
+		// subresource filtering!
+		if filter.Subresource != "" {
+			switch filter.Subresource {
+			case "source_type":
+				query = query.Joins("SourceType")
+				filterName = fmt.Sprintf("%v.%v", `"SourceType"`, filter.Name)
+			case "application_type":
+				query = query.Joins("ApplicationType")
```
lpichler, line 36: it looks that it doesn't work when primary collection is source:
```
http://localhost:{{port}}/api/sources/v3.1/sources?filter[application_type][name]=%2Finsights%2Fplatform%2Fcost-management
```
as there is joining table `applications` between `sources` and `application types`

what come to my mind you can use `NewRelationObject` thing used for subcollections: 

```diff
diff --git a/dao/filtering.go b/dao/filtering.go
index 2d071df..a1a41d5 100644
--- a/dao/filtering.go
+++ b/dao/filtering.go
@@ -2,6 +2,7 @@ package dao

 import (
        "fmt"
+       m "github.com/RedHatInsights/sources-api-go/model"
        "strings"

        "github.com/RedHatInsights/sources-api-go/util"
@@ -25,7 +26,15 @@ func applyFilters(query *gorm.DB, filters []util.Filter) (*gorm.DB, error) {
                                query = query.Joins("SourceType")
                                filterName = fmt.Sprintf("%v.%v", `"SourceType"`, filter.Name)
                        case "application_type":
-                               query = query.Joins("ApplicationType")
+                               //query = query.Joins("ApplicationType")
+
+                               relationObject, err := m.NewRelationObject(m.ApplicationType{Id: 3}, 1, DB.Debug())
+                               if err != nil {
+                                       return nil, util.NewErrNotFound("source")
+                               }
+
+                               query = relationObject.HasMany(&m.Source{}, DB.Debug())
+
                                filterName = fmt.Sprintf("%v.%v", `"ApplicationType"`, filter.Name)
                        default:
                                return nil, fmt.Errorf("invalid subresource type [%v]", filter.Subresource)
@@ -38,7 +47,7 @@ func applyFilters(query *gorm.DB, filters []util.Filter) (*gorm.DB, error) {

                switch filter.Operation {
                case "", "eq":
-                       query = query.Where(fmt.Sprintf("%v = ?", filterName), filter.Value[0])
+                       //query = query.Where(fmt.Sprintf("%v = ?", filterName), filter.Value[0])
                case "not_eq":
                        query = query.Where(fmt.Sprintf("%v != ?", filterName), filter.Value[0])
                case "gt":

```

but you need to get ID of application type from in this case name
or change NewRelationObject to work also with `Name`

it is just suggestion maybe it would be doable to write separate query but `New Relation Object` should already cover it.

lindgrenj6: Ah, I should have specified more: the subresources really should only be used for their respective types. Like:

- Source -> filter on SourceType fields
- Application -> filter on ApplicationType fields

like the only URI that I know that does this (minus graphql) is this request from the main platform experience UI: `/api/sources/v3.1/applications?filter%5Bapplication_type%5D%5Bname%5D%5Bcontains%5D=catalog` where they're filtering applications on the application type name. 

---

So it should be ok if we can't filter a Source on the applications connected to it.

lindgrenj6: So we don't run into this later I went and added a check to make sure that the _only_ subresource queries allowed are 
- sources -> source_types 
- applications -> application_types 

that way the user would get a bad request if they ever tried it. Thanks for catching that!

lpichler: thanks @lindgrenj6 !

lpichler (approved):
LGTM 👍

lpichler:
I realised last thing :
```
/sources?filter[application_type][name]=%2Finsights%2Fplatform%2Fcost-management

/sources?filter[<sub_resource>][<column>]=..
```

`<column>` - Should be only `name` column or should we able use other columns - currently it looks that it works only for "string" columns.

Other thing -  you can pass any word(by the regex) into sql query.

``` 
/sources?filter[source_type][whatever]=222
```

```

{
  "errors": [
    {
      "detail": "bad request: ERROR: column SourceType.whatever does not exist (SQLSTATE 42703)",
      "status": "400"
    }
  ]
}
```

executed SQL:

```sql
SELECT sources.id, sources.availability_status, Tenant.external_tenant,Tenant.id AS Tenant__id,Tenant.external_tenant AS Tenant__external_tenant,Tenant.created_at AS Tenant__created_at,Tenant.updated_at AS Tenant__updated_at,SourceType.id AS SourceType__id,SourceType.created_at AS SourceType__created_at,SourceType.updated_at AS SourceType__updated_at,SourceType.name AS SourceType__name,SourceType.product_name AS SourceType__product_name,SourceType.vendor AS SourceType__vendor,SourceType.icon_url AS SourceType__icon_url,SourceType.schema AS SourceType__schema 
FROM sources 
LEFT JOIN tenants Tenant ON sources.tenant_id = Tenant.id 
LEFT JOIN source_types SourceType ON sources.source_type_id = SourceType.id 
WHERE SourceType.whatever = '222' LIMIT 100
```

Maybe we should check which columns are allowed to use - I guess users should not be allowed to pick directly column from table. 

Let me know if you think that it should be solved otherwise I can merge ✅

lindgrenj6:
> Should be only name column or should we able use other columns - currently it looks that it works only for "string" columns.

I _think_ the rails app coerces strings to ints and back and forth, but I also do know that the only thing people usually filter on is the name. I agree maybe we should lock it down in a follow-up PR since that was also there before from what I can tell.

> Other thing - you can pass any word(by the regex) into sql query.

Yeah - that's expected per the current implementation and it would be a bad request (400) like it returns. If we locked it down maybe we could catch it in the middleware rather than letting the bad request get all the way to the db.

PR #185, opened 2022-03-27
Implement Actual GraphQL

lindgrenj6:
# Look below for a comment about the PR!

Started playing with gqlgen on friday 3/25 for fun - was able to get graphQL mostly working without too many hitches. 

\# TODO:
- [x] basic graphql functionality fetching fields/subresources (thanks for the help in finding what you need @rvsia)
- [x] limit/offset support
- [x] sort_by support
- [x] filtering
- [x] metadata (counts, etc)
- [x] use a mutex  per request to prevent N+1 queries

--- 

EDIT: I may take time to impelment the caching/mutex similar to what I did for Applications/Endpoints. But it can be reviewed in its current state.


lindgrenj6:
This is ready for review! I was able to get it all working. There are some complex things going on in some of the resolvers so I will try and lay them out here. 

---

For the backend:

- the config is just in `./gqlgen.yml`, it controls how the tool generates code based on our schema
- the graphQL schema lives in `graphs/schema.graphqls`, that is where the response structures are defined as well as any input objects
- the generated code lives in `graphs/generated/` and anything in there we **do not have to touch**, notably its the schema code + the generated models for the responses
- if the library cannot figure out how to expose a field (such as an int64 ID that we want returned as a string) it will create a function in `graph/schema.resolvers.go` which will need implemented to return the specified data. That is where most of the complex logic regarding querying subresources lives and converting between types.
- to prevent N+1 queries for applications and endpoints I have implemented a mutex/map system in the `RequestData` struct. That struct holds the "cache" (not really a cache, just all the current user's objects that are requested) of records which the resolvers only populate once (locking via a mutex) then reading from it for the rest of the request
- I forced implementations of subcollection resolvers so that they only get pulled when the user requests them
- #190 will be needed to make authentications return that do _not_ have applicationAuthentication join records. noticed that when testing. 

For the frontend:

- The sort_by and filter syntax has changed a bit, you can see in the schema what that looks like
- I have code in the main source resolver that parses them into regular `util.Filter` struct instances so our DAOs just handle them all the same
- the extra column comes through when requested (:exclamation:)
- instead of `sources_aggregate` I just created a `meta` object that contains a `count` field which resolves to the count of available sources per the query (if using pagination)


---

Here is a sample request if you want to play with it, this is similar to what the UI runs for the main page:
<details>
<summary>Details</summary>
{
	meta {
		count
	}
	sources(
		offset: 0
		limit: 100
		sort_by: { name: "created_at", direction: desc }
		filter: {
			name: "source_type.vendor"
			operation: "not_eq"
			value: "Red Hat"
		}
	) {
		id
		created_at
		app_creation_workflow
		source_type_id
		name
		imported
		availability_status
		source_ref
		last_checked_at
		updated_at
		last_available_at
		app_creation_workflow
		paused_at
		authentications {
			authtype
			username
		}
		applications {
			application_type_id
			id
			availability_status_error
			availability_status
			paused_at
			extra
			authentications {
				username
				authtype
			}
		}
		endpoints {
			id
			scheme
			host
			port
			path
			receptor_node
			role
			certificate_authority
			verify_ssl
			availability_status_error
			availability_status
			authentications {
				authtype
				availability_status
				availability_status_error
			}
		}
	}
}
</details>

graph/schema.graphqls
```diff
@@ -0,0 +1,110 @@
...
+}
+
+# validation on the direction for SortBy
+enum Direction {
+  asc, desc
+}
+
+# SortBy Object with just a field (which is required) and a direction asc/desc
+input SortBy{
+  field: String!
```
rvsia: Could we also use “name” here to be consistent with the filter?

MikelAlejoBR (approved):

graph/schema.graphqls
lindgrenj6: Absolutely, no problem.

graph/schema.graphqls
```diff
@@ -0,0 +1,110 @@
...
+
+# SortBy Object with just a field (which is required) and a direction asc/desc
+input SortBy{
+  name: String!
+  direction: Direction
+}
+
+# Base Query Object, which returns the array of sources with metadata
+type Query {
+  sources(
```
lpichler, line 35: Will we have just sources here ?

lpichler (approved):
I did some tests and review and this is looking and working great!

I like that graphql query is now simpler!

graph/schema.graphqls
lindgrenj6: Yes I was thinking of just having it be `sources` only at first and if we need to expand it later we can (e.g. someone wants to use graphQL to return other resource types). 

It's as easy as adding them below and regenerating and then implementing the resolvers. 

---

The main reason I didn't do them at first is there is not much of a usecase for fetching the subresources without also fetching the source -> so it makes sense to only expose sources.

PR #186, opened 2022-03-28
Add created_at to Authentication Model storage

lindgrenj6:
1. ~Returning the actual error if the relationship object fails to be created - rather than it bubbling up to not found~ this broke tests, don't feel like refactoring it. 
2. Setting created_at everywhere in the authDAO
3. ~Added support for pointers in the relationship object since it was bubbling up errors like the first point and was very confusing that pointers didn't work.~


petracihalova (approved):
LGTM 😁

MikelAlejoBR (approved):

lpichler (approved):

PR #175, opened 2022-03-23
[RHCLOUD-15003] Rename indexes and foreign keys

MikelAlejoBR:
Adds a migration to rename all the keys and foreign keys to mimic the default names that PosgreSql uses when it creates new indexes or foreign keys and no custom name is provided for them.

**NOTE:** https://github.com/RedHatInsights/sources-api-go/pull/174 needs to be merged first, and then, this PR needs to be rebased.

## Links

https://issues.redhat.com/browse/RHCLOUD-15003

PR #188, opened 2022-03-30
[RHCLOUD-18651] EAN to OrgId migration

MikelAlejoBR:
Modifies the tenancy table by adding the "org_id" column, and attempting a non-hard-failing translation from current EANs to OrgIds.

**NOTE:** https://github.com/RedHatInsights/sources-api-go/pull/175 needs to be merged first, and then, this PR needs to be rebased.

## Links

[[RHCLOUD-18651]](https://issues.redhat.com/browse/RHCLOUD-18651)

### DIFF
--- a/db/migrations/20220330120000_org_id_support.go
+++ b/db/migrations/20220330120000_org_id_support.go
@@ -89,8 +89,7 @@ func OrgIdSupport() *gormigrate.Migration {
 					Model(&Tenant{}).
 					Where("external_tenant = ?", result.EAN).
 					Updates(map[string]interface{}{
-						"external_tenant": *result.EAN,
-						"org_id":          result.OrgID,
+						"org_id": result.OrgID,
 					})
 
 				if dbResult.RowsAffected == 0 {
--- a/db/migrations/20220330120000_org_id_support.go
+++ b/db/migrations/20220330120000_org_id_support.go
@@ -81,6 +81,7 @@ func OrgIdSupport() *gormigrate.Migration {
 			for _, result := range results {
 				if result.Err != nil {
 					logging.Log.Errorf(`[external_tenant: %s] could not translate to "org_id": %s`, *result.EAN, err)
+					continue
 				}
 
 				dbResult := db.
--- a/db/migrations/20220330120000_org_id_support.go
+++ b/db/migrations/20220330120000_org_id_support.go
@@ -1,0 +1,175 @@
+package migrations
+
+import (
+	"context"
+	_ "embed"
+
+	"github.com/RedHatInsights/sources-api-go/config"
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/RedHatInsights/tenant-utils/pkg/tenantid"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// OrgIdSupport creates a new "org_id" column with an index in the "tenants" table and attempts to translate the
+// current EBS account numbers to "org_id"s.
+//
+// If the translation is not possible, it doesn't fail the migration: it simply logs an error. This is due to the
+// transition stage we are in. Since we are proactively taking steps towards the full "OrgId" support, not having a
+// translation for an EBS account number is acceptable.
+//
+// It also adds a comment to the "external_tenant" column, clarifying that it refers to "EBS account numbers".
+func OrgIdSupport() *gormigrate.Migration {
+	type Tenant struct {
+		OrgId string `gorm:"index; comment:Tenant identifier"`
+	}
+
+	return &gormigrate.Migration{
+		ID: "20220330120000",
+		Migrate: func(db *gorm.DB) error {
+			logging.Log.Info(`Migration "20220330120000" started`)
+			defer logging.Log.Info(`Migration "20220330120000" ended`)
+
+			// Perform the table migration.
+			err := db.Transaction(func(tx *gorm.DB) error {
+				// Add the comment to the "external_tenant" column manually, since doing it in the struct with a "gorm"
+				// tag makes "gormigrate" try to create the already existing column.
+				err := tx.
+					Debug().
+					Exec(`COMMENT ON COLUMN "tenants"."external_tenant" IS 'EBS account number'`).
+					Error
+
+				if err != nil {
+					return err
+				}
+
+				err = tx.
+					Debug().
+					AutoMigrate(&Tenant{})
+
+				if err != nil {
+					return err
+				}
+
+				return nil
+			})
+
+			if err != nil {
+				return err
+			}
+
+			// Get all the EBS account numbers from the database.
+			var ebsAccountNumbers []string
+			err = db.
+				Debug().
+				Model(&Tenant{}).
+				Where("external_tenant IS NOT NULL").
+				Pluck("external_tenant", &ebsAccountNumbers).
+				Error
+
+			if err != nil {
+				return err
+			}
+
+			// Attempt to translate the EANs.
+			translator := tenantid.NewTranslator(config.Get().TenantTranslatorUrl)
+			results, err := translator.EANsToOrgIDs(context.Background(), ebsAccountNumbers)
+			if err != nil {
+				return err
+			}
+
+			for _, result := range results {
+				if result.Err != nil {
+					logging.Log.Errorf(`[external_tenant: %s] could not translate to "org_id": %s`, *result.EAN, err)
+				}
+
+				dbResult := db.
+					Debug().
+					Model(&Tenant{}).
+					Where("external_tenant = ?", result.EAN).
+					Updates(map[string]interface{}{
+						"external_tenant": nil,
+						"org_id":          result.OrgID,
+					})
+
+				if dbResult.RowsAffected == 0 {
+					logging.Log.Errorf(`[external_tenant: %s] could not translate to "org_id", external tenant not found`, *result.EAN)
+				}
+
+				if err != nil {
+					logging.Log.Errorf(`[external_tenant: %s][org_id: %s] could no translate to "org_id": %s`, *result.EAN, result.OrgID, err)
+				}
+			}
+
+			return nil
+		},
+		Rollback: func(db *gorm.DB) error {
+			// Get all the EBS account numbers from the database.
+			var orgIds []string
+			err := db.
+				Debug().
+				Model(&Tenant{}).
+				Where("org_id IS NOT NULL").
+				Pluck("org_id", &orgIds).
+				Error
+
+			if err != nil {
+				return err
+			}
+
+			// Revert all OrgIds to EANs.
+			translator := tenantid.NewTranslator(config.Get().TenantTranslatorUrl)
+			results, err := translator.OrgIDsToEANs(context.Background(), orgIds)
+			if err != nil {
+				return err
+			}
+
+			// Revert the "OrgId"s back to "EAN"s.
+			for _, result := range results {
+				if result.Err != nil {
+					logging.Log.Errorf(`[org_id: %s] could not translate to "EAN": %s`, result.OrgID, err)
+				}
+
+				dbResult := db.
+					Debug().
+					Model(&Tenant{}).
+					Where("org_id = ?", result.OrgID).
+					Updates(map[string]interface{}{
+						"external_tenant": *result.EAN,
+						"org_id":          nil,
+					})
+
+				if dbResult.RowsAffected == 0 {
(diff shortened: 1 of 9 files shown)


db/migrations/20220330120000_org_id_support.go
```diff
@@ -0,0 +1,175 @@
...
+func OrgIdSupport() *gormigrate.Migration {
+	type Tenant struct {
+		OrgId string `gorm:"index; comment:Tenant identifier"`
+	}
+
+	return &gormigrate.Migration{
+		ID: "20220330120000",
+		Migrate: func(db *gorm.DB) error {
+			logging.Log.Info(`Migration "20220330120000" started`)
+			defer logging.Log.Info(`Migration "20220330120000" ended`)
```
lindgrenj6: Can we give this a "more pretty" name? Something like `Migration of external tenants to org_id's started|ended`

db/migrations/20220330120000_org_id_support.go
```diff
@@ -0,0 +1,175 @@
...
+				}
+
+				dbResult := db.
+					Debug().
+					Model(&Tenant{}).
+					Where("external_tenant = ?", result.EAN).
+					Updates(map[string]interface{}{
+						"external_tenant": nil,
+						"org_id":          result.OrgID,
+					})
```
lindgrenj6: So we **definitely** do not want to set `external_tenant` to nil during this migration. We are going to need to keep it around until none of our clients need it, we'll just drop the entire column at a future date.

db/migrations/20220330120000_org_id_support.go
```diff
@@ -0,0 +1,175 @@
...
+			// Attempt to translate the EANs.
+			translator := tenantid.NewTranslator(config.Get().TenantTranslatorHost)
+			results, err := translator.EANsToOrgIDs(context.Background(), ebsAccountNumbers)
+			if err != nil {
+				return err
+			}
+
+			for _, result := range results {
+				if result.Err != nil {
+					logging.Log.Errorf(`[external_tenant: %s] could not translate to "org_id": %s`, *result.EAN, err)
```
lindgrenj6: Maybe need a `continue` after the log.

db/migrations/20220330120000_org_id_support.go
```diff
@@ -0,0 +1,175 @@
...
+				}
+
+				if err != nil {
+					logging.Log.Errorf(`[external_tenant: %s][org_id: %s] could no translate to "org_id": %s`, *result.EAN, result.OrgID, err)
+				}
+			}
+
+			return nil
+		},
+		Rollback: func(db *gorm.DB) error {
```
lindgrenj6: I think the rollback for this function can just be using the migrator and dropping the column since we're going to leave the `external_tenant` field as is.

lindgrenj6 (commented):
Good first pass - I think this is basically all we'll need for the migration piece.

db/migrations/20220330120000_org_id_support.go
MikelAlejoBR: Done

db/migrations/20220330120000_org_id_support.go
MikelAlejoBR: I thought we would convert EBS and OrgIds back and forth on the fly, in the case we received an EBS account number for a tenant that does have a corresponding OrgId. However it is true that your suggestion adds that extra safety in the case that the translation service is down.

Done!

db/migrations/20220330120000_org_id_support.go
MikelAlejoBR: Done!

db/migrations/20220330120000_org_id_support.go
MikelAlejoBR: Great catch :tropical_fish: , done!

deploy/clowdapp.yaml
```diff
@@ -200,3 +202,7 @@ parameters:
   name: SOURCES_ENV
   required: true
   value: stage
+- description: Host for the EAN to OrgId translator
+  name: TENANT_TRANSLATOR_HOST
+  required: true
+  value: '?????????'
```
MikelAlejoBR: Do you know where I can find the tenant translator host for this one? I tried finding it on the documents to give the variable a sane default but... :sweat:

lindgrenj6:
<table>
<tr>
	<td>Eph:
	<td>apicast.3scale-dev.svc.cluster.local:8892
<tr>
	<td>Stage:
	<td> apicast.3scale-stage.svc.cluster.local:8892  
<tr>
	<td>Prod:
	<td> apicast.3scale-prod.svc.cluster.local:8893  
</table>

We'll also need to make sure we can hit these from our namespaces in OCP. I'll do that today.

---

@MikelAlejoBR the default should be the eph env one there.

MikelAlejoBR:
> apicast.3scale-dev.svc.cluster.local:8892

Thank you for this!

db/migrations/20220330120000_org_id_support.go
```diff
@@ -0,0 +1,131 @@
...
+
+			if err != nil {
+				return err
+			}
+
+			// Attempt to translate the EANs.
+			translator := tenantid.NewTranslator(config.Get().TenantTranslatorHost)
+			results, err := translator.EANsToOrgIDs(context.Background(), ebsAccountNumbers)
+			if err != nil {
+				return err
```
lpichler: What will follow when connection to translator service will fail ? 
In which state will be pod - do we need to do anything in this case ?

db/migrations/20220330120000_org_id_support.go
```diff
@@ -0,0 +1,131 @@
...
+
+// OrgIdSupport creates a new "org_id" column with an index in the "tenants" table and attempts to translate the
+// current EBS account numbers to "org_id"s.
+//
+// If the translation is not possible, it doesn't fail the migration: it simply logs an error. This is due to the
+// transition stage we are in. Since we are proactively taking steps towards the full "OrgId" support, not having a
+// translation for an EBS account number is acceptable.
+//
+// It also adds a comment to the "external_tenant" column, clarifying that it refers to "EBS account numbers".
+func OrgIdSupport() *gormigrate.Migration {
```
lpichler: (1)I would prefer whether this name would describe more what it is doing(+name file).
"AddOrgIDtoTenants"
"Translate EBS account number to ORG ID" and maybe it can also correspond sort of with "start" and "end" log messages 

(2) Also maybe it would be good to don't mix structural and data migrations.

Let me know what do you think about it.

db/migrations/20220330120000_org_id_support.go
MikelAlejoBR: This migration is supposed to be a "best effort" migration, since we're proactively adopting OrgIds. This means that if the service fails for whichever reason, we don't have to worry too much, since we are expecting some EANs to be untranslatable.

db/migrations/20220330120000_org_id_support.go
MikelAlejoBR: > (1)I would prefer whether this name would describe more what it is doing(+name file). "AddOrgIDtoTenants" "Translate EBS account number to ORG ID"

Sure. I am very bad at naming functions so that's why I end up writing comments on them to give them more context.

> (2) Also maybe it would be good to don't mix structural and data migrations.

Why is this? In this case the migration doesn't fail if the data migration does not succeed, because it's a "since we're at it, let's try populating the column too" kind of thing. That's why I conceived it as part of the same migration.

lpichler: (2) It is just my opinion - I think there is too much things in one migration:
- add column + add commenT
- translation 

it could be two migrations - reason is just to be more clear what is happening in one migration. I see one migration as one unit where less things (reasonable) as possible - and then then we don't need go to start back when anythings fail.
Also naming of migrations is more unambiguous without hiding additions in migration - you can see on first look what is going inside migration.
+ when we would test data migration it would be also more clear what is being tested.

yes and now I understand more role of the translation operation.

I don't insist  on it-  I leave it  up to you - I wanted to see your opinion - If you like we can adopt it as "rule" migration.

I would like also hear @lindgrenj6's opinion

MikelAlejoBR: > it could be two migrations - reason is just to be more clear what is happening in one migration. I see one migration as one unit where less things (reasonable) as possible - and then then we don't need go to start back when anythings fail.
> Also naming of migrations is more unambiguous without hiding additions in migration - you can see on first look what is going inside migration.

Okay that's a good point. After refactoring it is true that it is easier to follow. I submitted the commit.

lpichler: great, thanks!

db/migrations/20220404150000_translate_ebs_org_ids.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// TranslateEbsAccountNumbersToOrgIds attempts to translate the current EBS account numbers to "org_id"s.
+//
+// If the translation is not possible, it doesn't fail the migration: it simply logs an error. This is due to the
+// transition stage we are in. Since we are proactively taking steps towards the full "OrgId" support, not having a
+// translation for an EBS account number is acceptable.
+//
+// It also adds a comment to the "external_tenant" column, clarifying that it refers to "EBS account numbers".
+func TranslateEbsAccountNumbersToOrgIds() *gormigrate.Migration {
```
lpichler, line 21: this method  is not part of `migrations.migrationsCollection`

db/migrations/20220404150000_translate_ebs_org_ids.go
```diff
@@ -0,0 +1,130 @@
...
+		ID: "20220330120000",
+		Migrate: func(db *gorm.DB) error {
+			logging.Log.Info(`Migration "external tenants to org ids" started`)
+			defer logging.Log.Info(`Migration "external tenants to org ids" ended`)
+
+			// Perform the table migration.
+			err := db.Transaction(func(tx *gorm.DB) error {
+				// Add the comment to the "external_tenant" column manually, since doing it in the struct with a "gorm"
+				// tag makes "gormigrate" try to create the already existing column.
+				err := tx.
```
lpichler: I think this is in previous migration already :)

MikelAlejoBR:
OMG sorry! I copied & pasted the code and I swear I thought I had made the changes to remove the duplicated code, but I completely forgot about it :facepalm: 

The one thing I doubt is: should I implement the rollback for the second migration for this case? I mean to transform back the orgIds to EBSs.

lindgrenj6:
>The one thing I doubt is: should I implement the rollback for the second migration for this case? I mean to transform back the orgIds to EBSs.

I would just drop the org_id column in the rollback function there. The EBS account number column should contain the "old" numbers so no conversion back will be necessary.

MikelAlejoBR:
> > The one thing I doubt is: should I implement the rollback for the second migration for this case? I mean to transform back the orgIds to EBSs.
> 
> I would just drop the org_id column in the rollback function there. The EBS account number column should contain the "old" numbers so no conversion back will be necessary.

Oooooof you're right, you said it in a comment above, sorry.

lpichler (approved):

lindgrenj6 (approved):
(this is leftover from yesterday, I just didn't click submit)

LGTM, going to merge this next week after we do the cutover.

lindgrenj6:
~Adding TENANT_TRANSLTOR_HOST to stage here: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/37806~

lindgrenj6:
Update: we can't shadow the environment variables set in the refenv yaml here: https://gitlab.cee.redhat.com/jlindgre/app-interface/blob/d0bea6a25f79534a8d2c8dfe3818dceccd50c279/data/products/insights/environments/stage.yml#L57-58

The good news is that we won't need to specify these ourselves, great! But we need to change how we construct that URL. 

At the bottom of the template we can change the env var to `TENANT_TRANSLATOR_URL` and change the params to match the refenv yml: `${TENANT_TRANSLATOR_HOST}` and `${TENANT_TRANSLATOR_POSRT}`.

- [x] Add 2 params, TENANT_TRANSLATOR_HOST and TENANT_TRANSLATOR_PORT (defaulting to the ephemeral ones)
- [x] update the env var TENANT_TRANSLATOR_URL to build at runtime like we do for cost/swatch: https://github.com/RedHatInsights/sources-api-go/blob/950cab5a5d3cf3a2abf69b366ef7d4c517ce991e/deploy/clowdapp.yaml#L66

MikelAlejoBR:
All right I rebased the PR and included the `TRANSLATOR_HOST` and `TRANSLATOR_PORT` env vars as you said!

PR #189, opened 2022-03-30
renaming of return value from NewRelationObject() in SubCollectionList() dao methods

petracihalova:
without JIRA ticket

I found out that in `source_dao.go` in `SubCollectionList()` method we used `sourceType` name for value returned from `NewRelationObject()`

`sourceType, err := m.NewRelationObject(primaryCollection, *s.TenantID, DB.Debug())`

but this method is used not only for endpoint `source_types/id/sources` but for `application_types/id/sources` too => in second case was in variable `sourceType` the application type object 

so because this name is a little bit confusing I changed it in all `SubCollectionList()` methods into `relatedObject`


lpichler (approved):
LGTM, thanks @petracihalova

petracihalova:
rebased

lindgrenj6 (approved):

PR #190, opened 2022-03-30
Change the way the Vault AuthDao lists for Applications

lindgrenj6:
Ran into this when testing out GraphQL stuff. The old implementation was relying that the applications always had a join table record, but since we aren't creating that for the user (it is manual) we shouldn't be relying on it.

Instead I made it work similarly to how the `ListForEndpoints` method works, it just looks at the prefix of the secrets and from there fetches them.

### DIFF
--- a/dao/authentication_dao.go
+++ b/dao/authentication_dao.go
@@ -135,19 +135,22 @@ func (a *authenticationDaoImpl) ListForSource(sourceID int64, _, _ int, _ []util
 }
 
 func (a *authenticationDaoImpl) ListForApplication(applicationID int64, _, _ int, _ []util.Filter) ([]m.Authentication, int64, error) {
-	app := m.Application{ID: applicationID}
-	result := DB.
-		Where("tenant_id = ?", *a.TenantID).
-		Preload("ApplicationAuthentications").
-		First(&app)
-
-	if result.Error != nil {
-		return nil, 0, util.NewErrNotFound("application")
+	keys, err := a.listKeys()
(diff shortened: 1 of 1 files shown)


dao/authentication_dao.go
```diff
@@ -135,19 +135,22 @@ func (a *authenticationDaoImpl) ListForSource(sourceID int64, _, _ int, _ []util
...
 func (a *authenticationDaoImpl) ListForApplication(applicationID int64, _, _ int, _ []util.Filter) ([]m.Authentication, int64, error) {
-	app := m.Application{ID: applicationID}
-	result := DB.
-		Where("tenant_id = ?", *a.TenantID).
-		Preload("ApplicationAuthentications").
-		First(&app)
-
-	if result.Error != nil {
-		return nil, 0, util.NewErrNotFound("application")
+	keys, err := a.listKeys()
```
petracihalova, line 143: I believe we need first check, if given application ID exists ... something like in `ListForSource()`

petracihalova (changes requested):

dao/authentication_dao.go
lindgrenj6: Changed on all of them - and fixed the Source one since it wasn't using tenancy.

petracihalova: 👍🏻  great

petracihalova (approved):

MikelAlejoBR (approved):

lpichler (approved):

PR #193, opened 2022-03-31
Change BulkCreate ApplicationAuthentication logic to work with DB-backed DAO

lindgrenj6:
BulkCreate wasn't working for applicationAuthentications before mostly because of our DB schema being not-correct for the vault backend. 

In the future we're going to need to relax the constraints and change that table's layout a bit to fit both options. But today is not that day since we do not have a vault.

### DIFF
--- a/application_authentication_handlers_test.go
+++ b/application_authentication_handlers_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
+	"strconv"
 	"testing"
 
 	"github.com/RedHatInsights/sources-api-go/internal/testutils"
@@ -67,8 +68,8 @@ func TestApplicationAuthenticationList(t *testing.T) {
 		t.Error("ghosts infected the return")
 	}
 
-	authID := fixtures.TestAuthenticationData[0].ID
-	if appAuth["authentication_uid"] != authID {
+	authID := strconv.Itoa(int(fixtures.TestAuthenticationData[0].DbID))
+	if appAuth["authentication_id"].(string) != authID {
 		t.Error("ghosts infected the return")
 	}
 
@@ -128,8 +129,8 @@ func TestApplicationAuthenticationGet(t *testing.T) {
 	if err != nil {
 		t.Error("Failed unmarshaling output")
 	}
-	authID := fixtures.TestAuthenticationData[0].ID
-	if out.AuthenticationUID != authID {
+	authID := strconv.Itoa(int(fixtures.TestAuthenticationData[0].DbID))
+	if out.AuthenticationID != authID {
 		t.Error("ghosts infected the return")
 	}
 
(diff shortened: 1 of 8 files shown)


lpichler (approved):

MikelAlejoBR (approved):

MikelAlejoBR (changes requested):
Hang on a minute. I was curious about the following change:

```diff
- VaultPath string `json:"vault_path"`
+ VaultPath string `json:"vault_path" gorm:"-"`
```

Since I remember I had to add it back because some tests are failing. And even though the CI says everything is right... If I issue the `go test ./... -integration` command locally, I get a test failure in the DAO package....

```
--- FAIL: TestDeleteApplicationAuthentication (0.19s)                                                                                                                                                                                       
    application_authentication_dao_test.go:51: incorrect applicationAuthentication deleted. Want "complex uuid" in the authenticationUid field, got ""
```

And a panic from a dereference error in the status listener package. Can you please run the tests locally too with the above command and confirm that it's not just me?

In the meantime I'm going to investigate what's failing.

MikelAlejoBR:
> Since I remember I had to add it back because some tests are failing. And even though the CI says everything is right... If I issue the `go test ./... -integration` command locally, I get a test failure in the DAO package....
> 
> ```
> --- FAIL: TestDeleteApplicationAuthentication (0.19s)                                                                                                                                                                                       
>     application_authentication_dao_test.go:51: incorrect applicationAuthentication deleted. Want "complex uuid" in the authenticationUid field, got ""
> ```

Regarding this error, it corresponds to the following code block:

https://github.com/RedHatInsights/sources-api-go/blob/e70a41501ac415d28047f0dca80b59130fc05be4/dao/application_authentication_dao_test.go#L46-L53

It seems that when I wrote the test, I was using the UUID as an extra check to see if I was fetching the correct application authentication. I guess we can safely delete that code block.

MikelAlejoBR:
> And a panic from a dereference error in the status listener package. Can you please run the tests locally too with the above command and confirm that it's not just me?

Regarding this one, I've been reverting the main branch locally until the tests stopped failing, and it seems that the bug was introduced in the commit https://github.com/RedHatInsights/sources-api-go/commit/bfa08160a1eb3bd6121d75b1dadfec84f4356f3a, after merging the PR https://github.com/RedHatInsights/sources-api-go/pull/168 .

I've opened a ticket so that we don't forget to change the CI command so that it includes all the tests: https://issues.redhat.com/browse/RHCLOUD-18699

---

Since it's unrelated to this PR I think it is safe merging this PR after the DAO test change.

lindgrenj6:
I went through and removed all refs to `AuthenticationUID` and `VaultPath` since those aren't actually in the DB. We can re-add them in if/when we get a Vault.

MikelAlejoBR (approved):

lpichler (approved):

PR #195, opened 2022-04-01
[RHCLOUD-18700] fix: segmentation violation on status listener tests

MikelAlejoBR:
Due to the change introduced in bfa0816, the "lastAvailableAt" and "lastCheckedAt" attributes from the models can be nil. The status listener was counting on them always having a value, and that was causing the issue.

## Links

* https://github.com/RedHatInsights/sources-api-go/pull/168
* [[RHCLOUD-18700]](https://issues.redhat.com/browse/RHCLOUD-18700)

### DIFF
--- a/statuslistener/statuslistener_test.go
+++ b/statuslistener/statuslistener_test.go
@@ -95,13 +95,21 @@ func PopulateDateFieldsFrom(resource interface{}) DateFields {
 		dateFields.UpdatedAt = typedResource.UpdatedAt
 	case *m.Application:
 		dateFields.CreatedAt = typedResource.CreatedAt
-		dateFields.LastAvailableAt = *typedResource.LastAvailableAt
-		dateFields.LastCheckedAt = *typedResource.LastCheckedAt
+		if typedResource.LastAvailableAt != nil {
+			dateFields.LastAvailableAt = *typedResource.LastAvailableAt
+		}
+		if typedResource.LastCheckedAt != nil {
+			dateFields.LastCheckedAt = *typedResource.LastCheckedAt
+		}
 		dateFields.UpdatedAt = typedResource.UpdatedAt
 	case *m.Endpoint:
 		dateFields.CreatedAt = typedResource.CreatedAt
-		dateFields.LastAvailableAt = *typedResource.LastAvailableAt
-		dateFields.LastCheckedAt = *typedResource.LastCheckedAt
+		if typedResource.LastAvailableAt != nil {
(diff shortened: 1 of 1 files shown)


statuslistener/statuslistener_test.go
```diff
@@ -95,13 +95,21 @@ func PopulateDateFieldsFrom(resource interface{}) DateFields {
 		dateFields.UpdatedAt = typedResource.UpdatedAt
 	case *m.Application:
 		dateFields.CreatedAt = typedResource.CreatedAt
-		dateFields.LastAvailableAt = *typedResource.LastAvailableAt
-		dateFields.LastCheckedAt = *typedResource.LastCheckedAt
+		if typedResource.LastAvailableAt != nil {
```
petracihalova, line 98: Do we need the same change for case `*m.Source`?

petracihalova (changes requested):

statuslistener/statuslistener_test.go
MikelAlejoBR: No, and let me explain: in the case of these tests, we load the JSON in memory and we update the timestamps to the ones that the resources have in the database, because otherwise when the status listener generates the bulk message, it would have the latest timestamps from the database, but when comparing those timestamps to the ones hard coded in the JSON files, the comparison would fail. So, in order to avoid that, we fetch the resource and update the timestamps of the JSON:

https://github.com/RedHatInsights/sources-api-go/blob/e70a41501ac415d28047f0dca80b59130fc05be4/statuslistener/statuslistener_test.go#L308-L316

The `LastAvailableAt` and `LastCheckedAt` attributes are updated by the status listener upon reception of an availability status message:

https://github.com/RedHatInsights/sources-api-go/blob/e70a41501ac415d28047f0dca80b59130fc05be4/statuslistener/statuslistener.go#L144-L160

So, why do we need to set the `if`s only for the `Application` and `Endpoint` resources? Because they are the only ones that can be a subresource of a `Source`. The problem comes when we are trying to test the bulk message for the `Source` resource, because we fetch the source and all its subresources. However, those subresources —`Application`s and `Endpoint`s— have null `LastAvailableAt` and `LastCheckedAt` attributes because the status listener only received a status message for the `Source`, not for its subresources.

That explains the pointer dereference error.

petracihalova: Thanks for explanation !!!

petracihalova (approved):

lpichler (approved):

PR #196, opened 2022-04-01
[RHCLOUD-18699] fix: the entire test suite does not run

MikelAlejoBR:
It seems that having the integration flag before the "all tests" argument makes Go only execute the tests from the main package, instead of all the tests from all the packages.

## Links

* Needs https://github.com/RedHatInsights/sources-api-go/pull/195 to be merged first, since otherwise the tests will fail.
* [[RHCLOUD-18699]](https://issues.redhat.com/browse/RHCLOUD-18699)


petracihalova (approved):
Nice catch !!!

lpichler (approved):
good eye 👍

Makefile
```diff
@@ -34,7 +34,7 @@ test:
 	go test ./...
 
 alltest: test
-	go test --integration ./...
+	go test ./... -integration
```
lpichler: Is here difference between `-integration` and `--integration` ?

MikelAlejoBR: I've checked it and apparently there is not. I was confused since I've seen some of the Go commands using a single dash, and that's why I've been issuing these commands like that. But seeing that it works with a double dash, I'm sticking to that!

petracihalova (approved):

Makefile
lpichler: yes apparently there is not. When  I was trying it I had locally other db discrepancies - so it confuses me.
thanks for changing it 👍

PR #200, opened 2022-04-04
[RHCLOUD-18651][RHCLOUD-19093] Support orgIds in the incoming headers

MikelAlejoBR:
This PR takes care of the incoming headers both by the REST API and the Kafka messages, by allowing the creation and fetching of the tenants by either the EBS account number or the orgId. It also adds a log message to watch for anemic tenants.

## Links

* [[RHCLOUD-18651]](https://issues.redhat.com/browse/RHCLOUD-18651)
* [[RHCLOUD-19093]](https://issues.redhat.com/browse/RHCLOUD-19093)

### DIFF
--- a/dao/tenant_dao.go
+++ b/dao/tenant_dao.go
@@ -2,6 +2,7 @@ package dao
 
 import (
 	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
@@ -51,12 +52,32 @@ func (t *tenantDaoImpl) GetOrCreateTenantID(identity *identity.Identity) (int64,
 	return tenant.Id, result.Error
 }
 
-func (t *tenantDaoImpl) TenantByAccountNumber(accountNumber string) (*m.Tenant, error) {
-	tenant := m.Tenant{ExternalTenant: accountNumber}
+func (t *tenantDaoImpl) TenantByIdentity(id *identity.Identity) (*m.Tenant, error) {
+	// Start setting up the query.
+	query := DB.
+		Debug().
+		Model(&m.Tenant{})
 
-	result := DB.Debug().
-		Where("external_tenant = ?", accountNumber).
-		First(&tenant)
+	// Query by OrgId or EBS account number.
+	var tenant m.Tenant
+	if id.OrgID != "" {
+		tenant.OrgID = id.OrgID
 
-	return &tenant, result.Error
+		query.Where("org_id = ?", tenant.OrgID)
+	} else {
+		tenant.ExternalTenant = id.AccountNumber
+
+		query.Where("external_tenant = ?", tenant.ExternalTenant)
+	}
+
+	// Find the tenant, scanning into the struct above
+	err := query.
+		First(&tenant).
+		Error
+
+	if err != nil {
+		return nil, util.NewErrNotFound("tenant")
+	}
+
+	return &tenant, nil
 }
--- a/dao/tenant_dao.go
+++ b/dao/tenant_dao.go
@@ -23,55 +23,42 @@ func init() {
 type tenantDaoImpl struct{}
 
 func (t *tenantDaoImpl) GetOrCreateTenantID(identity *identity.Identity) (int64, error) {
-	// Start setting up the query.
-	query := DB.
-		Debug().
-		Model(&m.Tenant{})
-
-	// Query by OrgId or EBS account number.
+	// Try to find the tenant.
 	var tenant m.Tenant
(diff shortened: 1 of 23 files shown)


dao/tenant_dao.go
```diff
@@ -20,28 +22,62 @@ func init() {
...
+	var tenant m.Tenant
+	if identity.OrgID != "" {
+		tenant.OrgID = identity.OrgID
+
+		query.Where("org_id = ?", tenant.OrgID)
+	} else {
+		tenant.ExternalTenant = identity.AccountNumber
+
+		query.Where("external_tenant = ?", tenant.ExternalTenant)
+	}
```
lindgrenj6: So rather than looking up only by one or the other - let's just do something like this:
```go
result := DB.Debug().
	Where("org_id = ?", identity.OrgID).
	Or("external_tenant = ?", identity.AccountNumber).
	First(&tenant)

if result.Error != nil {
	tenant.OrgID = identity.OrgID
	tenant.AccountNumber = identity.AccoutNumber

	DB.Create(&tenant)
}
```

From there - make sure both the org_id and account_number columns are filled (updating if necessary, that way they're both covered). 

---

After doing the "both" query for a while we'll remove the `.Or(..)` line, drop the external_tenant column and remove it from the struct, and then nothing else will have to change!

middleware/headers.go
```diff
@@ -41,22 +38,16 @@ func ParseHeaders(next echo.HandlerFunc) echo.HandlerFunc {
 			// store it raw first.
 			c.Set("x-rh-identity", c.Request().Header.Get("x-rh-identity"))
```
lindgrenj6, line 43: So up above from here a bit we're going to need to check for the `x-rh-sources-org-id` header as well - that way they can specify either account OR org_id when using the PSK.

middleware/tenancy.go
```diff
@@ -36,32 +36,32 @@ func Tenancy(next echo.HandlerFunc) echo.HandlerFunc {
 			c.Logger().Debugf("Looking up Tenant ID for account number %v", accountNumber)
```
lindgrenj6, line 38: We'll handle the `psk-org-id` the same way we handle`psk-account` here.

lindgrenj6 (commented):
Good first pass, just need to change the lookup query + db population I think and then handle using an org id in the psk workflow.

dao/tenant_dao.go
MikelAlejoBR: I had to add a condition so that it checks we don't fetch the ones that have an empty `external_tenant` or `org_id`, since otherwise we could fetch the wrong tenant.

middleware/headers.go
MikelAlejoBR: Done!

middleware/tenancy.go
MikelAlejoBR: Done too!

lindgrenj6 (approved):
LGTM!

lpichler (approved):

lindgrenj6:
Been thinking about this one some more - we need to update `service.ForwardableHeaders(echo.Context)` to forward along the org-id. I think that'll really be the last thing and we'll be ready to go live with this shortly.

lindgrenj6:
Been thinking about this one some more - we need to update `service.ForwardableHeaders(echo.Context)` to forward along the org-id. I think that'll really be the last thing and we'll be ready to go live with this shortly.

MikelAlejoBR:
The PR has been rebased.

> Been thinking about this one some more - we need to update `service.ForwardableHeaders(echo.Context)` to forward along the org-id. I think that'll really be the last thing and we'll be ready to go live with this shortly.

Done!

service/events.go
```diff
@@ -60,5 +61,12 @@ func ForwadableHeaders(c echo.Context) []kafka.Header {
 		}
 	}
 
+	if c.Get("x-rh-sources-org-id") != nil {
```
lindgrenj6: One more thing here, up above on Line 59:

The "new" generated xrhid should include the org_id as well. Maybe we can move that extra else statement the end of this. Basically add a flag to see if the xrhid is there, and if it isn't generate it from the available fields (psk + org_id)

MikelAlejoBR: All right, done!

MikelAlejoBR:
I added the last request to the PR. It seems like I added a lot of code but the change boils down to:

* Since we're generating the `x-rh-identity` header from the `events.go` file, the marshal operation could go wrong, so the `ForwadableHeaders` function now can return an error as well.
* The signature change affected multiple files, so they got updated to handle the error, which involves just returning it to the error handling middleware.
* Added some tests for the `ForwadableHeaders` to solidify the function.
* Logged anemic tenants for identities that come with an OrgId and without an EBS account number.

PR #202, opened 2022-04-04
[RHCLOUD-18724] Use encryption when storing the password field on Authentication

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-18724

### DIFF
--- a/dao/authentication_dao_test.go
+++ b/dao/authentication_dao_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/marketplace"
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/redis"
+	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/hashicorp/vault/api"
 	"github.com/sirupsen/logrus"
 )
@@ -30,7 +31,7 @@ func setUpBearerToken() *marketplace.BearerToken {
 
 // setUpValidMarketplaceAuth sets up a valid authentication which is of the "marketplace" type.
 func setUpValidMarketplaceAuth() *m.Authentication {
-	return &m.Authentication{AuthType: "marketplace-token", Password: "apiKey"}
+	return &m.Authentication{AuthType: "marketplace-token", Password: util.StringRef("apiKey")}
 }
 
 // -----------------------------
@@ -199,7 +200,7 @@ func TestAuthFromVaultMarketplaceProviderEmptyPassword(t *testing.T) {
 	marketplaceTokenCacher = GetMarketplaceTokenCacher(&tenantId)
 
 	auth := setUpValidMarketplaceAuth()
-	auth.Password = "" // set up the empty password to simulate a missing API key
+	auth.Password = nil // set up the empty password to simulate a missing API key
 
 	// Call the function under test
 	err := setMarketplaceTokenAuthExtraField(auth)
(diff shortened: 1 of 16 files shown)


MikelAlejoBR (approved):

util/string_utils.go
```diff
@@ -5,3 +5,7 @@ import "strings"
 func Capitalize(str string) string {
 	return strings.ToUpper(string(str[0])) + str[1:]
 }
+
+func StringRef(str string) *string {
```
lpichler, line 9: we have already `PointerToString` which is same but it is in `testutils` folder maybe we can unify it somehow.

lindgrenj6: Yeah I'll remove that one and just make it use this.

dao/authentication_dao_test.go
```diff
@@ -320,7 +321,6 @@ func TestAuthFromVault(t *testing.T) {
 		Name:                    "test-vault-auth",
 		AuthType:                "test-authtype",
 		Username:                "my-username",
-		Password:                "my-password",
```
MikelAlejoBR, line 323: I would use `util.StringRef("my-password")` here...

dao/authentication_dao_test.go
```diff
@@ -348,6 +348,7 @@ func TestAuthFromVault(t *testing.T) {
 		t.Errorf(`wrong type for the secret's data object. Want "map[string]interface{}", got "%s"'`, reflect.TypeOf(vaultData["data"]))
 	}
 	data["last_available_at"] = authentication.LastAvailableAt.Format(time.RFC3339Nano)
+	data["password"] = "my-password"
```
MikelAlejoBR, line 354: ... so that you can use it here without hard coding the value
```diff
- data["password"] = "my-password"
+ data["password"] = *authentication.Password
```

dao/authentication_dao_test.go
```diff
@@ -393,10 +394,10 @@ func TestAuthFromVault(t *testing.T) {
 		}
 
 		{
-			want := authentication.Password
-			got := resultingAuth.Password
+			want := "my-password"
```
MikelAlejoBR, line 400: ... and same here: `*authentication.Password`.

model/authentication.go
```diff
@@ -74,6 +75,16 @@ func (auth *Authentication) ToResponse() *AuthenticationResponse {
 
 func (auth *Authentication) ToInternalResponse() *AuthenticationInternalResponse {
 	resourceID := strconv.FormatInt(auth.ResourceID, 10)
+	var pass string
+	if auth.Password == nil {
+		pass = ""
+	} else {
+		decrypted, err := util.Decrypt(*auth.Password)
+		if err != nil {
+			logger.Log.Errorf("failed to decrypt password: %v", err)
```
MikelAlejoBR: What do you think about including the authentication ID here, so that we can debug which one is causing problems?

model/authentication.go
```diff
@@ -74,6 +75,16 @@ func (auth *Authentication) ToResponse() *AuthenticationResponse {
 
 func (auth *Authentication) ToInternalResponse() *AuthenticationInternalResponse {
 	resourceID := strconv.FormatInt(auth.ResourceID, 10)
+	var pass string
+	if auth.Password == nil {
+		pass = ""
+	} else {
```
MikelAlejoBR: Since the `AuthenticationInternalResponse` has an `omitempty` tag, maybe we don't need to default it to the empty string

MikelAlejoBR (commented):
Just minor nitpicking comments, other than that looks good!

petracihalova (approved):

lpichler (approved):

model/authentication.go
lindgrenj6: I like that - added.

dao/authentication_dao_test.go
lindgrenj6: The main reason I didn't do that is because due to the way this unit test works the `toVaultMap` function would return a map[string]interface{} that then has `data["password"] = *"my-password"`, which isn't the way that the actual vault would be returning.

Instead a "real" vault would return a string OR null, so we're checking for `nil` before casting it to a `string`.

model/authentication.go
lindgrenj6: That is true, I let it default to `""` since that is the string default value.

PR #203, opened 2022-04-05
[RHCLOUD-18729] refactor: make integration tests use the real schema

MikelAlejoBR:
This accomplishes two things:

1. We test the migrations since they get applied all the time.
2. We actually test again the database structure we use in development,
staging and production environments.

## Links

[[RHCLOUD-18729]](https://issues.redhat.com/browse/RHCLOUD-18729)

### DIFF
--- a/dao/application_authentication_dao_test.go
+++ b/dao/application_authentication_dao_test.go
@@ -18,25 +18,13 @@ func TestDeleteApplicationAuthentication(t *testing.T) {
 
 	applicationAuthenticationDao := GetApplicationAuthenticationDao(&fixtures.TestSourceData[0].TenantID)
 
-	applicationAuthentication := fixtures.TestApplicationAuthenticationData[0]
-	// Set the ID to 0 to let GORM know it should insert a new applicationAuthentication and not update an existing one.
-	applicationAuthentication.ID = 0
-	// Set some data to compare the returned applicationAuthentication.
-	applicationAuthentication.AuthenticationUID = "complex uuid"
-
-	// Create the test applicationAuthentication.
-	err := applicationAuthenticationDao.Create(&applicationAuthentication)
-	if err != nil {
-		t.Errorf("error creating applicationAuthentication: %s", err)
-	}
-
-	deletedApplicationAuthentication, err := applicationAuthenticationDao.Delete(&applicationAuthentication.ID)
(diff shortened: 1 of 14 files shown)


lpichler:
CI Looks related.

endpoint_handlers_test.go
```diff
@@ -578,12 +579,19 @@ func TestEndpointDelete(t *testing.T) {
 	tenantID := fixtures.TestTenantData[0].Id
 	sourceDao := dao.GetSourceDao(&tenantID)
 
+	uid, err := uuid.NewUUID()
+	if err != nil {
+		t.Errorf(`could not create UUID fro the fixture source: %s`, err)
```
petracihalova: typo in err message "fro" - "from"

dao/application_authentication_dao_test.go
```diff
@@ -250,6 +238,7 @@ func TestApplicationAuthenticationsByAuthenticationsDatabase(t *testing.T) {
 // and correct count of returned objects
 func TestApplicationAuthenticationListOffsetAndLimit(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
+	testutils.SkipIfNotSecretStoreDatabase(t)
```
petracihalova, line 241: not sure why we need to skip this test when SECRET_STORE is not database?

petracihalova (approved):
LGTM!

endpoint_handlers_test.go
MikelAlejoBR: Thank you!

dao/application_authentication_dao_test.go
MikelAlejoBR: If you take a look at the `dao/application_authentication_dao#ApplicationAuthenticationsByAuthentications` function, you'll see that we've got a `query.Where("authentication_uid IN ?", authUuids)` condition when we're using Vault. That column does not exist yet, and therefore was giving me problems when running the tests.

Once we start using Vault, we will need to add a test for that one too.

petracihalova: thanks ... I was thinking that I tried is with vault option and the test passed but now I tried it again and we need this skip for vault ... not sure what I did wrong before, so again thanks

lindgrenj6 (approved):

PR #204, opened 2022-04-05
Add notification integration

lpichler:
This PR sends Kafka message (according to notification doc.) which represent notification when av. status on any resource was changed. There are two situation when status can happen - when PATCH request is send or when status message is processed in availability listener.

---
This PR adds notification emission when available status(Source, Application, Authentication, Endpoint) is changed, change are present in :
- PATCH requests:
   - updates on handlers
     - get previous status
     - get current status
     - set it to context - so it can be processed by notification middleware
  - `Notifier` middleware which call function for notification emission(`EmitAvailabilityStatusNotification`) from new `notifications` service package 
  - `EmailNotificationInfo` method added to models which covers attributes for email templates (sent via Kafka message)
  - added `notifications.go` in service package which covers sending notification in kafka message form (it is process by notification service)
- Availability Status Listener
  - get previous av. status from resource in status message 
  - change in signature of `FetchAndUpdateBy`  we need to return resource in order to populate  `EmitAvailabilityStatusNotification`  (together with previous av. status)

### Links
PR in notification backend:
- [x] https://github.com/RedHatInsights/notifications-backend/pull/1097
https://issues.redhat.com/browse/RHCLOUD-16269


model/email_notification_info.go
```diff
@@ -0,0 +1,13 @@
+package model
+
+type EmailNotificationInfoInterface interface {
```
lindgrenj6: What would you think of renaming the interface to `EmailNotification` then the method could just be named `ToEmail`? 

That way it largely resembles our `model.Event` interface as well.

routes.go
```diff
@@ -36,7 +36,7 @@ func setupRoutes(e *echo.Echo) {
 		r.GET("/sources", SourceList, tenancyWithListMiddleware...)
 		r.GET("/sources/:id", SourceGet, middleware.Tenancy)
 		r.POST("/sources", SourceCreate, permissionMiddleware...)
-		r.PATCH("/sources/:id", SourceEdit, permissionMiddleware...)
+		r.PATCH("/sources/:id", SourceEdit, append(permissionMiddleware, middleware.Notifier)...)
```
lindgrenj6, line 39: I think we could make a new `editMiddleware` slice that contains this. Since we're using the same middleware for all the PATCH methods it looks like.

service/notifications.go
```diff
@@ -0,0 +1,125 @@
...
+	err = mgr.Produce(msg)
+	if err != nil {
+		l.Log.Warnf("Failed to produce kafka message to emit notification: %v, error: %v", statusEventType, err)
+		return err
+	}
+
+	return nil
+}
+
+func EmailNotificationInfoFromResource(resource interface{}, previousStatus string) *m.EmailNotificationInfo {
```
lindgrenj6: Since the `ToEmailNotificationInfo` is hidden behind an interface this function shouldn't be needed anymore. Anywhere you need to call it just do this:

`var emailable model.EmailNotification = &resource`, then you will have access to that method. 

Heres a playable example: https://goplay.space/#zVZ1ZccOC-K

statuslistener/statuslistener.go
```diff
@@ -116,15 +117,30 @@ func (avs *AvailabilityStatusListener) processEvent(statusMessage types.StatusMe
...
 	if err != nil {
-		l.Log.Errorf("Update error in status availability: %s", err)
+		l.Log.Errorf("unable to update availability status: %s", err)
 		return
 	}
 
+	if previousStatus != statusMessage.Status {
+		emailInfo := service.EmailNotificationInfoFromResource(resultRecord, previousStatus)
+		if emailInfo != nil {
+			err = service.NotificationProducer.EmitAvailabilityStatusNotification(accountNumber, emailInfo)
```
lindgrenj6: Rather than calling a method on a package variable directly I'd rather have this just be a function that uses a default var at the package level. Much like we do for `service.RaiseEvent(...)`

service/notifications.go
```diff
@@ -0,0 +1,125 @@
...
+type notificationPayload struct {
+}
+
+type notificationEvent struct {
+	Metadata string `json:"metadata"`
+	Payload  string `json:"payload"`
+}
+
+type notificationMetadata struct {
+}
```
lindgrenj6: (minor): Probably worth adding some comments around these structs - I was a little confused at first but then I realized they're probably just here for the formatting of the JSON payload that they expect these fields even if they're empty.

lindgrenj6 (commented):
Just a few small comments - just some stuff about using an interface and how an implementation gets called. 

Other than those things looks great!!

service/notifications.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		l.Log.Warnf("error in marshalling payload ")
+		return err
+	}
+
+	metadata, err := json.Marshal(&notificationMetadata{})
+	if err != nil {
+		l.Log.Warnf("error in marshalling payload ")
+		return err
+	}
```
MikelAlejoBR, line 91: I see that we are marshalling empty structs here. I guess this is related to what Jacob said, right? The payload requires these to be on it.

MikelAlejoBR, line 91: I would like to suggest some changes to the `Warnf` messages:

```diff
if err != nil {
-    l.Log.Warnf("error in marshalling context ")
+    l.Log.Warnf(`error when marshalling the email notification information: %s`, err)
    return err
}
```

```diff
payload, err := json.Marshal(&notificationPayload{})
if err != nil {
-    l.Log.Warnf("error in marshalling payload ")
+    l.Log.Warnf(`error when marshalling the email notification payload: %s`, err)
    return err
}
```

```diff
metadata, err := json.Marshal(&notificationMetadata{})
if err != nil {
-    l.Log.Warnf("error in marshalling payload ")
+    l.Log.Warnf(`error when marshalling the email notification metadata: %s`, err)
    return err
}
```

Just for the sake of adding more context to the log message.

service/notifications.go
```diff
@@ -0,0 +1,125 @@
...
+		EventType:   statusEventType,
+		Timestamp:   time.Now().Format(time.RFC3339),
+		AccountID:   accountNumber,
+		Events:      []notificationEvent{event},
+		Context:     string(context),
+		Recipients:  []notificationRecipients{},
+	})
+
+	if err != nil {
+		l.Log.Warnf("Failed to add struct value as json to kafka message")
```
MikelAlejoBR: I tthink the error is missing from this log statement:

```diff
- l.Log.Warnf("Failed to add struct value as json to kafka message")
+ l.Log.Warnf("Failed to add struct value as json to kafka message: %s", err)
```

MikelAlejoBR (commented):
Some minor comments. Other than that it looks good :+1:

service/notifications.go
lpichler: yes, correct

service/notifications.go
lpichler: this is nice

statuslistener/statuslistener.go
lpichler: I did it but to accomplish it - I call this method on struct  in other method at  package level in order to use `NotificationProducer` with interface so I can use for mocking it in tests.

service/notifications.go
lindgrenj6: I'm glad you like it! It's a nice way to hide behavior behind interfaces.

lpichler:
rebased 👍

service/notifications.go
```diff
@@ -0,0 +1,123 @@
...
+type AvailabilityStatusNotifier struct {
+}
+
+var (
+	application                = "sources"
+	notificationTopic          = config.Get().KafkaTopic("platform.notifications.ingress")
+	statusEventType            = "availability-status"
+	bundle                     = "console"
+	notificationMessageVersion = "v1.1.0"
+)
```
MikelAlejoBR: I guess these could be `const`s

model/events.go
```diff
@@ -14,7 +14,7 @@ type Event interface {
 
 type EventModelDao interface {
 	BulkMessage(resource util.Resource) (map[string]interface{}, error)
-	FetchAndUpdateBy(resource util.Resource, updateAttributes map[string]interface{}) error
+	FetchAndUpdateBy(resource util.Resource, updateAttributes map[string]interface{}) (interface{}, error)
```
MikelAlejoBR, line 17: I think you can return `m.EmailNotification` here instead of an `interface{}`, since I think that all the models that implement the `m.EventModelDao` interface tend to be type checked against that interface later.

MikelAlejoBR (approved):
Just a couple of minor comments. Other than that, looks good :+1:

lindgrenj6 (approved):

model/events.go
lpichler: wow yes this is possible -  good eye - but I then it feels to me strange that this method would return `m.EmailNotification` I would like to look on it more in follow up PR and maybe also change some names here, if you are ok with it.

MikelAlejoBR: Sure, yeah!

lpichler:
I am merging this as I got approvals for @lindgrenj6 and @MikelAlejoBR, thanks!

PR #207, opened 2022-04-06
[RHCLOUD-18515] [WIP] Offset fix

petracihalova:
[RHCLOUD-18515](https://issues.redhat.com/browse/RHCLOUD-18515)

When we listing resources and we use offset filter, the count value was 0 in many cases (because "count" returns a single row and using offset returns an empty value). Now using of offset is fixed so queries works properly with and without offset.


application_handlers_test.go
```diff
@@ -263,7 +318,35 @@ func TestApplicationListBadRequestInvalidFilter(t *testing.T) {
...
+
+		err := ApplicationList(c)
+		if err != nil {
+			t.Error(err)
+		}
+
+		path := c.Request().RequestURI
+		templates.WithOffsetAndLimitTest(t, path, rec, wantApplicationsCount, i["limit"], i["offset"])
+	}
 }
```
MikelAlejoBR, line 350: If I recall correctly, the bug affected the DAO functions which had an offset different than zero. That is the reason why, in my opinion, it is not needed to test the handlers themselves but just the affected DAO functions. I was thinking on something like this for the `List*` functions:

```go
// TestApplicationListOffsetRegression is a regression test for RHCLOUD-18515. The test tests that the offset doesn't
// affect the returned total count.
func TestApplicationListOffsetRegression(t *testing.T) {
	CreateFixtures("applicationoffsetregression")

	applicationDao := GetApplicationDao(&fixtures.TestTenantData[0].Id)

	offset := 1
	_, count, err := applicationDao.List(0, offset, []util.Filter{})
	if err != nil {
		t.Errorf(`unexpected error when listing the applications: %s`, err)
	}

	want := len(fixtures.TestApplicationData)
	got := count

	if want != got {
		t.Errorf(`incorrect total count returned. Want "%d", got "%d"`, want, got)
	}
	
	DoneWithFixttures("applicationoffsetregression")
}
```

With this we ensure that we won't get hit by a regression in the future, and the test is focused on the bug itself exactly where it was located.

petracihalova: yes 😁  that makes sense ... I didn't realise that is not needed to go from handlers in this case

petracihalova:
I am closing this PR because I created new PR for requested changes: https://github.com/RedHatInsights/sources-api-go/pull/213

PR #208, opened 2022-04-06
[RHCLOUD-18723] Cascade delete resources

MikelAlejoBR:
Implements deleting the following resources with cascade:

- Source
- Application
- Endpoint

The way it has been done is the following:

1. All the resources and sub resources are deleted from the database in a transaction. We use the "RETURNING" statement to get all the deleted resources.
2. We raise an event for every single resources we have deleted.
3. We collect all the authentications for every resource we have deleted.
4. We delete the authentications in batch and raise an event for every authentication deleted.

This has been done to not leave the client with dangling resources on the database. Not being able to delete authentications isn't that bad, since:

1. They are encrypted.
2. Nobody else will have a reference to them, so they will be unrecoverable from the app.

This situation, if it happens, will require manual action from our side. I assumed that it is better and easier to clean up dangling authentications than it is to clean dangling resources and sub resources.

---

As a collateral change: I refactored the tests so that we only use one connection for each test, instead of opening a new one. Until now every time a test was ran a new connection was being opened, which has ended up making Postgres complain about "too many active connections".

## Links

[[RHCLOUD-18723]](https://issues.redhat.com/browse/RHCLOUD-18723)

### DIFF
--- a/service/subresource_destroyer.go
+++ b/service/subresource_destroyer.go
@@ -45,9 +45,9 @@ func DeleteCascade(tenantId *int64, resourceType string, resourceId int64, heade
 
 		// Raise an event for the deleted application authentications.
 		for _, appAuth := range applicationAuthentications {
-			err = RaiseEvent("ApplicationAuthentication.delete", &appAuth, headers)
+			err = RaiseEvent("ApplicationAuthentication.destroy", &appAuth, headers)
 			if err != nil {
-				logging.Log.Errorf(`Event "ApplicationAuthentication.delete" could not be raised for application authentication %v: %s`, appAuth.ToEvent(), err)
+				logging.Log.Errorf(`Event "ApplicationAuthentication.destroy" could not be raised for application authentication %v: %s`, appAuth.ToEvent(), err)
 			}
 		}
 
@@ -55,9 +55,9 @@ func DeleteCascade(tenantId *int64, resourceType string, resourceId int64, heade
 		var appIds []int64
 		for _, app := range applications {
 			appIds = append(appIds, app.ID)
-			err := RaiseEvent("Application.delete", &app, headers)
+			err := RaiseEvent("Application.destroy", &app, headers)
 			if err != nil {
-				logging.Log.Errorf(`Event "Application.delete" could not be raised for application %v: %s`, app.ToEvent(), err)
+				logging.Log.Errorf(`Event "Application.destroy" could not be raised for application %v: %s`, app.ToEvent(), err)
 			}
 		}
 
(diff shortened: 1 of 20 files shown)


service/subresource_destroyer.go
```diff
@@ -0,0 +1,158 @@
...
+		applications, endpoints, rhcConnections, source, err := sourceDao.DeleteCascade(resourceId)
+		if err != nil {
+			return fmt.Errorf(`could not completely delete the source: %s`, err)
+		}
+
+		// Raise events for the deleted applications.
+		var appIds []int64
+		for _, app := range applications {
+			appIds = append(appIds, app.ID)
+			err := RaiseEvent("Application.Delete", &app, headers)
```
lindgrenj6: One teeny thing - it has to be `<Resource>.delete`

service/subresource_destroyer.go
```diff
@@ -0,0 +1,158 @@
...
+		for _, res := range resourceTypes {
+			auths, err := authenticationsDao.ListIdsForResource(res.ResourceType, res.ResourceIds)
+			if err != nil {
+				logging.Log.Errorf(`[resource_type: "%s"][resource_ids: "%v"] Could not fetch authentications: %s`, res.ResourceType, res.ResourceIds, err)
+			}
+
+			authentications = append(authentications, auths...)
+		}
+
+	case "Application":
```
lindgrenj6, line 107: We do need to delete the `ApplicationAuthentications` as well - which might need to be deleted first due to fkeys.

service/subresource_destroyer.go
MikelAlejoBR: All right, fixed!

service/subresource_destroyer.go
MikelAlejoBR: I cascade deleted the applications too, to be able to notify about the deleted application authentications as well.

service/subresource_destroyer.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+
+		// Fetch all the application's authentications.
+		auths, err := authenticationsDao.ListIdsForResource("Application", []int64{resourceId})
+		if err != nil {
+			logging.Log.Errorf(`[resource_type: "Application"][resource_id: "%v"] Could not fetch authentications: %s`, resourceId, err)
+		}
+
+		authentications = append(authentications, auths...)
+	case "Endpoint":
```
lindgrenj6, line 135: Endpoints can have authentications attached to them - so we should probably add the cascade logic to that DAO as well.

MikelAlejoBR: The enpoints' authentications already get cleaned up: we call `authenticationsDao.ListIdsForResource("Endpoint", []int64{resourceId})` and then we deal with those authentications outside the `switch`.

lpichler:
@MikelAlejoBR it looks that we have named eventype as  `<resource_type>.destroy`   in raising events in rails app but in your PR there are event types as `<resource_type>.delete`. @lindgrenj6 is this ok or do we need `<resource_type>.destroy` ?

MikelAlejoBR:
> @MikelAlejoBR it looks that we have named eventype as `<resource_type>.destroy` in raising events in rails app but in your PR there are event types as `<resource_type>.delete`. @lindgrenj6 is this ok or do we need `<resource_type>.destroy` ?

I guess that we should be using the old "destroy". I just submitted a commit to fix that.

service/subresource_destroyer.go
```diff
@@ -0,0 +1,173 @@
...
+	case "Source":
+		sourceDao := dao.GetSourceDao(tenantId)
+		applicationAuthentications, applications, endpoints, rhcConnections, source, err := sourceDao.DeleteCascade(resourceId)
+		if err != nil {
+			return fmt.Errorf(`could not completely delete the source: %s`, err)
+		}
+
+		// Raise an event for the deleted application authentications.
+		for _, appAuth := range applicationAuthentications {
+			err = RaiseEvent("ApplicationAuthentication.destroy", &appAuth, headers)
```
lpichler, line 48: This is related to most likely with all raised events: 


```json
{
   "id":14,
   "created_at":"2022-04-11 17:28:38 UTC",
   "updated_at":"2022-04-11 17:28:38 UTC",
   "paused_at":null,
   "application_id":26,
   "authentication_id":22,
   "tenant":""
}
```

`tenant`  as an external tenant is missing in those payloads

lpichler:
@MikelAlejoBR awesome work 👏 I have 2 comments after addressing them I am good with it :+1 (and you already accomplished first one :) ) thanks!

MikelAlejoBR:
All right I fixed the issue with the tenants not being properly returned, and also performed some tests with Vault to check that everything works as intended.

lpichler (approved):
LGTM 👍

lindgrenj6 (approved):

PR #209, opened 2022-04-06
Make various Authentication fields nullable in order for Rails to interop fine.

lindgrenj6:
Found this when testing interoperability between the rails <-> go password fields. some fields such as availability_status and name and others _need to be nullable_. 

This PR fixes that.

### DIFF
--- a/dao/authentication_db_dao.go
+++ b/dao/authentication_db_dao.go
@@ -318,7 +318,7 @@ func (add *authenticationDaoDbImpl) ListForEndpoint(endpointID int64, limit, off
 }
 
 func (add *authenticationDaoDbImpl) Create(authentication *m.Authentication) error {
-	query := DB.Select("source_id").Where("tenant_id = ?", *add.TenantID)
+	query := DB.Debug().Select("source_id").Where("tenant_id = ?", *add.TenantID)
 
 	switch strings.ToLower(authentication.ResourceType) {
 	case "application":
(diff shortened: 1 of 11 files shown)


lindgrenj6:
I've been trying to figure out why the statuslistener tests have started failing but it looks like the ordering of some of the fields has changed consistently. I'm not sure how to fix them. There might be a null somewhere in there instead of a "" but even then the ordering is definitely different across the fields.

dao/authentication_db_dao.go
```diff
@@ -317,6 +318,31 @@ func (add *authenticationDaoDbImpl) ListForEndpoint(endpointID int64, limit, off
...
+		if result.Error != nil {
+			return fmt.Errorf("resource not found with type [%v], id [%v]", authentication.ResourceType, authentication.ResourceID)
+		}
+
+		authentication.SourceID = endpoint.SourceID
+	case "source":
+		authentication.SourceID = authentication.ResourceID
+	default:
+		return fmt.Errorf("bad resource type, supported types are [Application, Endpoint, Source]")
+	}
```
MikelAlejoBR, line 344: I'm missing the `.Debug` statements in the queries

MikelAlejoBR:
I don't know about the status listener bit, but the other tests seem to be failing just because the authentication is probably missing the resource type, so they should be easy to fix.

lpichler:
regard to listener test, probably it is needed to change in example json files:

```diff
diff --git a/statuslistener/test_data/bulk_message_Application_1.json b/statuslistener/test_data/bulk_message_Application_1.json
index a8ed208..7b86bd1 100644
--- a/statuslistener/test_data/bulk_message_Application_1.json
+++ b/statuslistener/test_data/bulk_message_Application_1.json
@@ -50,7 +50,7 @@
     {
       "authtype": "test",
       "availability_status": "available",
-      "availability_status_error": "",
+      "availability_status_error": null,
       "created_at": "2022-01-10T14:58:02.850209Z",
       "extra": {
       },
diff --git a/statuslistener/test_data/bulk_message_Source_1.json b/statuslistener/test_data/bulk_message_Source_1.json
index 8df57ed..4eae0aa 100644
--- a/statuslistener/test_data/bulk_message_Source_1.json
+++ b/statuslistener/test_data/bulk_message_Source_1.json
@@ -50,7 +50,7 @@
     {
       "authtype": "test",
       "availability_status": "available",
-      "availability_status_error": "",
+      "availability_status_error": null,
       "created_at": "2022-01-10T14:58:02.850209Z",
       "extra": {
       },
```

PR #210, opened 2022-04-08
[RHCLOUD-18784] added function for closing test db connection

petracihalova:
[RHCLOUD-18784](https://issues.redhat.com/browse/RHCLOUD-18784)

When I was working on tests in dao layer, I started to get this error:

```
2022/04/08 13:55:12 /Users/pcihalov/projects/sources-api-go/dao/main_test.go:53 
failed to connect to `host=localhost user=root database=sources_api_test_go`: server error 
(FATAL: sorry, too many clients already (SQLSTATE 53300))
```

When we call `DoneWithFixtures()`, the `ConnectToTestDB("dao")` is called again and again and connections are not closed. So I added function which closes this connection before we connect again to TestDB again.

### DIFF
--- a/dao/main_test.go
+++ b/dao/main_test.go
@@ -96,6 +96,7 @@ func CreateFixtures(schema string) {
 // the database in the previous schema.
 func DoneWithFixtures(schema string) {
 	DropSchema(schema)
+	CloseTestDbConnection()
 	ConnectToTestDB("dao")
 }
 
@@ -105,7 +106,14 @@ func DropSchema(dbSchema string) {
 	if result.Error != nil {
 		log.Fatalf("Error in drop schema %s %s: ", dbSchema, result.Error.Error())
 	}
+}
 
+func CloseTestDbConnection() {
+	rawDB, err := DB.DB()
+	if err != nil {
+		log.Fatal(err)
+	}
+	rawDB.Close()
 }
 
 // MigrateSchema migrates all the models.


lpichler:
Can we avoid closing connection as is pointed [here](https://github.com/go-gorm/gorm/issues/3145#issuecomment-658526565)  ?

petracihalova:
> Can we avoid closing connection as is pointed [here](https://github.com/go-gorm/gorm/issues/3145#issuecomment-658526565) ?

yes I know about this, but I chose this solution because I need it for my task I am working on now and it is easy fix for now ... but for sure it will be better to refactor this in future ... all tests in dao (dozens tests) use this logic and I believe refactor will be bigger change, but maybe I am wrong

petracihalova:
I created task for this potential refactor [RHCLOUD-18784](https://issues.redhat.com/browse/RHCLOUD-18784)

lpichler:
Always I prefer don't add technical debt if it is possible - for now - maybe we can do just something with call `ConnectToTestDB` in `DoneWithFixtures` - it looks that we don't need to create new connection each time when we call `DoneWithFixtures` but we might create it once and then we can assign it back to DB in this function (as I read it in comment of `DoneWithFixtures `)

MikelAlejoBR:
I have run into the same issue in https://github.com/RedHatInsights/sources-api-go/pull/208 . What I ended up doing was, as you did, closing the connection and reopening a new one for each time the tests are run, because I don't see any other way around Gorm. In fact I might have implemented that refactor that you linked, Petra.

> Can we avoid closing connection as is pointed [here](https://github.com/go-gorm/gorm/issues/3145#issuecomment-658526565) ?

I tried it myself using some ridiculously high values for those parameters but it keeps failing, since it's Postgres the one that barks at this for having too many connections. So the quick-n-dirty fix would be to increase the maximum active connections in the database, but we would eventually face the same problem in the long run.

> Always I prefer don't add technical debt if it is possible - for now - maybe we can do just something with call `ConnectToTestDB` in `DoneWithFixtures` - it looks that we don't need to create new connection each time when we call `DoneWithFixtures` but we might create it once and then we can assign it back to DB in this function (as I read it in comment of `DoneWithFixtures `)

The problem is that you can't modify Gorm's `TablePrefix` for an open connection, so if you create a new schema for your tests and try to migrate them, using `SET search_path TO <schema>` doesn't change Gorm's `TablePrefix`, and you end up migrating the schema the connection is pointing to.

The `AutoMigrate` doesn't allow you to change the schema, so I don't see any other way...

petracihalova:
This or change mentioned by Mikel in https://github.com/RedHatInsights/sources-api-go/pull/208 is needed for https://github.com/RedHatInsights/sources-api-go/pull/213

lindgrenj6 (approved):

PR #213, opened 2022-04-11
[RHCLOUD-18515] Offset fix version 2

petracihalova:
[RHCLOUD-18515](https://issues.redhat.com/browse/RHCLOUD-18515)

When we listing resources and we use offset filter, the count value was 0 in many cases (because "count" returns a single row and using offset returns an empty value). Now using of offset is fixed so queries works properly with and without offset.

First version here: https://github.com/RedHatInsights/sources-api-go/pull/207

I decided to create new PR for second version of this fix. 
The tests pass after the "too many clients" error issue is resolved (https://github.com/RedHatInsights/sources-api-go/pull/210).

### DIFF
--- a/dao/application_authentication_dao_test.go
+++ b/dao/application_authentication_dao_test.go
@@ -248,7 +248,7 @@ func TestApplicationAuthenticationsByAuthenticationsDatabase(t *testing.T) {
 // and correct count of returned objects
 func TestApplicationAuthenticationListOffsetAndLimit(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
-	CreateFixtures("offset_limit")
+	SwitchSchema("offset_limit")
 
 	appAuthDao := GetApplicationAuthenticationDao(&fixtures.TestTenantData[0].Id)
 	wantCount := int64(len(fixtures.TestApplicationAuthenticationData))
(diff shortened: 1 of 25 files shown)


petracihalova:
rebased and conflicts solved

MikelAlejoBR (approved):
Nice work, looks good to me! :+1: :+1:

lpichler (approved):
LGTM 👏

petracihalova:
rebased and conflicts solved

petracihalova:
rebased and conflicts resolved

petracihalova:
rebased and conflicts solved

lindgrenj6 (approved):
Ironically found this was the cause for the UI breaking after switching to the Go GraphQL impelmentation - I started working on a fix but then @MikelAlejoBR pointed me here. 


I apologize for not getting to this sooner. This will fix page 2+ of the UI.

dao/application_authentication_dao_test.go
```diff
@@ -243,3 +243,38 @@ func TestApplicationAuthenticationsByAuthenticationsDatabase(t *testing.T) {
 
 	DropSchema("appauthfind")
 }
+
+// TestApplicationAuthenticationListOffsetAndLimit tests that List() in app auth dao returns correct count value
+// and correct count of returned objects
+func TestApplicationAuthenticationListOffsetAndLimit(t *testing.T) {
```
lindgrenj6, line 249: For future reference - I think these should have been DAO tests since this functionality is handled by the DAO and not the handlers (e.g. we use the List functions from other parts of the code like the availability status listener or the job engine)

Merging anyway to fix the UI.

PR #216, opened 2022-04-11
[RHCLOUD-18864] fixed logic in SubCollectionList() methods in mock dao

petracihalova:
[RHCLOUD-18864](https://issues.redhat.com/browse/RHCLOUD-18864)

Logic in SubCollectionList() methods needed to be fixed. 

I found out in mock daos for `SubCollectionList()` for sources that there is not good logic for route `application_types/id/sources` ... before, we returned sources where source id = application type id (and it was working because our test data was not so perfect)

line 88 from mock_daos.go
```
// Application type exists = return sources subcollection
for index, source := range src.Sources {
    if source.ID == object.Id {
        sources = append(sources, src.Sources[index]
    )
```

but it is simply wrong ... first we need to find out if application type exists (if not, we return not found error) and then we need to find all applications with given app type, get source ids from these apps and then return list of these sources

similar situation was for `SubCollectionList() `for app type (`sources/id/application_types`) ... first we need to find out if source exists and if so, then find all applications with given source id and create list of app types and then return the list with app types objects

then I updated our fixtures to have there 2 apps with different app type for 1 source id and because this, I needed to update test data for status listener too

### DIFF
--- a/internal/testutils/helpers.go
+++ b/internal/testutils/helpers.go
@@ -6,6 +6,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/config"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/fixtures"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/parser"
+	"github.com/RedHatInsights/sources-api-go/model"
 )
 
 var conf = config.Get()
@@ -24,14 +25,30 @@ func SkipIfNotSecretStoreDatabase(t *testing.T) {
 	}
 }
 
-func GetSourcesCountWithAppType(appTypeId int64) int {
-	var sourcesCount int
+func GetSourcesWithAppType(appTypeId int64) []model.Source {
+	var sourceIds = make(map[int64]struct{})
 
+	// Find applications with given application type and get
+	// list of unique source IDs
 	for _, app := range fixtures.TestApplicationData {
 		if app.ApplicationTypeID == appTypeId {
-			sourcesCount++
+			_, ok := sourceIds[app.SourceID]
+			if !ok {
+				sourceIds[app.SourceID] = struct{}{}
+			}
 		}
 	}
 
-	return sourcesCount
+	// Find sources for source IDs
+	var sources []model.Source
+	for id := range sourceIds {
+		for _, src := range fixtures.TestSourceData {
+			if id == src.ID {
+				sources = append(sources, src)
+				break
+			}
+		}
+	}
+
+	return sources
 }
--- a/internal/testutils/helpers.go
+++ b/internal/testutils/helpers.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/RedHatInsights/sources-api-go/config"
+	"github.com/RedHatInsights/sources-api-go/internal/testutils/fixtures"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/parser"
 )
 
@@ -21,4 +22,16 @@ func SkipIfNotSecretStoreDatabase(t *testing.T) {
 	if conf.SecretStore == "vault" {
 		t.Skip("Skipping test")
 	}
+}
+
+func GetSourcesCountWithAppType(appTypeId int64) int {
+	var sourcesCount int
+
+	for _, app := range fixtures.TestApplicationData {
+		if app.ApplicationTypeID == appTypeId {
+			sourcesCount++
+		}
+	}
+
+	return sourcesCount
 }
--- a/application_handlers_test.go
+++ b/application_handlers_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"reflect"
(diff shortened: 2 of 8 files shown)


internal/testutils/helpers.go
```diff
@@ -13,3 +14,22 @@ func SkipIfNotRunningIntegrationTests(t *testing.T) {
...
+			sourceIdList = append(sourceIdList, app.SourceID)
+		}
+	}
+	for _, sourceID := range sourceIdList {
+		for _, src := range fixtures.TestSourceData {
+			if src.ID == sourceID {
+				sourcesCount++
+			}
+		}
+	}
```
MikelAlejoBR: Regarding this one, wouldn't this work as well?

```go
var sourcesCount int
for _, app := range fixtures.TestApplicationData {
    if app.ApplicationTypeID == appTypeId {
        sourcesCount++
    }
}
```

However by the title of the function I understand that you want to count the sources that have applications with an specific application ID, right? If that's the case the above code would return 3 in the case of a source that has 3 applications with the application type "X". Is that what is intended?

MikelAlejoBR (commented):
Looks good :+1: , I just have a question.

lpichler:
maybe it could be good if you can add few words about what is issue and how are you fixing it, it helps in review ;) thanks!

internal/testutils/helpers.go
petracihalova: nice catch, thanks ... it is exactly as you write ... I wanted to count sources that have apps with specific app type id and because before we had in fixtures "invalid" test data (both applications had same app type and same source id) I wanted to find only unique source ids ... but then I changed fixtures because I needed more test data for other tests and I ended with this function above in some "in progress" state and forgot to change it ... so simply ... yes, you are right, it is nonsense and need to be changed as you suggested 😁 😁 😁

petracihalova:
> maybe it could be good if you can add few words about what is issue and how are you fixing it, it helps in review ;) thanks!

yes, that's a good point ... I'll get better next time ... I added description at the top

internal/testutils/helpers.go
MikelAlejoBR: What I still don't clearly see is if you still want to distinctly count the sources? Because the above function would still return 3 if a source has 3 applications with an specific application type you're looking for.

petracihalova: But 1 source can have only 1 app with 1 specific app type. You cannot have 2 apps with same app type for 1 source. So I count with that data in fixtures are valid in this way. Or you mean that we need check that our test data are valid and we dont 2 apps with same app type for 1 source?

MikelAlejoBR: Well... this might be my misunderstanding. I did remember that I was able to create multiple applications with the same type for the same source, but as you say, this might not make sense and we might need to restrict it somehow. I'm going to ping @lpichler and @lindgrenj6 to help us out here :)

petracihalova: Well there is a story around ... we have ticket for it [RHCLOUD-18437](https://issues.redhat.com/browse/RHCLOUD-18437) and I started to solve it but then @lindgrenj6 told me that we need add sql constraint and because it was before your "migrations" PR was merged, it remained unchanged in backlog

MikelAlejoBR: Aaaha! I see. Thank you for that link :) Everything's clear now.

petracihalova:
I changed my help function to check that 1 source can have only 1 app with specific app type. Now the GetSourcesWithAppType() in helpers returns not count of sources but list with sources so it can be reuse on more places. Now we can add "invalid" application test data into fixtures and this option will be considered in the test.

internal/testutils/helpers.go
```diff
@@ -15,14 +16,36 @@ func SkipIfNotRunningIntegrationTests(t *testing.T) {
...
+		for _, src := range fixtures.TestSourceData {
+			if id == src.ID {
+				out = append(out, src)
+				break
+			}
+		}
+	}
+
+	return out
 }
```
MikelAlejoBR: That's good! I'm going to provide an alternative implementation too using maps which might be slightly more concise. The idea is for you to get comfy around maps too :)

```go
func GetSourcesWithAppType(appTypeId int64) []model.Source {
	// We are only interested in the keys, so we use a "struct{}" as a value so that Go doesn't allocate memory for it.
	var sourceIds = make(map[int64]struct{})
	
	// Find applications with given application type and create
	// list of unique source IDs
	for _, app := range fixtures.TestApplicationData {
		if app.ApplicationTypeID == appTypeId {
			_, ok := sourceIds[app.SourceID]
			if !ok {
				sourceIds[app.SourceID] = struct{}{}
			}
		}
	}

	var sources []model.Source
	// We only grab the key from the map here.
	for id := range sourceIds {
		for _, src := range fixtures.TestSourceData {
			if id == src.ID {
				sources = append(sources, src)
			}
		}
	}

	return sources
}
```

petracihalova: thanks @MikelAlejoBR this is really better

petracihalova:
rebased and conflicts solved

lindgrenj6 (approved):
LGTM! Thanks for fixing this.

PR #217, opened 2022-04-11
Move Job Engine to a new worker pod, using Redis as the backend

lindgrenj6:
Been working on this for a bit - might as well open it now since we're moving forward. This will make it easier to track in the stage ENV.

1. Moves the job worker to its own deployment, which is listening on a redis queue. It pops jobs from there and runs them. 
2. `jobs.Enqueue(jobs.Job)` now just pushes the job onto a queue to be handled later rather than the in-memory channel.
3. Side effect is that I also upgraded our redis library to v8, I had forgotten to upgrade it for the longest time. The biggest API change is that everything takes a context all the way down.

### DIFF
--- a/jobs/client.go
+++ b/jobs/client.go
@@ -1,0 +1,67 @@
+package jobs
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	l "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/RedHatInsights/sources-api-go/redis"
+)
+
+type JobRequest struct {
+	JobName string
+	JobRaw  []byte
+	Job     Job
+}
+
+// implementing binary mashaler/unmarshaler interfaces for redis encoding/decoding.
+func (jr JobRequest) MarshalBinary() (data []byte, err error) {
+	return json.Marshal(&jr)
+}
+func (jr *JobRequest) UnmarshalBinary(data []byte) error {
+	return json.Unmarshal(data, jr)
+}
+
+func (jr *JobRequest) Parse() error {
+	switch jr.JobName {
+	case "SuperkeyDestroyJob":
+		sdj := SuperkeyDestroyJob{}
+		err := json.Unmarshal([]byte(jr.JobRaw), &sdj)
+		if err != nil {
+			return err
+		}
+
+		jr.Job = &sdj
+	case "AsyncDestroyJob":
+		adj := AsyncDestroyJob{}
+		err := json.Unmarshal(jr.JobRaw, &adj)
+		if err != nil {
+			return err
+		}
+
+		jr.Job = &adj
+	default:
+		l.Log.Warnf("Unsupported job: %v", jr.JobName)
+		return fmt.Errorf("unsupported job %v", jr.JobName)
+	}
+
+	l.Log.Debugf("Successfully parsed job %v, args %v", jr.Job.Name(), jr.Job.Arguments())
+
+	return nil
+}
+
+// Throws a `job` on the redis list to be picked up by the worker
+func Enqueue(j Job) {
+	l.Log.Infof("Submitting job %v to redis with %v", j.Name(), j.Arguments())
+
+	req := JobRequest{
+		JobName: j.Name(),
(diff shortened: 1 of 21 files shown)


jobs/worker.go
```diff
@@ -1,46 +1,84 @@
...
+		// something gets queued.
+		//
+		// Once it does succeed - we check to see if the error is just
+		// `redis.Nil`, in which case things are fine and we continue.
+		val, err := redis.Client.BLPop(context.Background(), 0, workQueue).Result()
+		if err != nil {
+			if errors.Is(err, redis.Nil) {
+				l.Log.Debugf("nil value returned  from redis - continuing")
+			} else {
+				l.Log.Warnf("Failed to pop job from queue: %v", err)
```
MikelAlejoBR: I would simply refactor this to `if !errors.Is(err, redis.Nil)`, since we don't care about them, do we?

jobs/client.go
```diff
@@ -0,0 +1,67 @@
...
+}
+func (jr *JobRequest) UnmarshalBinary(data []byte) error {
+	return json.Unmarshal(data, jr)
+}
+
+func (jr *JobRequest) Parse() error {
+	switch jr.JobName {
+	case "SuperkeyDestroyJob":
+		sdj := SuperkeyDestroyJob{}
+		err := json.Unmarshal([]byte(jr.JobRaw), &sdj)
```
MikelAlejoBR, line 30: I think casting the `JobRaw` to `[]byte` is redundant here

jobs/superkey.go
```diff
@@ -11,11 +12,13 @@ import (
 	"github.com/RedHatInsights/sources-api-go/util"
 )
 
+var _ = (Job)(&SuperkeyDestroyJob{})
```
MikelAlejoBR: What is this for?

jobs/worker.go
```diff
@@ -1,46 +1,84 @@
...
+	}
 }
 
-// Package level function to enqueue jobs to the job channel
-func Enqueue(j Job) {
-	ch <- j
+// Run a Job with a delay but in the background so it does _not_ block the
+// calling routine.
+func RunJobAsync(j Job) {
+	l.Log.Infof("Running job asyncronously %v", j.Name())
```
MikelAlejoBR: What about including the ID here? This way if something goes wrong we know exactly which one failed

jobs/worker.go
```diff
@@ -1,46 +1,84 @@
...
+		RunJob(j)
+	}()
 }
 
-// Runs the worker just consuming off of a channel, right now just firing off
-// individual jobs in goroutines.
+// Run a job with a delay in front. This will block the calling thread. If you
+// do not want to block use `RunJobAsync`
+func RunJob(j Job) {
+	l.Log.Infof("Waiting %v", j.Delay())
```
MikelAlejoBR: Same here

MikelAlejoBR (commented):
Looks good! :+1:

jobs/worker.go
lindgrenj6: Ah yeah, great point. We don't really care if it was nil coming back or not.

jobs/client.go
lindgrenj6: Yep good catch that is leftover from when I had the field as a string instead of a []byte.

jobs/superkey.go
lindgrenj6: Basically this will statically validate that the struct implements the interface at compile time. Very nice for making sure the struct continues to implement the interface since compilation will fail if a method is added or changed and the struct does not reflect it right. But I think the compiler will catch it as well - I'll remove it.

MikelAlejoBR (approved):

jobs/superkey.go
MikelAlejoBR: Oooooooh. I'm okay with leaving it, but I would appreciate a "compiler check" comment or something, because I see myself forgetting about this in the future :joy:

lpichler (approved):

lindgrenj6:
Merging after rebase + 2 approvals :grin:

PR #219, opened 2022-04-11
Fix SuperkeyDestroy SubCollectionList call to not send a pointer

lindgrenj6:
I was wondering why this was failing - I had chalked it up to the job deleting out of order before, and I PR'd this in once but must have removed the commits.

1. The Relation Object doesn't recognize `*model.Source`, so it returns an error here: https://github.com/RedHatInsights/sources-api-go/blob/8b58b9d33bd7b292c713d09d0e513380b3aea415/model/relation_object.go#L112
2. That error message is *subsequently erased here, without logging* https://github.com/RedHatInsights/sources-api-go/blob/8b58b9d33bd7b292c713d09d0e513380b3aea415/dao/application_dao.go#L37

So we were getting messages for "source not found" when running the superkey delete job. 

I would like to refactor/remove the RelationObject eventually - it could be replaced by a query such as:
```go
func (a *applicationDaoImpl) ListForSource(sourceID int64, limit int, offset int, filters []util.Filter) ([]m.Application, int64, error) {
	exists, err := GetSourceDao(a.TenantID).Exists(sourceID)
	if !exists || err != nil {
		l.Log.Infof("Source not found %v", sourceID)
		return nil, 0, util.NewErrNotFound("source")
	}

	return a.List(limit, offset, append(filters, util.Filter{Name: "source_id", Value: []string{strconv.FormatInt(sourceID, 10)}}))
}
```
for the simple subcollections that don't involve many to many.

### DIFF
--- a/jobs/superkey.go
+++ b/jobs/superkey.go
@@ -56,7 +56,7 @@ func (sk SuperkeyDestroyJob) sendForSource(id int64) error {
 
 	a := dao.GetApplicationDao(&sk.Tenant)
 
-	apps, _, err := a.SubCollectionList(&m.Source{ID: id}, 100, 0, make([]util.Filter, 0))
+	apps, _, err := a.SubCollectionList(m.Source{ID: id}, 100, 0, make([]util.Filter, 0))
 	if err != nil {
 		return fmt.Errorf("failed to list applications for source: %v", err)
 	}


MikelAlejoBR:
I'm okay with that function, with a small change:

```diff
func (a *applicationDaoImpl) ListForSource(sourceID int64, limit int, offset int, filters []util.Filter) ([]m.Application, int64, error) {
	exists, err := GetSourceDao(a.TenantID).Exists(sourceID)
-	if !exists || err != nil {
+	if !exists {
		l.Log.Infof("Source not found %v", sourceID)
		return nil, 0, util.NewErrNotFound("source")
	}

+	if err != nil {
+		l.Log.Errorf("error checking if the source exists: %s", err)
+		return nil, 0, fmt.Errorf("error checking if the source exists: %s", err)
+	}

	return a.List(limit, offset, append(filters, util.Filter{Name: "source_id", Value: []string{strconv.FormatInt(sourceID, 10)}}))
}
```
Personally I am a huge fan of having one query per sub resource listing just for the sake of having things decoupled: that way the `List` doesn't become too critical and hard to change if something needs to be refactored.

With that said, I would vote "yes" to your proposal anyway. I think we might have two or three functions which would depend on the `List` one, so I guess that it wouldn't be too bad.

lpichler (approved):
LGTM 👍

lpichler:
I also agree with the proposal.

PR #220, opened 2022-04-11
[WIP] Change how subcollections query the db

lindgrenj6:
This is largely an exercise due to #219, it seems like we have a pattern we can follow from all of the subcollections other than `/application_types/:id/sources` and `/sources/:id/application_types`.

Basically the pattern consists of:
- calling `<primaryCollectionDao>.Exists(primaryId)`, which returns a bool, error. either of which indicate it was not found. 
- passing along the filters/limit/offset to the regular `List` method on the DAO but adding a filter for the subcollection. This is still done in one query for all of the endpoints except the 2 previous ones I mentioned.

---

This is kind of one of one of the first "big" refactors I've been wanting to do for a long time. LMK what you think. I think it's easier to follow.


lindgrenj6:
Closing due to age of PR. Maybe we'll get this refactored eventually.

PR #228, opened 2022-04-12
fix: endpoint delete tests failing

MikelAlejoBR:
I forgot to include a "does it exist?" check before trying to cascade delete an endpoint.

### DIFF
--- a/endpoint_handlers.go
+++ b/endpoint_handlers.go
@@ -208,6 +208,16 @@ func EndpointDelete(c echo.Context) error {
 		return util.NewErrBadRequest(err)
 	}
 
+	// Check if the endpoint exists before proceeding.
+	endpointExists, err := endpointDao.Exists(id)
+	if err != nil {
+		return util.NewErrBadRequest(err)
+	}
+
+	if !endpointExists {
+		return util.NewErrNotFound("endpoint")
+	}
+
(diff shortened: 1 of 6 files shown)


lpichler (approved):

endpoint_handlers.go
```diff
@@ -208,6 +208,16 @@ func EndpointDelete(c echo.Context) error {
 		return util.NewErrBadRequest(err)
 	}
 
+	// Check if the endpoint exists before proceeding.
+	endpointExists, err := endpointDao.Exists(id)
+	if err != nil {
+		return util.NewErrBadRequest(err)
```
petracihalova, line 214: I am not sure if I understand it correctly, but if we get some err here, then something went wrong in Exists() method and there is only 1 place where we catching error, but when existing or not existing endpoint id is provided then we get some bool value in endpointExists but variable err = nil, right? so then we get err != nil only when something wrong on server side and that is not 400?

MikelAlejoBR: I think you are right: `err` is not nil when something went wrong with the query itself, so it should be a 500. Maybe I could have returned `err` directly instead of a `NewErrBadRequest`. I opened a PR for the refactor: https://github.com/RedHatInsights/sources-api-go/pull/241

petracihalova: 👍🏻

PR #238, opened 2022-04-14
Improve filename in log messages in nested logger calls

lpichler:
Attribute `filename` in log message is displayed from place where log function(`Log.Error()`) was called.
This is not useful when  log function(`Log.Error()`)  is present in other common function - then we will have location of this 
common function - and it is not useful we don't know from which file the error came from.

This change adds setting for log message called `caller_depth`:

`Log.WithField("caller_depth", 1).Error("error message")`

this ensure that we will have in filename first outer method. If there will be caller_depth=2 - we will have in filename first and second outer method - etc.

Example(id is not number) - check `filename`:
```
GET http://localhost:{{port}}/api/sources/v3.1/sources/XXX
```

before:
```
{"@timestamp":"2022-04-14T21:23:38.499Z","@version":1,"app":"source-api-go","caller":"","filename":"/Users/liborpichler/Projects/sources-api-go/util/error_document.go:22","hostname":"lpichler1-mac","labels":{"app":"source-api-go"},"level":"error","message":"bad request: strconv.ParseInt: parsing \"XXX\": invalid syntax","tags":["source-api-go"]}
```

after:

```
{"@timestamp":"2022-04-14T21:22:14.655Z","@version":1,"app":"source-api-go","caller":"github.com/RedHatInsights/sources-api-go/util.NewErrBadRequest","caller_depth":2,"filename":"/Users/liborpichler/Projects/sources-api-go/source_handlers.go:79; /Users/liborpichler/Projects/sources-api-go/middleware/tenancy.go:70","hostname":"lpichler1-mac","labels":{"app":"source-api-go"},"level":"error","message":"strconv.ParseInt: parsing \"XXX\": invalid syntax","tags":["source-api-go"]}
```

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 	"time"
 
@@ -137,6 +138,10 @@ func (f *CustomLoggerFormatter) Format(entry *logrus.Entry) ([]byte, error) {
 		data["filename"] = fmt.Sprintf("%s:%d", entry.Caller.File, entry.Caller.Line)
 	}
 
+	if entry.Data["caller_depth"] != nil {
+		data["filename"] = filenameFromCurrentCaller(entry)
+	}
+
 	for k, v := range entry.Data {
 		switch v := v.(type) {
 		case error:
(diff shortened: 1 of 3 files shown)


lpichler:
I decided to replace it by series of PR beggining by https://github.com/RedHatInsights/sources-api-go/pull/240

PR #243, opened 2022-04-19
[RHCLOUD-19007] Secret store tests

petracihalova:
[RHCLOUD-19007](https://issues.redhat.com/browse/RHCLOUD-19007)

When you use in env variables the`SECRET_STORE=vault`, the tests fail.

We forgot to set the vault mock in `dao/main_test.go `
+
We need to skip some tests written only for the option when the secret store is database.

### DIFF
--- a/dao/application_authentication_dao_test.go
+++ b/dao/application_authentication_dao_test.go
@@ -69,6 +69,7 @@ func TestDeleteApplicationAuthenticationNotExists(t *testing.T) {
 // test only fetches the application authentications related to the given list of applications.
 func TestApplicationAuthenticationsByApplicationsDatabase(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
+	testutils.SkipIfNotSecretStoreDatabase(t)
 	SwitchSchema("appauthfind")
 
 	// Get all the DAOs we are going to work with.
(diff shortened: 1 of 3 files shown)


MikelAlejoBR (approved):
Nice on! Yeah, I'd like to get on the "unmock Vault" ticket so that we can test everything with a "real" Vault instance... but until then, this is a good quickfix!

petracihalova:
> Nice on! Yeah, I'd like to get on the "unmock Vault" ticket so that we can test everything with a "real" Vault instance... but until then, this is a good quickfix!

Do we have ticket for "unmock Vault"? I am asking because I am working on adding some missing tests that needs improvement of our Vault mock or "real" Vault instance for our tests ... so if you are working on it, it will be better for me to wait for it.

MikelAlejoBR:
> > Nice on! Yeah, I'd like to get on the "unmock Vault" ticket so that we can test everything with a "real" Vault instance... but until then, this is a good quickfix!
> 
> Do we have ticket for "unmock Vault"? I am asking because I am working on adding some missing tests that needs improvement of our Vault mock or "real" Vault instance for our tests ... so if you are working on it, it will be better for me to wait for it.

There is: https://issues.redhat.com/browse/RHCLOUD-18285

But I haven't started yet, so I would add those tests too and then remove the skipping part once we work on the above ticket.

petracihalova:
rebased

lindgrenj6 (approved):

PR #247, opened 2022-04-19
Fix Authentication ID resolver for GraphQL

lindgrenj6:
The query that was failing is this:
```graphql
{
	sources(filter: { name: "id", operation: "eq", value: "138" }) {
		applications {
			application_type_id
			id
			availability_status_error
			availability_status
			paused_at
			authentications {
				id
			}
		}
	}
}
```

which was returning:
```json
{
	"data": {
		"sources": [
			{
				"applications": [
					{
						"application_type_id": "2",
						"id": "245",
						"availability_status_error": "",
						"availability_status": "",
						"paused_at": null,
						"authentications": [
							{
								"id": ""
							}
						]
					}
				]
			}
		]
	}
}
```

Notice the `""` on Authentication ID. This was being caused by the resolver incorrectly pulling the string ID we have on the Authentication struct.

So what I did was force the resolver to be generated via editing gqlgen.yml - ran `make generate` then implemented the check. 

---

After the fix the response looks great:
```json
{
	"data": {
		"sources": [
			{
				"applications": [
					{
						"application_type_id": "2",
						"id": "245",
						"availability_status_error": "",
						"availability_status": "",
						"paused_at": null,
						"authentications": [
							{
								"id": "362"
							}
						]
					}
				]
			}
		]
	}
}
```

### DIFF
--- a/gqlgen.yml
+++ b/gqlgen.yml
@@ -72,3 +72,7 @@ models:
     fields:
       authentications:
         resolver: true
+  Authentication:
+    fields:
+      id:
+        resolver: true
--- a/graph/generated/generated.go
+++ b/graph/generated/generated.go
@@ -124,6 +124,8 @@ type ApplicationResolver interface {
 	TenantID(ctx context.Context, obj *model1.Application) (string, error)
 }
 type AuthenticationResolver interface {
+	ID(ctx context.Context, obj *model1.Authentication) (string, error)
+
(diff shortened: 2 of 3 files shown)


lindgrenj6:
cc @lpichler @MikelAlejoBR @petracihalova 

This is what it looks like when a field isn't quite working right in the graphql implementation. Merging to get the UI working.

PR #248, opened 2022-04-19
Prepare Clowdapp for cutover from rails -> go app

lindgrenj6:
This should mimic the way we're currently pulling the db from app-interface the same way in the ruby app.


lindgrenj6:
cc @lpichler I'm enabling feature flags for us here during the cutover.

syncrou (approved):
LGTM 👍

PR #255, opened 2022-04-20
Add support for feature flags

lpichler:
This adds function `FeatureEnabled(featureFlagName)` and it returns whether feature flag with name `featureFlagName` was set to enable or disable.

There is **env. variable** to enable feature flags for sources api: `FEATURE_FLAGS_SERVICE` and to use current implementation with unleash - it requires value `unleash` (as is set in `clowdapp.yaml`)

This was testes with minikube locally with clowder and also locally with running unleash server.

Testing 
## Locally with running unleash server

Run unleash server:

### 1. Run DB postgres

```
docker run -e POSTGRES_PASSWORD=some_password \
  -e POSTGRES_USER=unleash_user -e POSTGRES_DB=unleash \
  --network unleash --name postgres postgres
```

### 2. Unleash server
```
docker run -p 4242:4242 \
  -e DATABASE_HOST=postgres -e DATABASE_NAME=unleash \
  -e DATABASE_USERNAME=unleash_user -e DATABASE_PASSWORD=some_password \
  -e DATABASE_SSL=false \
  --network unleash unleashorg/unleash-server
```

### 3. GET Api TOKEN Access
Go to http://localhost:4242
l:admin
p:unleash4all
Configure -> API access - and generate here one token for you 

### 4. Set env. variables
```
FEATURE_FLAGS_SERVICE="unleash"
FEATURE_FLAGS_API_TOKEN=<from point 3>
FEATURE_FLAGS_HOST=localhost 
FEATURE_FLAGS_PORT=4242 
FEATURE_FLAGS_SCHEMA=http
```
### 5. Add testing feature code

- add feature toggle `test.feature` in [ui](http://localhost:4242) and add default strategy 
```go
// required imports:
//"fmt"
//	"github.com/RedHatInsights/sources-api-go/service"

// for example to add on line 32
	go func() {
		for {
			fmt.Printf("==== %v ", service.FeatureEnabled("test.feature"))
			time.Sleep(time.Second * 5)
		}
	}()

```

# minikube 

Normal process and you can do only point 5.
+ you need to expose port to access unleash UI:
`kubectl port-forward svc/dev-env-featureflags 4444:4242` and unleash UI will be on `localhost:4444`

# Additional notes

A) We can ensure that unleash is work when find logs message "connection to unleash instance is ready "

B) [unleash-api-go](https://github.com/Unleash/unleash-client-go) is doing polling to store state of feature flags to storage which is like cache for added feature flags in UI. Unleash client as sent regularly also metric data (how often feature flags were queried ). There are constant to set how frequently is state of feature flags sent/metrics are sent.

C) When there is error with unleash instance [unleash-api-go](https://github.com/Unleash/unleash-client-go) just use cache and last know state of feature flags will be used.
- I was trying to add control on error  and with FeatureFlags listener to turn off all flags in case  but we know only when error occurred  and not when unleash service is ready to reinit it back. Possible solution is to add own feature flag(and add own polling here) and use OnSent method from `FeatureFlagListener` as indication that service is ready again.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -34,6 +34,12 @@ type SourcesApiConfig struct {
 	DatabaseUser              string
 	DatabasePassword          string
 	DatabaseName              string
+	FeatureFlagsEnvironment   string
+	FeatureFlagsHost          string
+	FeatureFlagsPort          string
+	FeatureFlagsSchema        string
+	FeatureFlagsAPIToken      string
+	FeatureFlagsService       string
 	CacheHost                 string
 	CachePort                 int
 	CachePassword             string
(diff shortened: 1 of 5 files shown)


MikelAlejoBR (approved):

MikelAlejoBR:
Nice one! :+1:

lindgrenj6 (approved):

PR #256, opened 2022-04-20
Refactor marketplace code

MikelAlejoBR:
Essentially what this PR does is:

1. Moves all the marketplace related code to the "marketplace" package.
3. Removes the "append the marketplace token to authentication's extra field" code from the DAO layer, and puts it in the handlers.
4. Makes it easier to use because now you only need to call `marketplace.SetMarketplaceTokenAuthExtraField(tenantId, auth)` and it takes care of everything for you. Prior to this we had to be initializing structs with the `tenantId` here and there which felt like spaghetti code.
5. Extracts all the interfaces to an "interfaces.go" file at the package level, so that it is easier to see what is the package about.
6. Logs the error messages as detailed as possible, and returns a generic error message to the user. This is because usually the user can't do much about a "could not parse int from str" message.


lpichler:
This PR could be also more PRs :)

what is reason for `wait-to-merge` ?

MikelAlejoBR:
> This PR could be also more PRs :)

It looks like I'm doing a lot, but mostly is noise from moving code. It could be summarized as two main points:

1. Refactor marketplace code.
2. Move the "setting the marketplace token in the `extra.marketplace` field" from the DAO layer to the handlers.

> what is reason for `wait-to-merge` ?

It had the label because it depended on a couple of PRs but they're already merged. I forgot to remove the label :label:

marketplace/interfaces.go
```diff
@@ -0,0 +1,21 @@
+package marketplace
+
+import "net/http"
+
+// HttpClient abstracts away the client to be used in the GetToken function, and allows mocking it easily for the
+// tests.
+type HttpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
```
lindgrenj6: since we're refactoring this - would you be ok with using a request with a context rather than the custom http client? it removes the need for the interface + getter methods.

http.NewRequest -> http.NewRequestWithContext

I did it for rbac here: https://github.com/RedHatInsights/sources-api-go/blob/97c2a6316d83f7ee2363c7a911fca6d2e72f3e30/middleware/authorization.go#L120-L121

--- 
I totally get it if not - just a thought.

lindgrenj6 (approved):
This looks good - I just had a thought about the custom http interface.

lpichler (approved):

marketplace/interfaces.go
MikelAlejoBR: It seems there is a way to "mock" the HTTP server with `httptest`: https://stackoverflow.com/a/51227660. Let me try that to see if I can remove the interface...

MikelAlejoBR: Done!

marketplace/marketplace_token_provider.go
```diff
@@ -50,21 +31,27 @@ func (mtp *MarketplaceTokenProvider) RequestToken() (*BearerToken, error) {
...
+
+		return nil, errors.New("could not properly prepare to reach the marketplace")
 	}
 
 	// Set the proper headers to accept JSON, and let the server know we're sending urlencoded data.
 	request.Header.Add("Accept", "application/json")
 	request.Header.Add("Content-Type", "application/x-www-form-urlencoded")
 
-	client := GetHttpClient()
+	client := http.Client{Timeout: 3 * time.Second}
```
lindgrenj6: I guess maybe I wasn't clear - rather than having a custom HTTP client I was thinking of just using `http.DefaultClient.Do()`, but then up where `http.NewRequest` is called we can just use `http.NewRequestWithContext`. 

like so: 
```diff
diff --git a/marketplace/marketplace_token_provider.go b/marketplace/marketplace_token_provider.go
index 4bb19c1..d04bd1e 100644
--- a/marketplace/marketplace_token_provider.go
+++ b/marketplace/marketplace_token_provider.go
@@ -1,6 +1,7 @@
 package marketplace
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -43,7 +44,12 @@ func (mtp *MarketplaceTokenProvider) RequestToken() (*BearerToken, error) {
 	data.Set("apikey", *mtp.ApiKey)
 	data.Set("grant_type", "urn:ibm:params:oauth:grant-type:apikey")
 
-	request, err := http.NewRequest(
+	// set up a context to control the timeout of the request
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	request, err := http.NewRequestWithContext(
+		ctx,
 		"POST",
 		config.Get().MarketplaceHost+"/api-security/om-auth/cloud/token",
 		strings.NewReader(data.Encode()),
@@ -57,8 +63,7 @@ func (mtp *MarketplaceTokenProvider) RequestToken() (*BearerToken, error) {
 	request.Header.Add("Accept", "application/json")
 	request.Header.Add("Content-Type", "application/x-www-form-urlencoded")
 
-	client := GetHttpClient()
-	response, err := client.Do(request)
+	response, err := http.DefaultClient.Do(request)
 	if err != nil {
 		return nil, fmt.Errorf("could not perform the request to the marketplace: %s", err)
 	}
```

MikelAlejoBR: Sure! Done.

PR #262, opened 2022-04-20
Add Smoke Tests to PR Check

lindgrenj6:
Just adding in the IQE smoke tests to the go side since the svc names match now.

It worked like a charm! The builds will take a bit longer now and they will rely on the ephemeral environment, BUT we now have this job running:

1. Unit tests ala `go test ./...`
2. Integration tests ala `go test --integration ./...`
3. Smoke tests via the IQE test framework (which I assume we're free to add more there, we could do crazy things like validate kafka messages even)

All in all I'm glad this worked just as a drop in.


MikelAlejoBR (commented):
This is exciting! So it runs the smoke tests for sources? That's great!!!!!

MikelAlejoBR (approved):

lpichler (approved):

PR #265, opened 2022-04-21
[RHCLOUD-19051] Tests fix

petracihalova:
In some tests we are changing secret store (database, vault) = we change original value from environment variables and we dont change it back (in the end of test)

in this PR:
- the original value of secret store is saved in affected tests and changed back as last step in the test
- mock daos updated for `GetById()` and `Delete()` authentication methods because for database we need to compare given ID with `auth.DbID` and for vault with `auth.iD`
- some tests written for database are now skipped when secret store is vault

### Links:

- [RHCLOUD-19051](https://issues.redhat.com/browse/RHCLOUD-19051)

### DIFF
--- a/authentication_handlers_test.go
+++ b/authentication_handlers_test.go
@@ -93,6 +93,7 @@ func TestAuthenticationList(t *testing.T) {
 
 func TestAuthenticationGet(t *testing.T) {
 	var id string
+	originalSecretStore := conf.SecretStore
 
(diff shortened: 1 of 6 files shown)


MikelAlejoBR (approved):

petracihalova:
after rebase the tests started failing ... so I added new commit with fix

lindgrenj6 (approved):

PR #270, opened 2022-04-22
[RHCLOUD-17795] fix: change the marketplace authentication type to "marketplace-token"

MikelAlejoBR:
Instead of having a generic "token" authentication type name which could clash with some other authentication type, by giving a unique name to these authentications processing them will be straightforward.

This was already fixed in https://github.com/RedHatInsights/sources-api/pull/475, but I forgot to bring it over to the rewrite.

## Links

[[RHCLOUD-17795]](https://issues.redhat.com/browse/RHCLOUD-17795)

### DIFF
--- a/dao/seeds/source_types.yml
+++ b/dao/seeds/source_types.yml
@@ -385,7 +385,7 @@ rh-marketplace:
   vendor: Red Hat
   schema:
     authentication:
-    - type: token
+    - type: marketplace-token
       name: API Token
       fields:
       - component: text-field


dao/seeds/source_types.yml
```diff
@@ -385,20 +385,20 @@ rh-marketplace:
...
+          name: authentication.authtype
+          hideField: true
+          initializeOnMount: true
+          initialValue: token
+        - component: text-field
+          name: authentication.password
+          label: API Token
+          isRequired: true
+          validate:
+          - type: required
```
lindgrenj6: I _think_ these would need to have the 2 extra spaces removed - that way it's an array of objects.

MikelAlejoBR: Right, thanks!

lpichler (approved):

lindgrenj6 (approved):

PR #274, opened 2022-04-25
Remove created_at from authentication responses

lpichler:
Also this fixes bug in UI,

when user is trying to submit for with editing authentication:
L44Uqx0VEk

(please ignore part with developer console)

it is because (apparently) UI takes attributes from GET authentication endpoint to define attributes of body for PATCH request and one such attribute is `created_at` which we don't support anymore and thanks to that we refuse it as _unknown_ attribute during update.

PR #276, opened 2022-04-26
Remove version from authentication_response struct

lpichler:
There is similar [PR](https://github.com/RedHatInsights/sources-api-go/pull/274) as version doesn't exist in DB and as it looks it was intended for vault.


MikelAlejoBR (approved):

lindgrenj6:
We'll have to add some code around these with the vault config in the future - but merging for now.

PR #277, opened 2022-04-26
Add log type to logs

lpichler:
Currently we have let's say 4 types of log messages:

**default**:
`logging.Log.Warnf("Received %v, exiting", s)`

**echo**:
`c.Logger().Infof("tenant: %v", *sourcesDB.Tenant())`
Typical usage in echo handlers method, `c` is `echo.Context`.

**sql**:
this type is used sql query is being logged.

**request**:
this is also related to echo but it is for logging api requests.

it helps for filtering in kibana.


lindgrenj6 (approved):
Nice, I've been just filtering on `.rows` for sql logs, `.uri` for middleware logs, and everything else is "regular" logs

PR #278, opened 2022-04-26
[WIP][RHCLOUD-18283] Endpoint tests

petracihalova:
This PR is about adding missing test for `EndpointListAuthetications()` and about fixing related code with main goal to have tests that pass when secret store is database and vault.
- special fixtures for `TestEndpointDelete()` added (previous test manipulate with test data which I needed for different test)
- `TestAuthenticationEdit()` updated (the test failed with secret store = vault)
- test for `EndpointListAuthentications() `created (first version only for secret store = database)
- fixed logic around start and end index in keys array (related with limit, offset) in `List()` method in `authentication_dao.go`
- updated `Read()` and `List() `method in mock vault and added new auths fixtures - so skip condition in several tests was removed because now these tests are working for secret store = vault
- `authFromVault()` from `authentication_dao.go` updated = added parsing tenant id from secret (needed for `TestAuthenticationEdit()` tests - this test started failing again with secret store = vault)
- in the end created missing bad request and not found missing tests for `EndpointListAuthentications()`

**JIRA:** [RHCLOUD-18283](https://issues.redhat.com/browse/RHCLOUD-18283)

PR #286, opened 2022-04-28
[RHCLOUD-19157] Add auths fixtures

petracihalova:
when I added some new test data for authentications (I need them mainly in next PR), several tests started failing
- `TestSourceListAuthentications() `and `TestAuthenticationList()` need only small fix
- but `TestAuthenticationListOffsetAndLimit()` from authentication_dao_test.go failed because in List() method we had this logic for calculation end index for "keys" array

```
end := 0
if limit > len(keys) {
    end = len(keys)
} else {
    end = limit
}
if offset > limit {
    return nil, int64(len(keys)), nil
}
```
and later in code we iterate

`for _, val := range keys[offset:end] ... `
 
but there is mistake in logic for e.g. this situation

10 keys, limit = 5, offset = 3.   =>.    end = 5
we iterate in range [3 : 5] => we get 2 auths, but expected is 5

------------

after fixing this logic the `TestAuthenticationListOffsetAndLimit()` is working but only for secret store = database
so I added there temporary condition for skipping test when secret store is vault and this will be removed after changes in next PR (I will try to not forget and add here link to the PR)

This change was originally part of https://github.com/RedHatInsights/sources-api-go/pull/278

**LINKS:** [RHCLOUD-19157](https://issues.redhat.com/browse/RHCLOUD-19157)


MikelAlejoBR (approved):

dao/authentication_dao.go
```diff
@@ -95,22 +95,23 @@ func (a *authenticationDaoImpl) List(limit int, offset int, filters []util.Filte
 		return nil, 0, err
 	}
 
-	// Handle if the limit is longer than the keys available
-	end := 0
-	if limit > len(keys) {
+	// Set start and end index and take into account value of offset and limit
+	start := offset
```
lpichler, line 99: Nice naming!

lpichler (approved):

PR #299, opened 2022-05-03
[RHCLOUD-19197] feature: "provisioning" app type

MikelAlejoBR:
They are starting to integrate with us and they need this application
type.

## Links

[[RHCLOUD-19197]](https://issues.redhat.com/browse/RHCLOUD-19197)

### DIFF
--- a/dao/seeds/application_types.yml
+++ b/dao/seeds/application_types.yml
@@ -69,3 +69,11 @@
   supported_authentication_types:
     satellite:
     - receptor_node
+"/insights/platform/provisioning":
+  display_name: Provisioning
+  dependent_applications: []
+  supported_source_types:
+    - amazon
+  supported_authentication_types:
+    amazon:
+      - provisioning-arn
--- a/dao/seeds/source_types.yml
+++ b/dao/seeds/source_types.yml
@@ -66,6 +66,24 @@ amazon:
           pattern: "^arnaws:.*"
         - type: min-length
           threshold: 10
+    - type: provisioning-arn
+      name: Provisioning's ARN
+      fields:
+        - component: text-field
+          name: authentication.authtype
+          hideField: true
+          initializeOnMount: true
+          initialValue: provisioning-arn
+        - name: authentication.username
+          component: text-field
+          label: ARN
+          isRequired: true
+          validate:
+            - type: required
+            - type: pattern
+              pattern: "^arnaws:.*"
+            - type: min-length
+              threshold: 10
     endpoint:
       hidden: true
       fields:


dao/seeds/source_types.yml
```diff
@@ -66,6 +66,24 @@ amazon:
...
+        - name: authentication.username
+          component: text-field
+          label: ARN
+          isRequired: true
+          validate:
+            - type: required
+            - type: pattern
+              pattern: "^arnaws:.*"
+            - type: min-length
+              threshold: 10
```
lindgrenj6, line 86: These just need moved 2 spaces to the left - fields is an array of objects.

MikelAlejoBR: I've reduced the indentation in two spaces. I was confused by this, though:

> fields is an array of objects.

I think that these are equivalent?

```yaml
array:
  - one
  - two
  - three
```

```yaml
array:
- one
- two
- three
```

lindgrenj6: They might be treated the same by some parsers but not by others - thats where it gets fun. For example parsing the second one in ruby works, but the first one fails. For the k8s Go library it works, and for gopkg.in/yaml v2 & v3 it works.


tl;dr: yaml sucks. but the whitespace usually denotes an object rather than an array.

MikelAlejoBR: Wow, TIL. [Yaml Spec V1.2 §10.1.1.2.](https://yaml.org/spec/1.2.2/#10112-generic-sequence)

petracihalova (approved):

lpichler (approved):

dao/seeds/application_types.yml
```diff
@@ -69,3 +69,11 @@
   supported_authentication_types:
     satellite:
     - receptor_node
+"/insights/platform/provisioning-service":
+  display_name: Provisioning service
```
ezr-ondrej: I'd rather go with Provisioning only, but unimportat ATM :)
```suggestion
  display_name: Provisioning
```

dao/seeds/source_types.yml
```diff
@@ -66,6 +66,24 @@ amazon:
           pattern: "^arnaws:.*"
         - type: min-length
           threshold: 10
+    - type: provisioning-service-arn
+      name: Provisioning service's ARN
```
lzap: Just not to get this skipped, it's also here:

```suggestion
      name: Provisioning's ARN
```

MikelAlejoBR:
All right thank you for your inputs @ezr-ondrej and @lzap . I've modified the name. @lpichler, I think this is ready :)

lpichler (approved):

PR #300, opened 2022-05-03
Change authentication DB DAO to list all authentications for a source…

lindgrenj6:
… rather than the directly attached authentications

So this was something I must have missed in the code review for the auth DB DAO. For `/sources/:id/authentications` we want to fetch all of the authentications related to that source not just the ones that are directly attached. We have a nice field `source_id` which we set on create so the relation is there.


MikelAlejoBR (commented):
Ah, I see :eyes:

lpichler (approved):

PR #301, opened 2022-05-04
[RHCLOUD-19204] fix: race condition when creating tenants

MikelAlejoBR:
If two simultaneous requests contain a non-existent tenant, the back end
creates two rows for the tenant table with the same "external tenant" or
"org id" values.

Depending on which created row gets picked up by the "get or create
tenant" function, some of the resources would be inaccessible.

Adding the unique constraint for both columns from the tenants table
fixes the issue.

## Links

[[RHCLOUD-19204]](https://issues.redhat.com/browse/RHCLOUD-19204)

### DIFF
--- a/db/migrations/20220504080000_add_tenants_external_tenant_org_id_unique_constraints.go
+++ b/db/migrations/20220504080000_add_tenants_external_tenant_org_id_unique_constraints.go
@@ -1,0 +1,67 @@
+package migrations
+
+import (
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// AddTenantsExternalTenantOrgIdUnique adds a unique constraint to the "external tenant" and "org id" columns to avoid
+// a race condition that enabled having multiple rows of data sharing the same "external tenant" or "org id" values.
+// This was caused by sending two parallel requests having a non-existent tenant on the database, which would create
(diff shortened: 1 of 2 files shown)


petracihalova (approved):

db/migrations/20220504080000_add_tenants_external_tenant_org_id_unique_constraints.go
```diff
@@ -0,0 +1,69 @@
+package migrations
+
+import (
+	_ "embed"
```
lindgrenj6: We probably don't need this import - I think it was leftover from the original :|

lindgrenj6 (approved):
One teeny comment that doesn't matter - but would be a nice fix.

lpichler:
My the only concern is whether data in stage(or prod) env are suitable for those constraints.

db/migrations/20220504080000_add_tenants_external_tenant_org_id_unique_constraints.go
MikelAlejoBR: Import removed!

MikelAlejoBR:
> My the only concern is whether data in stage(or prod) env are suitable for those constraints.

I'll try to submit a query through AppInterface for both stage and prod to see if we could run into any issues with this.

MikelAlejoBR:
Closing in favor of https://github.com/RedHatInsights/sources-api-go/pull/352

PR #303, opened 2022-05-04
[RHCLOUD-19207] Refactor migrations improve redis connection error message

MikelAlejoBR:
Changes the signature of the migration function, since we were only returning the error to log it afterwards. I thought that, in that case, it would be better to spare that return and to log the error directly.

The PR also improves the error messages, so that it is clearer why migrations are failing. For example, when Redis was not present, the error message would look like as if the database were not running, when in reality the "connection error" came from the Redis client.

## Links

[[RHCLOUD-19207]](https://issues.redhat.com/browse/RHCLOUD-19207)

### DIFF
--- a/dao/db.go
+++ b/dao/db.go
@@ -94,10 +94,7 @@ func Init() {
 	rawDB.SetMaxOpenConns(20)
 
 	// Perform database migrations.
-	err = migrations.Migrate(DB)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


lindgrenj6 (approved):

lpichler (approved):

lpichler:
Nice!

PR #309, opened 2022-05-09
Refactor: use rxhidentity everywhere

MikelAlejoBR:
The goal of this PR is to remove all traces of `c.Set("accountNumber")` from the code, so that we end up using the standardized `xRhIdentity.Identity´ struct instead. This allows us to:

* Support both EBS account numbers and OrgIds easily, and be able to pass them using a single struct.
* Open the window for any other data that may identify the user, because if the identity struct gets updated, we won't have to change any of the function's signatures.

### DIFF
--- a/service/notifications.go
+++ b/service/notifications.go
@@ -8,6 +8,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/kafka"
 	l "github.com/RedHatInsights/sources-api-go/logger"
 	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
 var NotificationProducer Notifier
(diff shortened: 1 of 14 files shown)


lindgrenj6 (approved):

service/notifications.go
```diff
@@ -93,7 +95,8 @@ func (producer *AvailabilityStatusNotifier) EmitAvailabilityStatusNotification(a
 		Application: application,
 		EventType:   statusEventType,
 		Timestamp:   time.Now().Format(time.RFC3339),
-		AccountID:   accountNumber,
+		AccountID:   id.AccountNumber,
+		OrgId:       id.OrgID,
```
lpichler, line 99: I think it is not allowed to have org id here: https://core-platform-apps.pages.redhat.com/notifications-docs/dev/user-guide/send-notification.html for now. I will try to get more info.

MikelAlejoBR: I see. I've removed the "OrgId" from the payload and kept only the "AccountNumber"

lpichler: Currently they accept it now: https://coreos.slack.com/archives/C01UC7NRV19/p1652199785586619 but they ignore it - so we can leave it as you have in last change. thanks

lpichler (approved):

service/notifications.go
MikelAlejoBR: All right, I've reverted the code so that it includes the "OrgId" now as well.

PR #310, opened 2022-05-10
[RHCLOUD-19245] Fix logic for list authentications for app auth

petracihalova:
in method ListForApplicationAuthentication() we list authentications for given application authentication ID => app auth is relation between 1 app and 1 auth => if app auth exists, the "list" contains only 1 authentication

we need:
- get application authentication object (if exists),
- get authentication ID from this app auth,
- return this authentication in list

**JIRA:** [RHCLOUD-19245](https://issues.redhat.com/browse/RHCLOUD-19245)

### DIFF
--- a/dao/authentication_db_dao.go
+++ b/dao/authentication_db_dao.go
@@ -167,58 +167,29 @@ func (add *authenticationDaoDbImpl) ListForApplication(applicationID int64, limi
 	return authentications, count, nil
 }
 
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (approved):

lpichler (approved):

lpichler:
CI failure is unrelated.

PR #311, opened 2022-05-10
[RHCLOUD-19271] feature: remove the "Catalog" application from sources

MikelAlejoBR:
It seems like this application will no longer be used, so we've been asked to remove it. The related "ansible-tower" source can be removed as well, since there are no other applications compatible with that source.

## Links

[[RHCLOUD-19271]](https://issues.redhat.com/browse/RHCLOUD-19271)

### DIFF
--- a/dao/seeds/application_types.yml
+++ b/dao/seeds/application_types.yml
@@ -20,14 +20,6 @@
     quay:
       - quay-encrypted-password
 
-"/insights/platform/catalog":
-  display_name: Automation Services Catalog
-  dependent_applications: []
-  supported_source_types:
-  - ansible-tower
-  supported_authentication_types:
-    ansible-tower:
-    - username_password
 "/insights/platform/cost-management":
   display_name: Cost Management
   dependent_applications: []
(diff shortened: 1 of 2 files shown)


lpichler (approved):

lindgrenj6 (approved):

lindgrenj6:
One question - when we delete this all of the sources + applications will be destroyed out of the DB. Are we ok with that happening on prod? I assume yes. cc @rvsia

MikelAlejoBR:
> One question - when we delete this all of the sources + applications will be destroyed out of the DB. Are we ok with that happening on prod? I assume yes. cc @rvsia

I don't know if the question is for me, but I assumed it was okay. Either that or I misinterpreted the ticket...

petracihalova (approved):

MikelAlejoBR:
Hey all,

I've confirmed that it is safe to remove catalog from our codebase. So it is safe to merge this.

lindgrenj6 (approved):
:wave:

PR #312, opened 2022-05-10
Return 400 instead of 500 on a bad ID on superkey-related delete requ…

lindgrenj6:
…ests

This was causing a 500 to return on DELETE requests to applications and I assume sources. 

EDIT:
also fixing some linter errors - not sure why they didn't get caught before.

### DIFF
--- a/internal/testutils/templates/bad_request_template.go
+++ b/internal/testutils/templates/bad_request_template.go
@@ -2,7 +2,6 @@ package templates
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http/httptest"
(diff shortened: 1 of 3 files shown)


lindgrenj6:
@petracihalova also don't use `fmt.Error(fmt.Sprintf(...))`, there is `fmt.Errorf(...)` that does the same thing - I'm not sure why the linter didn't catch it before.

MikelAlejoBR (approved):

petracihalova (approved):

lpichler (approved):

PR #325, opened 2022-05-13
Close Kafka Producer after producing in order to not leak memory

lindgrenj6:
:zap: *promoting to prod exposed a memory leak!*

After letting the pods marinate in prod for a while we approached over 500MB ram usage, which is kind of nuts considering we only ever hit ~50MB on stage. After watching the graphs for a while it looked like a memory leak. So I took some time to profile it and deduce where it is.

----

In order to profile, first thing you need to do is set up the profiling exporter, with this teeny diff: 

<details>
<summary>Profiling Diff</summary>

```diff
diff --git a/main.go b/main.go
index e654195..100b703 100644
--- a/main.go
+++ b/main.go
@@ -18,6 +18,8 @@ import (
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	_ "net/http/pprof"
 )
 
 var conf = config.Get()
@@ -105,6 +107,10 @@ func runServer(shutdown chan struct{}) {
 		}
 	}()
 
+	go func() {
+		e.Logger.Fatal(http.ListenAndServe("localhost:6060", nil))
+	}()
+
 	// wait for the shutdown signal to come
 	<-shutdown
 	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)

```
</details>

From here just load the app somehow. I have a [little tool](https://gist.github.com/lindgrenj6/51c40e82d5a58f2e1327886a7f4330e4) that basically spams bulk create a bunch - and from here it was really easy to see the memory leakage happening. 

---

But loading is useless without being able to actually see the results! I followed this guide: https://jvns.ca/blog/2017/09/24/profiling-go-with-pprof/

the tl;dr: is you add the code above, then run this command to generate the report after loading the server: 
`go tool pprof -web http://localhost:6060/debug/pprof/heap`
which will open the SVG in your browser. You can also generate pngs by changing `-web` to `-png`.

Here are the results....
**BEFORE:**
before

**AFTER:**
after

After loading them both for ~2-3 minutes with that program...before we were using 150+mb, with most of that living in the kafka producer. Afterwards we're at a whopping 15M-30M depending when the GC runs.

This change should bubble out to the availability status listener too since that pod was leaking too albeit slower. 

\# TODO:
- [x] fix tests

### DIFF
--- a/application_handlers_test.go
+++ b/application_handlers_test.go
@@ -1032,7 +1032,7 @@ func TestPauseApplicationPauseRaiseEventCheck(t *testing.T) {
 	// We back up the producer so that we can restore it once the test has finished. This way we don't mess up with
 	// other tests that may need to raise events.
 	backupProducer := service.Producer
-	service.Producer = events.EventStreamProducer{Sender: MockSender{}}
+	service.Producer = func() events.Sender { return events.EventStreamProducer{Sender: MockSender{}} }
 
 	var applicationRaiseEventCallCount int
 	raiseEventFunc = func(eventType string, payload []byte, headers []kafka.Header) error {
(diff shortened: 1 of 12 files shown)


MikelAlejoBR (approved):
Nice catch, and thanks for the write-up! It's nice to learn about these tools :grinning:

lpichler (approved):
Great 👏

PR #326, opened 2022-05-16
[RHCLOUD-18437] Add unique index into applications table for source id + app type id + tenant id

petracihalova:
One source can't have 2 applications with same app type - to prevent this, we need to add a unique index to the db.
**JIRA:** [RHCLOUD-18437](https://issues.redhat.com/browse/RHCLOUD-18437)

But first we need to be sure that in the stage and production environment we don't have this data.
We can merge this migration after [RHCLOUD-19355](https://issues.redhat.com/browse/RHCLOUD-19355).


db/migrations/20220510112000_add_application_constraint.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+func AddApplicationConstraint() *gormigrate.Migration {
+	type Application struct {
+		gorm.Model
+		SourceID          int64 `gorm:"uniqueIndex:index_applications_on_source_id_tenant_id_app_type_id"`
+		ApplicationTypeID int64 `gorm:"uniqueIndex:index_applications_on_source_id_tenant_id_app_type_id"`
+		TenantID          int64 `gorm:"uniqueIndex:index_applications_on_source_id_tenant_id_app_type_id"`
```
MikelAlejoBR: I would say to stick to how [Postgres names its indexes](https://stackoverflow.com/a/4108266). So that would make the index name `applications_app_type_id_source_id_tenant_id_idx`.

MikelAlejoBR (approved):
LGTM!

db/migrations/20220510112000_add_application_constraint.go
petracihalova: well I followed what we already have in db ... e.g. these are our existing indexes names in applications table 
image
 
My first option was `"idx_xxxxx"` because that is how it is used in [documentation](https://gorm.io/docs/indexes.html), but then I decided to let it in same way how we have other existing indexes

MikelAlejoBR: I see, but there's a couple of things:

* We still have that naming because the PR which transforms all those names is still pending to be merged: https://github.com/RedHatInsights/sources-api-go/pull/175
* I chose Postgres naming because that's standard to any user that uses Postgres, instead of framework specific like it was with Active Record, or like it is with Gorm.

Anyway we can merge as it is, and I'll update my PR afterwards.

petracihalova: I see 😁  thanks for the reminder, i completely forgot we have these pending PRs ... I will set same naming here 👍🏻

MikelAlejoBR (approved):

petracihalova:
after rebase the tests started fail (because they violates the new unique index added in the migration) ... this PR contains migration itself and fix for these tests

petracihalova:
we cannot merge this PR before we solve the duplicates we have on prod

lindgrenj6 (approved):
Dupes have been eliminated on stage/prod.

PR #330, opened 2022-05-17
Get feature flag config only from ACG_CONFIG

lpichler:
Remove dependency from `UNLEASH_URL` and `UNLEASH_TOKEN`. It is not needed as we confirmed it on stage.

`ACG_CONFIG` contains bearer token and we can use it - as there is benefit - url is nicely parsed already and we don't need to parse it and check if it is http and https and then add port to url.


lindgrenj6 (approved):

lindgrenj6:
the pr check for that job has pods stuck in pending so the eph env must be out of resources, merging anyway since it is just a config change.

PR #332, opened 2022-05-17
Add RetryCreate scheduled job that re-sends the create messages if the application fails to go available

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-19154

This PR adds the "retry mechanism" to Sources.

---

Basically it does a few things:
1. An application is created, retry_counter defaults to 0
2. The Job runs every 2 minutes, and looks for any applications that have NOT been available yet AND have been retried less than `RETRY_MAX` which I am leaving at 5 initially
3. For those applications, it loads them up from the db, checks to make sure that the applicationType is "opted in" to retrying, and then re-raises the various create messages in order
4. ...then increments the retry_counter

The goal here is that for applications that want this feature (currently cloud-meter) we will retry validating the source 5 times before giving up. 

I also want to note that the first thing the job does is **select all applications that went available and the retry counter was less than RETRY_MAX and updates the retry counter so they do not get retried once they're available**.

### DIFF
--- a/jobs/retry_create.go
+++ b/jobs/retry_create.go
@@ -1,0 +1,146 @@
+package jobs
+
+import (
+	"time"
+
+	"github.com/RedHatInsights/sources-api-go/dao"
+	"github.com/RedHatInsights/sources-api-go/kafka"
+	l "github.com/RedHatInsights/sources-api-go/logger"
+	h "github.com/RedHatInsights/sources-api-go/middleware/headers"
+	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/service"
+	"github.com/RedHatInsights/sources-api-go/util"
+	"gorm.io/gorm"
+)
+
+const (
+	RETRY_MAX        = 5
+	RECORD_AGE_LIMIT = -1 * 30 * time.Minute
+)
+
+type RetryCreateJob struct{}
+
+// implementing the interface - but these functions aren't really needed since
+// this is a scheduled job.
+func (r *RetryCreateJob) Delay() time.Duration              { return 0 }
+func (r *RetryCreateJob) Arguments() map[string]interface{} { return map[string]interface{}{} }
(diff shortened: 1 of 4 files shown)


jobs/retry_create.go
```diff
@@ -0,0 +1,137 @@
...
+		// find all applications that are unavailable/null/empty
+		// AND
+		// created_at less than 30m ago
+		// AND
+		// retry counter less than configured amount
+		apps := make([]m.Application, 0)
+		result = tx.Debug().
+			Select("id", "tenant_id", "application_type_id").
+			Model(&m.Application{}).
+			Where("availability_status != ? ", m.Available).
```
lpichler, line 54: it did not take when records with availability_status are null.
We can use `IS DISTINCT FROM`.
```
SELECT id,tenant_id,application_type_id, availability_status FROM applications WHERE (availability_status IS DISTINCT FROM 'available' ) AND created_at > '2022-06-01 15:15:10';
```

jobs/retry_create.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		l.Log.Infof("Found %v Applications that need to be retried", result.RowsAffected)
+
+		// resend messages
+		for i := range apps {
+			go resendCreateMessages(apps[i].ID, apps[i].ApplicationTypeID, apps[i].TenantID)
+		}
+
+		// increment retry counter on the apps we sent create messages for
```
lpichler, line 74: but we also increment retry counter when we didn't send create message :) (application not opted in, failures)

lpichler:
LGTM,  I had minor comments.

jobs/retry_create.go
MikelAlejoBR: I guess it's okay to increment the retry counter even if we encountered an error when attempting to resend a message, right? Like "I tried sending a message, but I failed again. Add +1 to the retry counter".

lpichler:
I am closing this in favour of https://github.com/RedHatInsights/sources-api-go/pull/362

PR #334, opened 2022-05-18
Add tests for status propagation from Application to Source

lpichler:
Added tests for https://github.com/RedHatInsights/sources-api-go/pull/294

### DIFF
--- a/application_handlers_test.go
+++ b/application_handlers_test.go
@@ -734,6 +734,36 @@ func TestApplicationEdit(t *testing.T) {
 	backupNotificationProducer := service.NotificationProducer
 	service.NotificationProducer = &mocks.MockAvailabilityStatusNotificationProducer{}
 
+	sourceDao := dao.GetSourceDao(&fixtures.TestTenantData[0].Id)
+
+	fixtureSource := m.Source{
+		SourceTypeID:       fixtures.TestSourceTypeData[0].Id,
+		TenantID:           fixtures.TestTenantData[0].Id,
+		AvailabilityStatus: "",
+	}
+
+	err := sourceDao.Create(&fixtureSource)
+	if err != nil {
+		t.Error(err)
+	}
+
+	sourceID := fixtureSource.ID
+
+	applicationDao := dao.GetApplicationDao(&fixtures.TestTenantData[0].Id)
+	fixtureApp := m.Application{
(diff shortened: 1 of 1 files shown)


application_handlers_test.go
```diff
@@ -618,6 +658,30 @@ func TestApplicationEdit(t *testing.T) {
...
+		t.Errorf("Invalid email notification data:")
+	}
+
+	src, err = sourceDao.GetById(&application.SourceID)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if src.AvailabilityStatus == previousSourceStatus {
+		t.Errorf("Invalid email notification data:")
```
MikelAlejoBR: I think you forgot the colons, ":", from the error messages

MikelAlejoBR (approved):
LGTM!

application_handlers_test.go
```diff
@@ -618,6 +658,30 @@ func TestApplicationEdit(t *testing.T) {
...
+	if application.AvailabilityStatus == previousApplicationStatus {
+		t.Errorf("Invalid email notification data:")
+	}
+
+	src, err = sourceDao.GetById(&application.SourceID)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if src.AvailabilityStatus == previousSourceStatus {
```
petracihalova, line 848: in this condition, the `src.AvailabilityStatus` has value "available" and `previousSourceStatus` has value "", so we check that `"available" == "" `... but in this case we know that expected is value "available" ... from my perspective it is maybe better to have here something like ... what you think?

`if src.AvailabilityStatus != "available" {`

petracihalova (approved):
LGTM, I have only tiny comment

application_handlers_test.go
lindgrenj6: @lpichler If you fix this message I'll merge this.

application_handlers_test.go
lpichler: yes make sense but here we are testing whether status was changed and this condition cover only destination value after this update operation - I would leave as it is if you are ok with it anyway good insight

application_handlers_test.go
lpichler: @lindgrenj6 message was fixed and also I changed words.

PR #336, opened 2022-05-25
[RHCLOUD-19370] fix: don't panic when feature flags are missing from the Clowdapp JSON

MikelAlejoBR:
The performance team is having issues deploying sources on their cluster, because Clowder doesn't put the feature flags key on the JSON that the back end reads from. This causes the back end to panic when they try to boot it up, which is blocking them.

## Links

[[RHCLOUD-19370]](https://issues.redhat.com/browse/RHCLOUD-19370)


MikelAlejoBR:
Closing in favor of https://github.com/RedHatInsights/sources-api-go/pull/337

PR #338, opened 2022-05-26
[RHCLOUD-13364] refactor: make the spec match the back end's endpoints

MikelAlejoBR:
The things that have been done:

- Fix license URL to use the "https" scheme.
- Add the ability to authorize using headers.
- Include all the environments: local, staging and production.
- Improve some of the descriptions for the endpoints.
- "400" returns for many endpoints that lacked them.
- Update the models for the payloads and responses, which weren't 100%
  accurate.
- Add a missing "/application_authentications/{id}/authentications" endpoint.

## Links

[[RHCLOUD-9245]](https://issues.redhat.com/browse/RHCLOUD-9245)
[[RHCLOUD-18477]](https://issues.redhat.com/browse/RHCLOUD-18477)
[[RHCLOUD-13364]](https://issues.redhat.com/browse/RHCLOUD-13364)

### DIFF
--- a/public/openapi-3-v3.1.json
+++ b/public/openapi-3-v3.1.json
@@ -1326,7 +1326,6 @@
         "operationId": "getRhcConnectionSources",
         "summary": "List sources related to an RHC Connection",
         "tags": [
-          "sources",
           "rhc-connections"
         ]
       }
@@ -1375,8 +1374,7 @@
         "operationId": "getSourcesRhcConnection",
         "summary": "List RHC Connections related to a source",
         "tags": [
-          "sources",
-          "rhc-connections"
+          "sources"
         ]
       }
     },
--- a/public/openapi-3-v3.1.json
+++ b/public/openapi-3-v3.1.json
@@ -500,14 +500,7 @@
             "$ref": "#/components/responses/BadRequest"
           },
           "404": {
-            "description": "Not found",
-            "content": {
-              "application/json": {
-                "schema": {
-                  "$ref": "#/components/schemas/ErrorNotFound"
-                }
-              }
-            }
+            "$ref": "#/components/responses/NotFound"
           }
         },
         "tags": [
@@ -1609,14 +1602,7 @@
             "$ref": "#/components/responses/BadRequest"
           },
           "404": {
-            "description": "The specified source was not found",
-            "content": {
-              "application/json": {
-                "schema": {
-                  "$ref": "#/components/schemas/ErrorNotFound"
-                }
-              }
-            }
+            "$ref": "#/components/responses/NotFound"
           }
         },
         "tags": [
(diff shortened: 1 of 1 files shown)


MikelAlejoBR:
The SonarQube failure is related to this PR: https://github.com/RedHatInsights/sources-api-go/pull/335 .

lpichler:
I am not asking to split up it - but there are too much changes in this PR. 

There is good article - https://smallbusinessprogramming.com/optimal-pull-request-size/

lpichler (approved):

public/openapi-3-v3.1.json
```diff
@@ -139,90 +154,82 @@
...
-        "operationId": "deleteApplicationAuthentication",
-        "description": "Deletes a ApplicationAuthentication object",
+      }
+    },
+    "/applications_authentications/{id}/authentications": {
+      "get": {
+        "summary": "List authentications for the given application authentication",
+        "operationId": "listApplicationAuthenticationsAuthentications",
+        "description": "Returns an array of authentications",
         "parameters": [
```
petracihalova, line 203: I think in parameters is missing something like 
```
{
"$ref": "#/components/parameters/ID"
}
```

public/openapi-3-v3.1.json
```diff
@@ -489,6 +493,9 @@
               }
             }
           },
+          "400": {
+            "$ref": "#/components/responses/BadRequest"
+          },
           "404": {
```
petracihalova, line 502: here can be the one line version to have it same as is on other places
`"$ref": "#/components/responses/NotFound"`

I noticed that the "more lines version" is several times in document ... I think it is not wrong but maybe we can have it in same way for all occurrences

public/openapi-3-v3.1.json
```diff
@@ -1487,9 +1372,6 @@
           "400": {
             "$ref": "#/components/responses/BadRequest"
           },
-          "401": {
-            "$ref": "#/components/responses/Unauthorized"
-          },
           "404": {
             "$ref": "#/components/responses/NotFound"
           }
```
petracihalova, line 1372: in the end of `/sources/{id}/rhc_connections` and` /rhc_connections/{id}/sources` we have tags with values "sources" and "rhc-connections" ... it causes that both methods are displayed in endpoints related with both of this groups (example on picture) ... but we follow this approach only here and not for others endpoints ... I think, it should be same in whole spec

image

petracihalova (changes requested):
Nice job Mikel, this is really lot of changes. I have only tiny comments.

public/openapi-3-v3.1.json
MikelAlejoBR: Done, thanks!

public/openapi-3-v3.1.json
MikelAlejoBR: Right, I must have missed those. Fixed!

public/openapi-3-v3.1.json
MikelAlejoBR: All right, I've removed the redundant tags!

petracihalova (approved):
Awesome! Let's merge it 😁

PR #340, opened 2022-06-03
[RHCLOUD-19742] Tenant check in methods GetById() and SubCollectionList()

petracihalova:
We need to check a tenant id in GetById() and SubCollectionList() methods

**JIRA:** [RHCLOUD-19742](https://issues.redhat.com/browse/RHCLOUD-19742)


syncrou (commented):
Great to see this!

lindgrenj6 (approved):
Thanks for this!

PR #345, opened 2022-06-07
remove not needed condition in TestSourceListAuthentications()

petracihalova:
without JIRA ticket

we don't need removed condition:
- there is no need to create new fixtures because they are created in beginning of tests (main_test.go, line 58: `database.CreateFixtures()` )
- "else" part causes that secret store is changed to "vault" when we run unit tests (even if secret store is database)
- in case we run integration tests and `config.IsVaultOn() = true` the conf.SecretStore is already "vault" and we change it in this "else" statement into "vault" again


petracihalova:
failing tests are not related with code change (confirmed with Diego)

lpichler (approved):

MikelAlejoBR (approved):

PR #346, opened 2022-06-07
Extract code to get application from bulk create struct

lpichler:
This function will be used in upcoming PR about supporting user based sources in bulk creation.


MikelAlejoBR (approved):

petracihalova (approved):
👍🏻

PR #347, opened 2022-06-07
[RHCLOUD-19374] Migration - added user_id column into related tables

petracihalova:
For **Single User based Sources** we need to add column "user_id" into related tables and field "UserID" into related models.

The user_id columns will contain NULL value or some concrete user_id which exists in table users => the type of UserID fields in models have to be pointer (to support NULL values) and there must be foreign key between user_id and id from table "users".

Test DB had to be updated to support new table and new UserID fields.

**JIRA:** [RHCLOUD-19374](https://issues.redhat.com/browse/RHCLOUD-19374)

### DIFF
--- a/db/migrations/20220607140800_add_user_id_column.go
+++ b/db/migrations/20220607140800_add_user_id_column.go
@@ -1,0 +1,131 @@
+package migrations
+
+import (
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// AddUserIdColumnIntoTables adds into tables "sources", "applications", "authentications"
+// and "application_authentications" a new column "user_id" and creates in each table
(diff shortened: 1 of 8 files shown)


db/migrations/20220607140800_add_user_id_column.go
```diff
@@ -0,0 +1,128 @@
...
+	"gorm.io/gorm"
+)
+
+func AddUserIdColumnIntoTables() *gormigrate.Migration {
+	type User struct {
+		ID int64
+	}
+
+	type Source struct {
+		UserID *int64
```
lpichler, line 18: ~~UserID is *string (everywhere on Source, Application, AA, Auth)~~

petracihalova:
rebased & conflict fixed

MikelAlejoBR (approved):
LGTM!

petracihalova:
rebased & conflicts solved & added function comment

PR #348, opened 2022-06-07
Create users in bulk create request for user resource ownership application types

lpichler:
### What ?
When you create source (+related objects) with `/bulk_create` - it adds record (`user_id`, `tenant_id`) to table `users`.

### Required conditions in bulk create request
- application type for the sources has `ApplicationType#ResourceOwnership` column set to `user`.(URO Application Type = User Resource Ownership Application Type)
- env. variable `RESOURCE_OWNERSHIP` is set to `user`. (set to empty string now to disable this feature)
- user_id of user has to be in header

### How ?

We need to determine from request and headers previously mentioned points. 
There is used new struct + its function to handle it:

```go
type UserResource struct {
 	SourceNames           []string
 	ApplicationTypesNames []string
 	User                  *User
        ResourceOwnership     string
 }
```

`ApplicationTypesNames` -  array to gather URO application types names from request  
`SourceNames` - array to gather source names from request which are tied with  "URO Application Type" through application.
`User` - user from header
`ResourceOwnership ` - it is for enabling/disabling this feature (value `user` means enabled)

There are couple function with ensures correct populating this struct or also functions like `UserOwnershipActive` to ask whether all required conditions are active so we can create users.

### Test
x-rh-identity:

```json
{
   "identity":{
      "account_number":"${ACCOUNT_NUMBER}",
      "user":{
         "is_active":true,
         "locale":"en_US",
         "is_org_admin":true,
         "username":"Test",
         "email":"test@test.com",
         "first_name":"Test",
         "user_id":"55555555",
         "last_name":"User",
         "is_internal":true
      }
   },
   "internal":{
      "org_id":"000001"
   }
}
```

```
POST http://localhost:{{port}}/api/sources/v3.1/bulk_create

{
  "sources": [
    {
      "name": "A10",
      "source_type_name": "bitbucket"
    }
  ],
  "applications": [
    {
      "source_name": "A10",
      "application_type_name": "app-studio"
    }
  ],
  "authentications": [
    {
      "resource_name": "app-studio",
      "username": "myUser",
      "password": "myPass"
    }
  ]
}
```

### Links
part of https://issues.redhat.com/browse/RHCLOUD-19196

### DIFF
--- a/bulk_handlers_test.go
+++ b/bulk_handlers_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"testing"
 
+	"github.com/RedHatInsights/sources-api-go/dao"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/fixtures"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/request"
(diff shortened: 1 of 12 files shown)


dao/user_dao.go
```diff
@@ -0,0 +1,57 @@
+package dao
+
+import (
+	m "github.com/RedHatInsights/sources-api-go/model"
+)
+
+var GetUserDao func(*int64) UserDao
+
+// getDefaultRhcConnectionDao gets the default DAO implementation which will have the given tenant ID.
```
petracihalova: typo in comment ... should be "getDefaultUserDao gets the ..."

bulk_handlers_test.go
```diff
@@ -40,3 +41,185 @@ func TestBulkCreateMissingSourceType(t *testing.T) {
...
+	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
+	c.Set("identity", &identity.XRHID{Identity: identityHeader})
+
+	err = BulkCreate(c)
+	if err != nil {
+		t.Error(err)
+	}
+
+	var users []m.User
+	err = dao.DB.Model(&m.User{}).Find(&users).Error
```
petracihalova: I think here we can have problem in the future when we add fixtures for users into tests or if different test creates new users ... now the test relies on the fact that in "users" table is only 1 record (and this one record is "our" user)

bulk_handlers_test.go
```diff
@@ -40,3 +41,185 @@ func TestBulkCreateMissingSourceType(t *testing.T) {
...
+
+	if len(users) != 1 {
+		t.Errorf("1 user expected instead of %d", len(users))
+	}
+
+	if users[0].UserID != testUserId {
+		t.Errorf("expected userid is %s instead of %s", testUserId, users[0].UserID)
+	}
+
+	err = dao.DB.Model(&m.User{}).Delete(&users).Error
```
petracihalova: here all users are deleted, not only the one created in this test

bulk_handlers_test.go
```diff
@@ -40,3 +41,185 @@ func TestBulkCreateMissingSourceType(t *testing.T) {
...
+
+	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
+	c.Set("identity", &identity.XRHID{Identity: identityHeader})
+
+	err = BulkCreate(c)
+	if err != nil {
+		t.Error(err)
+	}
+
+	var users []m.User
```
petracihalova, line 144: similar here ... the test assumes that table users is empty, but it may be different ... of course we can change it in these tests after we add some conflicting tests in the future

bulk_handlers_test.go
```diff
@@ -40,3 +41,185 @@ func TestBulkCreateMissingSourceType(t *testing.T) {
...
+	err = dao.DB.Model(&m.User{}).Find(&users).Error
```
petracihalova: similar here

petracihalova (commented):
⭐️ looks really good ⭐️
I added some comments to tests, not sure if we need to change it, at least we need to be aware of it

model/user_resource.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (ur *UserResource) UserOwnershipActive() bool {
+	return len(ur.SourceNames) > 0 &&
+		len(ur.ApplicationTypesNames) > 0 &&
+		ur.userIDPresent() &&
+		ur.userResourceOwnership()
+}
+
+func (ur *UserResource) userIDPresent() bool {
+	return ur.User.UserID != "" && ur.User != nil
```
MikelAlejoBR: I think you have to swap the conditions here, as otherwise the first one might throw a pointer dereferencing error. If you swap them and `ur.User == nil`, the second one won't be checked.

model/user_resource.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (ur *UserResource) userResourceOwnership() bool {
+	return ur.ResourceOwnership == "user"
```
MikelAlejoBR: Maybe this string could be a `const`

MikelAlejoBR (approved):
LGTM! Good work!

There is one question I have regarding store source names and application type names: I am thinking that we might have `users_sources` and `users_applications` join tables in the future to see which resources belong to which users, so, shouldn't we be picking up the IDs from those?

lpichler:
> LGTM! Good work!
> 
> There is one question I have regarding store source names and application type names: I am thinking that we might have `users_sources` and `users_applications` join tables in the future to see which resources belong to which users, so, shouldn't we be picking up the IDs from those?

It is possible scenario in future and we don't know how request for next steps will look like - so I would prefer to have most simplest querying/DB modeling as we can. Anyway this PR is just about creating users into users tables.


@petracihalova @MikelAlejoBR Thanks for review 

and this ready for re-review.

cc @syncrou

model/user_resource.go
```diff
@@ -0,0 +1,30 @@
...
+		ur.userIDPresent() &&
+		ur.userResourceOwnership()
+}
+
+func (ur *UserResource) userIDPresent() bool {
+	return ur.User != nil && ur.User.UserID != ""
+}
+
+func (ur *UserResource) userResourceOwnership() bool {
+	return ur.ResourceOwnership == UserOwnership
```
petracihalova, line 31: I know that variable UserOwnership is in model/application_type.go, but maybe it will be better to have it in model/user_resource.go ... but this is only my subjective opinion

petracihalova (approved):
👍🏻 LGTM

MikelAlejoBR (approved):
LGTM again!



>> LGTM! Good work!
>>
>> There is one question I have regarding store source names and application type names: I am thinking that we might have users_sources and users_applications join tables in the future to see which resources belong to which users, so, shouldn't we be picking up the IDs from those?
>
> It is possible scenario in future and we don't know how request for next steps will look like - so I would prefer to have most simplest querying/DB modeling as we can. Anyway this PR is just about creating users into users tables.

Understood, makes sense. Thank you for the explanation!

lpichler:
@syncrou can we merge this as we have 2 approvals ?

PR #349, opened 2022-06-08
[RHCLOUD-19378] feature: "NULL" on "external_tenant" and "org_id" columns by default

MikelAlejoBR:
The unique indexes on Postgres don't count "NULL" values as duplicates,
but empty strings do.

Since our "Tenant" model uses non-pointer values, whenever we receive an
identity header without an "AccountNumber" or an "OrgId" number, we are
inserting empty strings instead of null values.

This change would prevent that by inserting "NULL" values in the
database whenever we receive empty values.

## Links


lpichler (approved):

petracihalova (approved):
👍🏻

PR #350, opened 2022-06-08
[RHCLOUD-19378] feature: migration to make empty "external_tenants" and "org_ids" null

MikelAlejoBR:
The empty values on the "external_tenants" and "org_id" columns are
considered when applying a unique index or a unique constraint, so we
need to make sure that the empty values are replaced by "NULL", since
"NULL" is not considered when applying such constraints.

## Links


lpichler (approved):

petracihalova (approved):
LGTM

PR #351, opened 2022-06-08
[RHCLOUD-19378] Feature migration fix duplicated tenants subresources

MikelAlejoBR:
This migration grabs all the duplicated tenants —the ones that have the
same "external_tenant" or "org_id" column values— and picks the first
one as the main one.

Then, for the other duplicates of this tenant, it updates all the
subresources' —application authentications, applications, authentications,
endpoints, sources and source rhc connections— tenant id to the main tenant.

Finally, it updates the duplicated tenants with an easily identifiable
value for the "external_tenant" and "org_id" columns.

## Links

### DIFF
--- a/db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
+++ b/db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
@@ -1,0 +1,255 @@
+package migrations
+
+import (
+	"fmt"
+
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/RedHatInsights/sources-api-go/model"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// duplicateTenantMark is the mark that will be left on the "external_tenant" and "org_id" columns for those tenants
+// that are duplicated. The idea is to prepare the tenants to be able to apply the unique constraint on the columns
+// right after.
+const duplicateTenantMark = "processed-duplicate-tenant-of-%s-%d"
+
+// RemoveDuplicatedTenantIdsOrgIds fetches all the duplicated tenants, makes sure that they are indeed duplicated, and
+// unifies all the subresources in a single tenant before deleting the duplicated one.
+func RemoveDuplicatedTenantIdsOrgIds() *gormigrate.Migration {
+	return &gormigrate.Migration{
+		ID: "20220608090500",
+		Migrate: func(db *gorm.DB) error {
+			logging.Log.Info(`Migration "remove duplicated tenants" started`)
+			defer logging.Log.Info(`Migration "remove duplicated tenants" ended`)
+
+			// Updatable tables contains all the table names which have the "tenant_id" column which must be updated.
+			updatableTables := []string{
+				"application_authentications",
+				"applications",
+				"authentications",
+				"endpoints",
+				"sources",
+				"source_rhc_connections",
+				"users",
+			}
+
+			// This query selects all the tenants which have duplicated "external_tenant" or "org_id" column values.
+			duplicatesSql := `
+				SELECT
+					"t"."id", "t"."name", "t"."description", "t"."external_tenant", "t"."org_id", "t"."created_at", "t"."updated_at"
+				FROM
+					"tenants" AS "t"
+				WHERE
+					"t"."external_tenant" IN (
+						SELECT
+							"texten"."external_tenant"
+						FROM
+							"tenants" AS "texten"
+						GROUP BY
+							"texten"."external_tenant"
+						HAVING
+							count("texten"."external_tenant") > 1
+						AND
+							"texten"."external_tenant" != ''
+					)
+				OR
+					"t"."org_id" IN (
+						SELECT
+							"torgid"."org_id"
+						FROM
+							"tenants" AS "torgid"
+						GROUP BY
+							"torgid"."org_id"
+						HAVING
+							count("torgid"."org_id") > 1
+						AND
+							"torgid"."org_id" != ''
+					)
+			`
+
+			// The idea is that the structure of the following maps will be as follows:
+			//
+			// externalTenantIds:
+			// 	"12345" -> [12, 5123, 123]
+			// 	"4717"	-> [47, 511, 511]
+			//
+			// orgIdIds:
+			// 	"abcde-12345"	-> [51, 98]
+			// 	"dfhea-4918"	-> [5811, 368128]
+			var externalTenantIds = make(map[string][]int64)
+			var orgIdIds = make(map[string][]int64)
+
+			// The migration updates the subresources of a tenant. It first groups all the tenant IDs with their
+			// duplicated "external_tenant" and "org_id". Once it has those, it loops through every table that depends
+			// on the "tenants" table, and updates the "tenant_id" column of the resources from the duplicated tenants
+			// to the main tenant ID. Finally, it updates the "tenants" table by marking the duplicated tenants with
+			// an easily identifiable "external_tenant" and "org_id" values.
+			err := db.Transaction(func(tx *gorm.DB) error {
+				var tenants []model.Tenant
+
+				err := tx.Debug().Raw(duplicatesSql).Scan(&tenants).Error
+				if err != nil {
+					return err
+				}
+
+				// Group all "external_tenant" and "org_id"s into the above defined maps.
+				for _, t := range tenants {
+					// Append the ID to the external tenants map.
+					if t.ExternalTenant != "" {
+						externalTenantIds[t.ExternalTenant] = append(externalTenantIds[t.ExternalTenant], t.Id)
+					}
+
+					// Append the ID to the orgIds map.
+					if t.OrgID != "" {
+						orgIdIds[t.OrgID] = append(orgIdIds[t.OrgID], t.Id)
+					}
+				}
+
+				// mainTenantId is the main tenant ID to which the subresources will be tied to.
+				var mainTenantId int64
+
+				// Start processing the external tenants and grouping them into the main tenant.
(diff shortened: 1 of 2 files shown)


petracihalova (approved):
nice job !!!

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
```diff
@@ -0,0 +1,255 @@
...
+						SELECT
+							"texten"."external_tenant"
+						FROM
+							"tenants" AS "texten"
+						GROUP BY
+							"texten"."external_tenant"
+						HAVING
+							count("texten"."external_tenant") > 1
+						AND
+							"texten"."external_tenant" != ''
```
lpichler: do we also need to cover when `external_tenant` is not `null` ?

MikelAlejoBR: Actually I think we can remove the `"texten"."external_tenant" != ''`, since the previous migration takes care of doing the `'' -> NULL` conversion on the database.

As to your question: no, we don't need that condition. The `AND "external_tenant" != ''` was there because the empty string counted as a "duplicate" as well. So, if we had the following:



id | external_tenant | org_id
-- | -- | --
1 | 12345 | 9876543210
2 |   | abcde
3 |   | deeehg

It would pick up rows with id `2` and `3` as if they were a duplicate, when in reality the only duplicate there is the empty string on the `external_tenant` column. With `NULL`s that doesn't happen because `NULL == NULL -> false`. You can test this yourself:

```sql
INSERT INTO
	"tenants"("external_tenant", "org_id", "created_at", "updated_at")
VALUES
	(NULL, 'abcde', now(), now()),
	(NULL, 'edcba', now(), now()),
	('', 'hello', now(), now()),
	('', 'notsame', now(), now())
```

And then run the query without the `AND` condition:


```sql
SELECT
	"t"."id", "t"."name", "t"."description", "t"."external_tenant", "t"."org_id", "t"."created_at", "t"."updated_at"
FROM
	"tenants" AS "t"
WHERE
	"t"."external_tenant" IN (
		SELECT
			"texten"."external_tenant"
		FROM
			"tenants" AS "texten"
		GROUP BY
			"texten"."external_tenant"
		HAVING
			count("texten"."external_tenant") > 1
	)
OR
	"t"."org_id" IN (
		SELECT
			"torgid"."org_id"
		FROM
			"tenants" AS "torgid"
		GROUP BY
			"torgid"."org_id"
		HAVING
			count("torgid"."org_id") > 1
	)
```

You'll see that the query picks up the two rows with the empty string, when it shouldn't. However, as I said, in the previous migration I take care of the empty strings on both columns, and also I ran a query on AppInterface on the stage and prod environments to check if we have empty strings on either of the columns and we don't.

lpichler: thanks for detailed explanation, now I understand

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
```diff
@@ -0,0 +1,255 @@
...
+				var tenants []model.Tenant
+
+				err := tx.Debug().Raw(duplicatesSql).Scan(&tenants).Error
+				if err != nil {
+					return err
+				}
+
+				// Group all "external_tenant" and "org_id"s into the above defined maps.
+				for _, t := range tenants {
+					// Append the ID to the external tenants map.
```
lpichler, line 96: maybe log message here ? to see which tenants are duplicate

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
MikelAlejoBR: Here are the results of the queries that I ran on both clusters: https://issues.redhat.com/browse/RHCLOUD-19802 (restricted to the team for confidentiality reasons). The MR that I sent to AppInterface checked for `external_tenant = ''`, `"external_tenant" = '' AND "org_id" != ''` and `"external_tenant" = '' AND "org_id" = ''` and yielded zero results, so we should be safe.

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
```diff
@@ -0,0 +1,255 @@
...
+						orgIdIds[t.OrgID] = append(orgIdIds[t.OrgID], t.Id)
+					}
+				}
+
+				// mainTenantId is the main tenant ID to which the subresources will be tied to.
+				var mainTenantId int64
+
+				// Start processing the external tenants and grouping them into the main tenant.
+				for externalTenant, ids := range externalTenantIds {
+					// Grab the first tenant ID as the main one.
```
lpichler, line 114: maybe log message "..processing externalTenant..."

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
```diff
@@ -0,0 +1,255 @@
...
+						}
+
+						duplicateCounter++
+					}
+				}
+
+				// Finish by processing the org IDs to group them on the main tenant. The difference with the above
+				// code is that if an ID has already been processed, it will be skipped to avoid problems.
+				for orgId, ids := range orgIdIds {
+					// Grab the first tenant ID as the main one.
```
lpichler, line 152: maybe log message "..processing orgId..."

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
```diff
@@ -0,0 +1,251 @@
...
+				"endpoints",
+				"sources",
+				"source_rhc_connections",
+				"users",
+			}
+
+			// This query selects all the tenants which have duplicated "external_tenant" or "org_id" column values.
+			duplicatesSql := `
+				SELECT
+					"t"."id", "t"."name", "t"."description", "t"."external_tenant", "t"."org_id", "t"."created_at", "t"."updated_at"
```
lpichler, line 40: do we need all columns here ?

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
MikelAlejoBR: Included, thanks!

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
MikelAlejoBR: Done, thank you!

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
MikelAlejoBR: Thank you :)

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
MikelAlejoBR: You are right, we can probably narrow it down just to the "id", "external_tenant" and "org_id" columns

MikelAlejoBR: Well now that I think about it, I log the entire tenant object with a `"%v#` on the logging statement. So maybe we can leave it like this, or simply a `*`?

lpichler: yes we leave as it is, thanks !

lpichler (approved):

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return err
+	}
+
+	logging.Log.Infof(`Tenant "%d"'s "external_tenant" marked as processed duplicate with mark "%s"`, tenantId, mark)
+
+	return nil
+}
+
+// markExternalTenantAsDuplicate marks the old duplicated tenant as duplicate. It modifies the "org_id" column to add a
```
syncrou, line 245: Small nit, but it looks like the documentation for `markOrgIdAsDuplicate` is actually the copied documenation for `markExternalTenantAsDuplicate`

syncrou: Or the name of the function is - I see the `org_id` reference later on.

syncrou (commented):
Very small documentation nit, but otherwise this looks great.

I have 1 question about the transaction and bulk updates:

Is there any possibility that the bulk transaction could fail based on running so many updates and hence kill the entire process ( ie maybe due to hitting some rate limit on RDS. )

If yes, then obviously no destructive behavior will happen, but the migration will not be able to ever complete.

If yes, then potentially we may need to move to an atomic update where an update is called per tenant id and that is in a transaction. If that transaction fails then all the other successful updates stay as is.  Then when the problem ( that killed the last transaction ) is fixed the migration can pick up where it left off instead of starting from scratch.

I realize there isn't an answer to this now, but thought I would bring it up since I was concerned about it.

syncrou:
All in all ( a part from my concerns, this LGTM 👍 ) I more wanted to discuss hypothetical situation and if we run into it, an option we can used to fix it.

So don't hold merging this due to that concern.

db/migrations/20220608090500_remove_duplicated_tenant_ids_org_ids.go
MikelAlejoBR: Yes, initially I was doing one pass and updating both columns to the same value, but then I thought that doing a second pass with org ids wouldn't hurt, since the redundant updates would essentially be NOPs. That's why we have two functions for the migration. I'll update the docs.

PR #352, opened 2022-06-08
[RHCLOUD-19378] Feature migration unique index external tenant org

MikelAlejoBR:
If two simultaneous requests contain a non-existent tenant, the back end
creates two rows for the tenant table with the same "external tenant" or
"org id" values.

Depending on which created row gets picked up by the "get or create
tenant" function, some of the resources would be inaccessible.

Adding the unique index for both columns from the tenants table
fixes the issue.

I chose to use an index instead of a simple constraint because we
usually look up tenants by their "external_tenant" and "org_id" columns,
so we are going to benefit from having an index on those columns.

## Links

### DIFF
--- a/db/migrations/20220608091000_add_tenants_external_tenant_org_id_unique_index.go
+++ b/db/migrations/20220608091000_add_tenants_external_tenant_org_id_unique_index.go
@@ -1,0 +1,70 @@
+package migrations
+
+import (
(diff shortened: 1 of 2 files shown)


petracihalova (approved):
👍🏻

lpichler (approved):

PR #353, opened 2022-06-09
[RHCLOUD-19374] Migration and related changes for Resource Ownership for App studio application type 

petracihalova:
For **Single User based Sources** we need to distinguish if application type is user based.

This PR:
- adds a migration for new column `resource_ownership` in table `application_types`
- updates the models `Application_Type` and `applicationTypeSeed` - a pointer is used because we want to save NULL values into db in case that the resource ownership is not set for the application type
- updates the function `seedApplicationTypes()` that is responsible for changes from `application_types.yml`

### DIFF
--- a/db/migrations/20220607215900_add_resource_ownership_column.go
+++ b/db/migrations/20220607215900_add_resource_ownership_column.go
@@ -1,0 +1,36 @@
+package migrations
+
+import (
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// AddResouceOwnershipToApplicationTypes adds new column "resource_ownership" into table
+// "application_types".
+func AddResouceOwnershipToApplicationTypes() *gormigrate.Migration {
+	type ApplicationType struct {
+		ResourceOwnership *string `gorm:"type:CHARACTER VARYING"`
+	}
+
+	return &gormigrate.Migration{
+		ID: "20220607215900",
+		Migrate: func(db *gorm.DB) error {
+			logging.Log.Info(`Migration "add_resource_ownership_column" started`)
+			defer logging.Log.Info(`Migration "add_resource_ownership_column" ended`)
+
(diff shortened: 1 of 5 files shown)


petracihalova:
rebased & conflicts solved

db/migrations/20220607215900_add_app_type_user_flag.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// AddUserFlagIntoApplicationTypesTable adds new column "resource_ownership" into table
+// "application_types".
+func AddUserFlagIntoApplicationTypesTable() *gormigrate.Migration {
```
lpichler: Maybe `AddUserFlagIntoApplicationTypesTable`
 -> `AddResouceOwnershipToApplicationTypes` ?

petracihalova:
rebased & changed the name of function as suggested

db/migrations/20220607215900_add_resource_ownership_column.go
```diff
@@ -0,0 +1,36 @@
...
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// AddResouceOwnershipToApplicationTypes adds new column "resource_ownership" into table
+// "application_types".
+func AddResouceOwnershipToApplicationTypes() *gormigrate.Migration {
+	type ApplicationType struct {
+		ResourceOwnership string `gorm:"type:CHARACTER VARYING"`
```
lpichler: should not be here pointer as it is in `ApplicationType` in next commit ?

petracihalova: sharp eye 😁 fixed

MikelAlejoBR (approved):
It looks good to me :+1:

lpichler (approved):

PR #354, opened 2022-06-09
refactor: log the request and responses for RHC connection av check

MikelAlejoBR:
In order to better debug the problem, I would like to have the exact request we're sending and the exact response we're receiving to try to debug what's going on.


lpichler (approved):

syncrou (approved):
LGTM 👍

PR #357, opened 2022-06-10
Refactor param of getDefaultSourceDao and GetSourceDao

lpichler:
PR changes `getDefaultSourceDao` and `GetSourceDao` parameter tenantId to struct use `SourceDaoParams`.

This allows us to extend parameters for `GetSourceDao` function
without changing callers on lot of places.

Lately, this will be use to add `UserID` parameter to struct
`SourceDaoParams` and `UserID` will be passed to `sourceDaoImpl`
struct.

We need `UserID` in dao layer to add support user based filtering.
`UserID` in dao layer allow us to restrict queries to ensure
filtering according to `userID`.

### DIFF
--- a/dao/source_dao.go
+++ b/dao/source_dao.go
@@ -13,18 +13,27 @@ import (
 
 // GetSourceDao is a function definition that can be replaced in runtime in case some other DAO provider is
 // needed.
-var GetSourceDao func(*int64) SourceDao
+var GetSourceDao func(*SourceDaoParams) SourceDao
 
 // getDefaultRhcConnectionDao gets the default DAO implementation which will have the given tenant ID.
-func getDefaultSourceDao(tenantId *int64) SourceDao {
+func getDefaultSourceDao(daoParams *SourceDaoParams) SourceDao {
+	var tenantID *int64
+	if daoParams != nil {
+		tenantID = daoParams.TenantID
+	}
+
 	return &sourceDaoImpl{
-		TenantID: tenantId,
+		TenantID: tenantID,
 	}
 }
 
 // init sets the default DAO implementation so that other packages can request it easily.
 func init() {
 	GetSourceDao = getDefaultSourceDao
+}
+
+type SourceDaoParams struct {
+	TenantID *int64
 }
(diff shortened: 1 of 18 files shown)


dao/source_dao.go
```diff
@@ -13,12 +13,17 @@ import (
...
 // GetSourceDao is a function definition that can be replaced in runtime in case some other DAO provider is
 // needed.
-var GetSourceDao func(*int64) SourceDao
+var GetSourceDao func(*SourceDaoParams) SourceDao
 
 // getDefaultRhcConnectionDao gets the default DAO implementation which will have the given tenant ID.
-func getDefaultSourceDao(tenantId *int64) SourceDao {
+func getDefaultSourceDao(daoParams *SourceDaoParams) SourceDao {
+	var tenantID *int64
+	if daoParams != nil {
```
petracihalova: I am not sure, it I understand correctly what is expected here, but there should be `if daoParams.TenantID != nil` or?

petracihalova (commented):
LGTM, I have only one question

dao/source_dao.go
MikelAlejoBR: The first check is required since `daoParams` is a pointer to `SourceDaoParams`. If you do `daoParams.TenantID != nil` directly you may end up with a pointer dereference error.

I guess the intent of Libor on this one is to pass a `nil` tenant ID if it is missing, and that way the queries won't return anything meaningful. But I don't know.

MikelAlejoBR (approved):
LGTM!

dao/source_dao.go
petracihalova: thanks Mikel, it makes sense to me now

petracihalova (approved):

dao/source_dao.go
lpichler: yes as @MikelAlejoBR said but it is right when we call it with empty struct `&dao.SourceDaoParams{})` we can have dereference error - so I will fix it.

petracihalova (approved):

Issue #356, opened 2022-06-10
User Based Sources


lpichler:
Tracking issue
- [x] **Database Changes**
  - [x] https://github.com/RedHatInsights/sources-api-go/pull/341
  - [x] https://github.com/RedHatInsights/sources-api-go/pull/347
  - [x] https://github.com/RedHatInsights/sources-api-go/pull/353
- [x] Changes for **CREATE**  resources (bulk create)
  - [x] https://github.com/RedHatInsights/sources-api-go/pull/346
  - [x] https://github.com/RedHatInsights/sources-api-go/pull/348
  - [x] https://github.com/RedHatInsights/sources-api-go/pull/412
  - [x] https://github.com/RedHatInsights/sources-api-go/pull/419
- [x] **Filtering** - restrict by user's resource ownership in source api
  - [x] **Source**
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/357
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/395
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/420 
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/453
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/402
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/466
  - [x] **Application**
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/435
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/442
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/469
  - [x] **ApplicationAuthentication**
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/436
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/443
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/456
  - [x] **Authentication**
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/397
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/447
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/470
  - [x] Other helping PRs
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/424
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/426
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/428
    - [x] https://github.com/RedHatInsights/sources-api-go/pull/441
[...]

PR #358, opened 2022-06-10
Access UserID from DAO in source handlers

lpichler:
Issue https://github.com/RedHatInsights/sources-api-go/issues/356

This add to source DAO layer users#id from **only when** there is record in users table which matches users#userid from XRHID header. 

NOTE: When `sourceDaoImpl#UserID` is `nil` it means that user single based sources filtering will be skipped and it will behave like before this change. (this is implement in other PR)

This PR populates userid from header only for sources routes.

### DIFF
--- a/dao/source_dao.go
+++ b/dao/source_dao.go
@@ -17,13 +17,16 @@ var GetSourceDao func(*SourceDaoParams) SourceDao
 
 // getDefaultRhcConnectionDao gets the default DAO implementation which will have the given tenant ID.
 func getDefaultSourceDao(daoParams *SourceDaoParams) SourceDao {
-	var tenantID *int64
+	var tenantID, userID *int64
+
 	if daoParams != nil && daoParams.TenantID != nil {
 		tenantID = daoParams.TenantID
+		userID = daoParams.UserID
 	}
 
 	return &sourceDaoImpl{
 		TenantID: tenantID,
+		UserID:   userID,
 	}
 }
 
@@ -34,10 +37,12 @@ func init() {
 
 type SourceDaoParams struct {
 	TenantID *int64
+	UserID   *int64
 }
 
 type sourceDaoImpl struct {
 	TenantID *int64
+	UserID   *int64
 }
 
 func (s *sourceDaoImpl) SubCollectionList(primaryCollection interface{}, limit, offset int, filters []util.Filter) ([]m.Source, int64, error) {
--- a/routes.go
+++ b/routes.go
@@ -33,11 +33,11 @@ func setupRoutes(e *echo.Echo) {
 		r.POST("/bulk_create", BulkCreate, permissionMiddleware...)
 
 		// Sources
-		r.GET("/sources", SourceList, tenancyWithListMiddleware...)
-		r.GET("/sources/:id", SourceGet, middleware.Tenancy)
+		r.GET("/sources", SourceList, append(tenancyWithListMiddleware, middleware.ResourceOwnership)...)
+		r.GET("/sources/:id", SourceGet, middleware.Tenancy, middleware.ResourceOwnership)
 		r.POST("/sources", SourceCreate, permissionMiddleware...)
-		r.PATCH("/sources/:id", SourceEdit, append(permissionMiddleware, middleware.Notifier)...)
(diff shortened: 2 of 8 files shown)


dao/source_dao.go
```diff
@@ -17,13 +17,16 @@ var GetSourceDao func(*SourceDaoParams) SourceDao
 
 // getDefaultRhcConnectionDao gets the default DAO implementation which will have the given tenant ID.
 func getDefaultSourceDao(daoParams *SourceDaoParams) SourceDao {
-	var tenantID *int64
+	var tenantID, userID *int64
+
 	if daoParams != nil && daoParams.TenantID != nil {
```
lindgrenj6, line 22: I would be ok with this panic-ing. We should _never ever_ pass in a daoParams struct with a nil tenant/user id.

routes.go
```diff
@@ -33,11 +33,11 @@ func setupRoutes(e *echo.Echo) {
 		r.POST("/bulk_create", BulkCreate, permissionMiddleware...)
 
 		// Sources
-		r.GET("/sources", SourceList, tenancyWithListMiddleware...)
-		r.GET("/sources/:id", SourceGet, middleware.Tenancy)
+		r.GET("/sources", SourceList, append(tenancyWithListMiddleware, middleware.ResourceOwnership)...)
```
lindgrenj6: We could probably make a new collection of middleware like `userTenancyWithListMiddleware`.

source_handlers.go
```diff
@@ -18,12 +18,29 @@ var getSourceDao func(c echo.Context) (dao.SourceDao, error)
...
+	}
 
+	userID, err := getUserFromEchoContext(c)
 	if err != nil {
 		return nil, err
 	}
 
-	return dao.GetSourceDao(&dao.SourceDaoParams{TenantID: &tenantId}), nil
+	var user m.User
+	result := dao.DB.Model(&m.User{}).Where("tenant_id = ? and user_id = ?", tenantId, userID).Find(&user)
```
lindgrenj6: This is going to get moved to a `UserDAO` eventually I assume? I also would say this check should happen in a middleware much like `middleware.Tenancy`.

dao/source_dao.go
lpichler: (1) It looks we can't now because of internal source list: https://github.com/RedHatInsights/sources-api-go/blob/ca719c20a38f5ad07adb4e33bdcd39f86a6e3718/internal_handlers.go 

(2) Anyway it would be probably better this way. Previously I was expecting that 
when `SourceDaoParams.Userid == nil` - we are filtering sources record with user_id = NULL
when `SourceDaoParams.Userid != nil` - we are filtering sources record with user_id = SourceDaoParams.Userid

But agree to have always populated `SourceDaoParams.Userid` and I will update it with this approach.

**Question** when we receive status message in status listener - can we rely that in header we will have also `Userid` in headers in received status message or  do we need to confirm this with teams who sends those message ?

source_handlers.go
lpichler: I moved it to UserDAO.

yes but maybe we should move to DAO layer later.

To be able pass always any value of UserID to `dao.GetSourceDao` through `SourceDaoParams` I need to change `SourceDaoParams#UserID` to string and populate it from header.

Thanks to that there will be always `SourceDaoParams#UserID` from header.

With this I can fetch "our" user ID (`users#id`) from users table (if there will be present) and then we do filtering logic - it will be somewhere [here](https://github.com/RedHatInsights/sources-api-go/pull/360/commits/62d126133eb3f9bf4cfd00d720f63e79bc5588ca) which is based on your [work](https://github.com/RedHatInsights/sources-api-go/pull/343).


Does this sound OK to you ?

dao/source_dao.go
lpichler: Update: 

> We should never ever pass in a daoParams struct with a nil tenant/user id.

To reach this we need to use string userid from xhr header - so I need to convert `SourceDaoParams.Userid` to string.

See comment here https://github.com/RedHatInsights/sources-api-go/pull/358#discussion_r918136834

or we can follow original pattern mentioned (2) and allow nil userid in SourceDaoParams struct - then we can do GetIdByUserID(from comment below) check in middleware.

lpichler:
I am closing this as there is replacement https://github.com/RedHatInsights/sources-api-go/pull/420

PR #359, opened 2022-06-10
[WIP] Extract global DB to function

lpichler:
idea pull out from https://github.com/RedHatInsights/sources-api-go/pull/343

This PR puts occurrences `DB.Debug()`  to own function with `sourceDaoImpl ` receiver - ` `getDb`
 according to usage with one params - it is  instance of source - which is usually passed to GORM Model( function

This adds check whether `sourceDaoImpl#TenantID` is nil or not 

In other PR we can use this to add tenancy condition as well condition to check whether userid=nil or not.

### Links
required
- [ ] https://github.com/RedHatInsights/sources-api-go/pull/402
- [x] https://github.com/RedHatInsights/sources-api-go/pull/445

Issue: https://github.com/RedHatInsights/sources-api-go/issues/356


lpichler:
this is replaced by https://github.com/RedHatInsights/sources-api-go/pull/466

PR #368, opened 2022-06-16
[RHCLOUD-19931] fix: bulk create with an endpoint returns a "missing source id" error

MikelAlejoBR:
Since the endpoint validator was running before the source ID for the
endpoint had been set, the validator was complaining about that source
ID missing, which caused bulk create requests with endpoints in the
payload to be rejected.

## Links

[[RHCLOUD-19931]](https://issues.redhat.com/browse/RHCLOUD-19931)


lpichler:
CI failure looks related.

MikelAlejoBR:
> CI failure looks related.

That's right, yes, I was using the wrong fixture on one of the checks. It went all :green_circle: after the fix!

lpichler (approved):

PR #369, opened 2022-06-16
[RHCLOUD-19793] Source handler tests update

petracihalova:
Tests update for source handlers - part I. 
- main goal is to add/update tests to better tenancy testing
- each commit contains one updated or added test (or new helper function)
- changes for other functions in source handler will be part of next PR

handlers covered:
 - SourceListAuthentications()
 - SourceTypeListSource()
 - ApplicationTypeListSource()

**JIRA:** [RHCLOUD-19793](https://issues.redhat.com/browse/RHCLOUD-19793)

### DIFF
--- a/source_handlers_test.go
+++ b/source_handlers_test.go
@@ -332,13 +332,9 @@ func TestSourceTypeSourceSubcollectionList(t *testing.T) {
 		}
 	}
 
-	sourcesBelongToTenant, err := checkAllSourcesBelongToTenant(tenantId, out.Data)
+	err = checkAllSourcesBelongToTenant(tenantId, out.Data)
 	if err != nil {
 		t.Error(err)
-	}
-
-	if !sourcesBelongToTenant {
-		t.Error("the tenant doesn't own all sources")
 	}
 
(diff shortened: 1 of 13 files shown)


source_handlers_test.go
```diff
@@ -205,16 +330,28 @@ func TestSourceTypeSourceSubcollectionList(t *testing.T) {
...
+	}
+
+	sourcesBelongToTenant, err := checkAllSourcesBelongToTenant(tenantId, out.Data)
+	if err != nil {
+		t.Error(err)
+	}
 
+	if !sourcesBelongToTenant {
+		t.Error("the tenant doesn't own all sources")
 	}
```
MikelAlejoBR: Maybe you don't need a boolean value being returned from the `checkAllSourcesBelongToTenant` function. You could simply make it return `err` if they don't belong, and you just have to keep the `if err != nil` check. What do you think?

MikelAlejoBR (approved):
LGTM! Just a minor comment.

source_handlers_test.go
petracihalova: good point, thanks ... solved in new commit

PR #373, opened 2022-06-22
[RHCLOUD-20008] Source pause unpause fix

petracihalova:
In `SourcePause()` handler we get source dao, then we call `Pause()` method and then we load the source's applications. 
```
	sourceDao, err := getSourceDao(c)
	if err != nil {
		return err
	}

	err = sourceDao.Pause(sourceId)
	if err != nil {
	}

	source, err := sourceDao.GetByIdWithPreload(&sourceId, "Applications")
	if err != nil {
		return err
	}
```
But when source id doesn't exist (and we expect Not Found Err) we call `Pause()` method and then when we try to load the source's applications, the not found err is returned.

But I believe we want to know that source doesn't exist BEFORE we call `Pause()` method.

Same for SourceUnpause()

**JIRA:** [RHCLOUD-20008](https://issues.redhat.com/browse/RHCLOUD-20008)

### DIFF
--- a/source_handlers.go
+++ b/source_handlers.go
@@ -491,11 +491,19 @@ func SourceUnpause(c echo.Context) error {
 		return err
 	}
 
+	// Check if the source exists
+	_, err = sourceDao.GetById(&sourceId)
+	if err != nil {
+		return err
+	}
+
+	// Unpause the existing source
 	err = sourceDao.Unpause(sourceId)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lpichler (approved):

petracihalova:
rebased

MikelAlejoBR (approved):
Nice one! LGTM!

PR #376, opened 2022-06-24
Add functionality for Delete allow for Satellite only

lphiri:
@lpichler  @dehort  here is a POC for the behavior we want to have for 

https://issues.redhat.com/browse/RHCLOUD-19895

It needs some more Sources handler tests, but did not get time to figure out how the mocks work (need to set Satellite source_type_id).
Not crazy about putting the "source" specifics in the permission middleware - can probably be done more cleanly, but was looking for a fast path.

cc @syncrou 

Lindani

### DIFF
--- a/middleware/authorization.go
+++ b/middleware/authorization.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/RedHatInsights/rbac-client-go"
@@ -60,7 +61,13 @@ func PermissionCheck(next echo.HandlerFunc) echo.HandlerFunc {
 			// org_admin/rbac/psk
 			if identity.Identity.System != nil {
 				// system-auth only allows GET and POST requests.
-				if c.Request().Method != http.MethodGet && c.Request().Method != http.MethodPost {
+				method := c.Request().Method
+				if method != http.MethodGet && method != http.MethodPost && method != http.MethodDelete {
+					c.Response().Header().Set("Allow", "GET, POST, DELETE")
+					return c.JSON(http.StatusMethodNotAllowed, util.ErrorDoc("Method not allowed", "405"))
+				}
+				// Secondary check for delete - we could move this to middleware
+				if method == http.MethodDelete && !certDeleteAllowed(c) {
 					c.Response().Header().Set("Allow", "GET, POST")
 					return c.JSON(http.StatusMethodNotAllowed, util.ErrorDoc("Method not allowed", "405"))
 				}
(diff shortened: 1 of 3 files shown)


lpichler:
Yes, sources handler tests needs other changes:
1. Add another Satellite SourceType as data which are used for tests: https://github.com/RedHatInsights/sources-api-go/blob/main/internal/testutils/fixtures/source_type.go
2. Add another Source here with Satellite SourceType: https://github.com/RedHatInsights/sources-api-go/blob/main/internal/testutils/fixtures/source_type.go
3. It looks that cache for Source/App type which is used for `dao.Static.GetSourceTypeId("satellite")` needs to be initiated for tests - which probably need to make `typeCache` struct global also with its attributes (sourceType,...). I played with it maybe it can be used: http://pastebin.test.redhat.com/1061108.
With this change can add to each test(or better place) where it is needed:
```go
	tc := dao.TypeCache{}
	tc.SourceTypes = make(map[string]int64)
	tc.SourceTypes = map[string]int64{"satellite": 1} // there is ID defined 1. for satelite.
	dao.Static = tc
```

Also we can skip mocks test and add only integration in follow up PR if we need this quickly.

middleware/authorization.go
```diff
@@ -60,7 +61,13 @@ func PermissionCheck(next echo.HandlerFunc) echo.HandlerFunc {
 			// org_admin/rbac/psk
 			if identity.Identity.System != nil {
 				// system-auth only allows GET and POST requests.
-				if c.Request().Method != http.MethodGet && c.Request().Method != http.MethodPost {
+				method := c.Request().Method
+				if method != http.MethodGet && method != http.MethodPost && method != http.MethodDelete {
+					c.Response().Header().Set("Allow", "GET, POST, DELETE")
+					return c.JSON(http.StatusMethodNotAllowed, util.ErrorDoc("Method not allowed", "405"))
+				}
+				// Secondary check for delete - we could move this to middleware
```
lpichler, line 69: yes and it looks we can add the middleware to `/sources` route in routes.go so we would not need to method `certDeleteAllowed`. We can do it in follow up PR.

lpichler:
Thanks for the PR. 

Looks good to me 👍

dehort (approved):
LGTM!

I think @lpichler is right.  You might be able to simplify things a bit if you moved some of the logic out into its own middleware and use that in routes.go where DELETE on /sources is configured.

lphiri:
Thanks for the review @lpichler and @dehort  .  Lets merge and test this one as is - I will do a cleanup MR split out the delete filter  logic to separate middleware.

syncrou (approved):
LGTM 👍

PR #380, opened 2022-06-28
Fix endpoint matching

lphiri:
@dehort  @syncrou this fixes our matching of the sources path.
Will clean this up when we move to middleware implementation.


syncrou (approved):
LGTM 👍

dehort:
LGTM!

PR #386, opened 2022-06-30
[RHCLOUD-20153] added check if rhc connection exists for given tenant

petracihalova:
this PR adds check if record exists into Delete method for rhc connections

problem is, that table with rhc connections doesn't contain info about tenant and before delete we need to check if the rhc connection belongs to the given tenant

so I added a GetById() call that already contains a query that solves this

**JIRA:** [RHCLOUD-20153](https://issues.redhat.com/browse/RHCLOUD-20153)

### DIFF
--- a/dao/rhc_connection_dao.go
+++ b/dao/rhc_connection_dao.go
@@ -216,6 +216,12 @@ func (s *rhcConnectionDaoImpl) Update(rhcConnection *m.RhcConnection) error {
 func (s *rhcConnectionDaoImpl) Delete(id *int64) (*m.RhcConnection, error) {
 	var rhcConnection m.RhcConnection
 
+	// Check if rhc connection exists for given tenant
+	_, err := s.GetById(id)
+	if err != nil {
+		return nil, err
+	}
+
 	// The foreign key and the "cascade on delete" in the join table takes care of deleting the related
 	// "source_rhc_connection" row.
 	result := DB.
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (commented):
Looks good to me :+1: , although I've checked other DAOs and they seem to simply add a `.Where("tenant_id = ?", a.TenantID)` condition to the query — and then throwing a not found if the affected rows are zero. What do you think of just doing that for the sake of consistency?

petracihalova:
> Looks good to me 👍 , although I've checked other DAOs and they seem to simply add a `.Where("tenant_id = ?", a.TenantID)` condition to the query — and then throwing a not found if the affected rows are zero. What do you think of just doing that for the sake of consistency?

yes, that was my first idea, but table rhc_connections doesn't have column "tenant_id" ... that is why I reused GetById() method for check this because this method contains the query we needed ... another option can be to add tenant_id to this table, but this wouldn't be so simple/short as mine ... maybe another possible solution exists but nothing else came into my mind

petracihalova:
rebased

MikelAlejoBR:
> > Looks good to me +1 , although I've checked other DAOs and they seem to simply add a `.Where("tenant_id = ?", a.TenantID)` condition to the query — and then throwing a not found if the affected rows are zero. What do you think of just doing that for the sake of consistency?
> 
> yes, that was my first idea, but table rhc_connections doesn't have column "tenant_id" ... that is why I reused GetById() method for check this because this method contains the query we needed ... another option can be to add tenant_id to this table, but this wouldn't be so simple/short as mine ... maybe another possible solution exists but nothing else came into my mind

That is right, yeah. Thanks for the explanation!

lpichler (approved):

PR #389, opened 2022-07-01
[RHCLOUD-20192] updated method checkIfPrimaryRecordExists() to check tenancy for primary object

petracihalova:
in application_dao.go `SubCollectionList()` method we check if primary object exists, we use for it method `checkIfPrimaryRecordExists()` in relation_object.go

in the query we are using object without tenant id

consequences are that we are returning the subcollection list of objects even the primary object is not owned by tenant

in this PR is one test for `SubCollectionList()` in application_dao.go and endpoint `/api/sources/v3.1/sources/<id>/applications` but more tests will be added in text PRs (see [RHCLOUD-20138](https://issues.redhat.com/browse/RHCLOUD-20138))

**JIRA:** [RHCLOUD-20192](https://issues.redhat.com/browse/RHCLOUD-20192)

### DIFF
--- a/application_handlers_test.go
+++ b/application_handlers_test.go
@@ -240,6 +240,35 @@ func TestSourceApplicationSubcollectionListBadRequestInvalidFilter(t *testing.T)
 	templates.BadRequestTest(t, rec)
 }
 
+func TestSourceApplicationSubcollectionListTenantNotExists(t *testing.T) {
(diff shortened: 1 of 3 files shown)


MikelAlejoBR (approved):

lpichler (approved):
Thanks!

PR #393, opened 2022-07-07
[RHCLOUD-19727] Make sources support managed Kafka

MikelAlejoBR:
This PR adds support for managed Kafka. Basically what we do is:

* Read the Sasl authentication settings from Clowder.
* Apply those authentication settings to the kafka-go reader/writer.

When calling `.Consumer()` and `.Producer()` we can receive an error now due to the above changes, so that's why I created the handy `.CloseConsumer()` and `.CloseProducer()` functions. I'm welcome to refactoring this if you can think of any other way of doing this.

## Links

[[RHCLOUD-19727]](https://issues.redhat.com/browse/RHCLOUD-19727)


config/config.go
```diff
@@ -83,8 +89,22 @@ func Get() *SourcesApiConfig {
...
+
+		if kafkaBroker.Authtype != nil {
+			options.SetDefault("KafkaSaslEnabled", true)
+
+			caPath, err := cfg.KafkaCa(kafkaBroker)
+			if err != nil {
+				log.Fatalf(`Could not write the Kafka broker's CA to a temporary directory: %s`, err)
+			}
+
+			options.SetDefault("KafkaSaslCaPath", caPath)
```
lpichler, line 102: can we test those cars for emptiness/nil ? + log them in case of when there are empty/nil - it could help with potential debugging.

service/availability_check.go
```diff
@@ -174,9 +174,9 @@ func publishSatelliteMessage(mgr *kafka.Manager, source *m.Source, endpoint *m.E
 		l.Log.Warnf("Failed to produce kafka message for Source %v, error: %v", source.ID, err)
 	}
 
-	err = mgr.Producer().Close()
+	err = mgr.CloseProducer()
 	if err != nil {
-		l.Log.Warnf("Failed to close kafka producer: %v", err)
+		l.Log.Errorf(`unable to close the Kafka producer when publishing a Satellite availability status check message: %s`, err)
```
lpichler, line 179: is this related just to Satellite ?

lpichler:
Looks to good to me I had just few comments - very straightforward and clear implementation 👏

service/availability_check.go
MikelAlejoBR: In this case the log message is in `service/availability_check#publishSatelliteMessaage` function, so that's why I thought that I could specify the error a little bit more.

config/config.go
MikelAlejoBR: Sure, I've added those nil checks and logged the values except for the sensitive credentials. I was doubting if I should simply `Fatal` if any of the `mechanism`, `username` or `password` values are empty, since without those we wouldn't be able to connect to Kafka... Everything would work, but not Kafka.

lpichler (approved):

PR #371, opened 2022-06-17
[RHCLOUD-19793] Source handler tests update3

petracihalova:
Tests update for source handlers - part III. 
(part I. #369, part II. #370)

- main goal is to add/update tests to better tenancy testing
- each commit contains one updated or added test (or new helper function)
- changes for other functions in source handler will be part of next PR

handlers covered:
 - SourceEdit()
 - SourceDelete()
 - SourceCheckAvailability()
 - SourcesRhcConnectionList()

**JIRA:** [RHCLOUD-19793](https://issues.redhat.com/browse/RHCLOUD-19793)

PR #370, opened 2022-06-17
[RHCLOUD-19793] Source handler tests update2

petracihalova:
Tests update for source handlers - part II. 
(part I. #369)

- main goal is to add/update tests to better tenancy testing
- each commit contains one updated or added test (or new helper function)
- changes for other functions in source handler will be part of next PR

handlers covered:
 - SourceList()
 - SourceGet()

**JIRA:** [RHCLOUD-19793](https://issues.redhat.com/browse/RHCLOUD-19793)

PR #372, opened 2022-06-20
[RHCLOUD-19962] Update of tests for SourcePause() and SourceUnpause() handlers

petracihalova:
Tests for SourcePause() and SourceUnpause() were updated to have better test data and better checks of pause/unpause status of source and applications.
+
Added tests for not found and bad request scenarios.

PR will be rebased after merging #371 

**JIRA:** [RHCLOUD-19962](https://issues.redhat.com/browse/RHCLOUD-19962)

PR #394, opened 2022-07-08
[RHCLOUD-19793] Source handler tests update4

petracihalova:
Tests update for source handlers - part IV. 
(part I. #369, part II. #370, part III. #371)

- main goal is to add/update tests to better tenancy testing
- each commit contains one updated or added test (or new helper function)
- changes for other functions in source handler will be part of next PR

handlers covered:
 - SourcePause()
 - SourceUnpause()

**JIRA:** [RHCLOUD-19793](https://issues.redhat.com/browse/RHCLOUD-19793)

### DIFF
--- a/source_handlers_test.go
+++ b/source_handlers_test.go
@@ -1955,6 +1955,35 @@ func TestUnpauseSourceAndItsApplications(t *testing.T) {
 	}
 }
 
+// TestUnpauseSourceAndItsApplicationsInvalidTenant tests that not found is returned
+// when tenant tries to unpause not owned source
+func TestUnpauseSourceAndItsApplicationsInvalidTenant(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
+	// The source is not owned by the tenant
+	tenantId := int64(2)
+	sourceId := int64(1)
+
+	c, rec := request.CreateTestContext(
+		http.MethodPost,
+		"/api/sources/v3.1/sources/1/unpause",
+		nil,
+		map[string]interface{}{
+			"tenantID": tenantId,
+		},
+	)
+
+	c.SetParamNames("source_id")
+	c.SetParamValues(fmt.Sprintf("%d", sourceId))
+
+	notFoundSourceUnpause := ErrorHandlingContext(SourceUnpause)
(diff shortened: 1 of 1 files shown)


petracihalova:
PR is waiting for merge of https://github.com/RedHatInsights/sources-api-go/pull/372

source_handlers_test.go
```diff
@@ -1984,6 +1984,34 @@ func TestUnpauseSourceAndItsApplicationsInvalidTenant(t *testing.T) {
 	templates.NotFoundTest(t, rec)
 }
 
+// TestUnpauseSourceAndItsApplicationsTenantNotExists tests that not found is returned
+// for not existing tenant
+func TestUnpauseSourceAndItsApplicationsTenantNotExists(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
+	tenantId := fixtures.NotExistingTenantId
+	sourceId := int64(1)
```
lpichler, line 1992: (+ tenant id )I know it is on more places but what about to take it from fixtures when we have we defined it ?

petracihalova: well it is for discussion ... I was thinking about if is better to use exact values like `tenantId := int64(1)` or refer to the fixtures like `tenantId := fixtures.TestTenantData[0].Id` .... and I decided to use first option because I was in situation when I added new fixtures data in the middle of existing and then referring to the index of test data in fixtures doesn't work anymore and then was for my easier to know that I want `sourceId = 100` compared with `sourceId = fixtures.TestSourceData[3].ID` ... and that is why I use in tests this attitude .... it is more readable for me and the manipulation with fixtures data will not affect it (or it is for me easier to figure out, what was origin idea) ... I hope that it is a little bit clear what I am trying to explain here and for sure I am open to any different ideas

lpichler: Ok it sounds good

lpichler (approved):

source_handlers_test.go
lpichler: (anyway I would prefer to stop using constants - maybe we can discuss it later as it is wider thing)

lindgrenj6 (approved):

PR #397, opened 2022-07-08
Use AuthenticationDaoParams struct as parameter of GetAuthenticationDao

lpichler:
This change is required to be able pass userid to `GetAuthenticationDao` through `AuthenticationDaoParams` struct.

Similarly it is done for `GetSourceDao`: in https://github.com/RedHatInsights/sources-api-go/pull/357 and https://github.com/RedHatInsights/sources-api-go/pull/358

Issue: https://github.com/RedHatInsights/sources-api-go/issues/356

### DIFF
--- a/dao/authentication_dao.go
+++ b/dao/authentication_dao.go
@@ -25,17 +25,22 @@ const vaultSecretPathFormat = "secret/data/%d/%s_%d_%s"
 
 // GetAuthenticationDao is a function definition that can be replaced in runtime in case some other DAO provider is
 // needed.
-var GetAuthenticationDao func(*int64) AuthenticationDao
+var GetAuthenticationDao func(daoParams *AuthenticationDaoParams) AuthenticationDao
 
 // getDefaultAuthenticationDao gets the default DAO implementation which will have the given tenant ID.
-func getDefaultAuthenticationDao(tenantId *int64) AuthenticationDao {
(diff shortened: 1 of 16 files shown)


petracihalova (approved):

MikelAlejoBR (approved):
LGTM!

dao/authentication_dao.go
```diff
@@ -45,6 +50,10 @@ func init() {
 	GetAuthenticationDao = getDefaultAuthenticationDao
 }
 
+type AuthenticationDaoParams struct {
+	TenantID *int64
+}
```
lindgrenj6, line 55: (non-blocking)

I like the idea of this for now - but I think we can take it a step even further and use the same params struct for all of the DAOs, but also include all of the request info in the struct. 

Having "only" the tenant_id was all we needed back in the day but now the list is:
- tenant_id + account_number + org_id
- user_id + maybe user_id
- (future) groups the user belongs to

Having all of these contained in a struct on the echo context that we can jsut pass aroudn will be nice. It also might be time for us to extend the `echo.Context` interface with our own methods. 

---

This will be something done in the future though - I'm ok with this implementation.

lindgrenj6 (approved):

dao/authentication_dao.go
lpichler: Yes, that is right - it looks that it is even is easier 👍 and I will leave it for other PR. Thanks!

PR #399, opened 2022-07-11
[RHCLOUD-19727] Refactor kafka module

MikelAlejoBR:
Basically:

* The Kafka manager was using the `config.Get()` function to grab the `Sasl` settings. When trying to reuse the `kafka` module in the superkey repository, the `config.Get()` was loading/trying to load all the settings that `sources-api-go` requires to work, instead of only loading the settings required for the superkey application. I've refactored the manager's methods so that it grabs all the settings from its config structs.
* Replaces the manager with a couple of functions that return a Kafka reader and a Kafka writer, which take a `clowder.BrokerConfig` struct as the configuration. This is because the manager seemed to be designed as a singleton, but I realized that all of our readers/writers were getting closed after using them. This is why I thought that it might just be better to use a couple of functions that create the readers/writers.
* Includes the fix to avoid crashes with managed Kafka if no self-signed certificates have been provided to be configured at the TLS level.
* Keeps some of the Kafka manager's functions, like `Close`, `Consume` and `Produce`, since those are very handy functions.
## Links

[[RHCLOUD-19727]](https://issues.redhat.com/browse/RHCLOUD-19727)


main.go
```diff
@@ -25,11 +25,13 @@ var conf = config.Get()
...
+		saslSettings := conf.KafkaSaslSettings
+
 		logging.Log.Infof(`Sasl settings detected, using managed Kafka`)
-		logging.Log.Infof(`Kafka Sasl certificate path: %s`, conf.KafkaSaslCaPath)
-		logging.Log.Infof(`Kafka Sasl mechanism: %s`, conf.KafkaSaslMechanism)
-		logging.Log.Infof(`Kafka Sasl security protocol: %s`, conf.KafkaSaslSecurityProtocol)
+		logging.Log.Infof(`Kafka Sasl certificate path: %s`, saslSettings.CaPath)
+		logging.Log.Infof(`Kafka Sasl mechanism: %s`, saslSettings.Mechanism)
+		logging.Log.Infof(`Kafka Sasl security protocol: %s`, saslSettings.SecurityProtocol)
 	}
```
lindgrenj6: Can we move this to the config package during the parsing? I'm not sure it belongs in main.

internal/events/event_stream_producer.go
```diff
@@ -27,9 +27,7 @@ type EventStreamSender struct {
 func (esp *EventStreamSender) RaiseEvent(eventType string, payload []byte, headers []kafka.Header) error {
 	logging.Log.Debugf("publishing message %v to topic %q...", eventType, EventStreamTopic)
 
-	producerConfig := kafka.ProducerConfig{Topic: config.KafkaTopic(EventStreamTopic)}
-	kafkaConfig := kafka.Config{KafkaBrokers: config.KafkaBrokers, ProducerConfig: producerConfig}
-	kf := &kafka.Manager{Config: kafkaConfig}
+	kf := kafka.GetKafkaProducer(config.KafkaSaslSettings, config.KafkaBrokers, EventStreamTopic)
```
lindgrenj6: Technically for the first 2 args you can just do `config.Get()` in the kafka package - what would you think about just having this use that and only accept the one argument (the topic we need to get a producer for) and for consumer the topic + groupid? 

I know it probably ties it more to our `config/` package - but we're already using the kafkabrokers/sasl settings struct so we may as well use `config.Get()` as well.

lindgrenj6 (commented):
Just a couple small suggestions.

main.go
MikelAlejoBR: I did it in main because the logger is not initialized when parsing the configuration. I thought of manually writing a JSON with `log.Println` but I went for this option. Let me change it...

lindgrenj6: Ah I get it - maybe we should add a `config.Log()` call that we can call at the end of initialization.

internal/events/event_stream_producer.go
MikelAlejoBR: I thought of that, but I was concerned about the reusability. The [superkey-worker uses `source-api-go`'s `kafka/` module](https://github.com/RedHatInsights/sources-superkey-worker/blob/master/main.go#L47-L53). So in that case, if we use `GetKafkaProducer`, it would try to load the `sources-api-go`'s config, which may or may not fail, depending on the env vars that are passed to the superkey worker and the resulting JSON file.

That's why I thought of making the `kafka/` module "independent": if you do want to use it, you need to `go get github.com/RedHatInsights/sources-api-go`, and therefore you'll have access to the `config.KafkaSaslSettings` struct.

Maybe I'm suffering of tunnel vision here...

main.go
MikelAlejoBR: Great idea — I've created a story for it: https://issues.redhat.com/browse/RHCLOUD-20347 . However I feel like this is out of the scope of this PR, so I'm just going to remove those logs from `main`.

internal/events/event_stream_producer.go
lindgrenj6: Oh yeah great point. We can definitely just leave it the way it is then! My only thing is maybe we could have the brokers be the first argument? That's just me thinking through it too much though.

config/config.go
```diff
@@ -13,19 +13,23 @@ import (
...
 var parsedConfig *SourcesApiConfig
 
+// KafkaSaslSettings is a struct containing the settings for authenticating to the managed Kafka instances.
+type KafkaSaslSettings struct {
+	CaPath           string
+	Mechanism        string
+	Password         string
+	SecurityProtocol string
+	Username         string
+}
```
lindgrenj6: Digging through the code some more I'm starting to see why you did this. It's annoying that the ca path is separate from the rest of these...app-common-go's sasl config just contains the bottom four fields :rage: 

I wish we could just use their config :cry: 

---

I almost wonder if we should just use the [`clowder.BrokerConfig`](https://github.com/RedHatInsights/app-common-go/blob/6ca81c147285330f80af3384fa41397abfa9cdaa/pkg/api/v1/types.go#L404-L419) since that has all of the pieces we need. Then we wouldn't have to have our own struct, could just pass in the broker config from our config. 

Curious your thoughts here - it would save us having to parse it out etc.

lpichler (approved):
LGTM 👍

MikelAlejoBR:
All right, so after trying to implement the latest suggestions with the feedback, I gave it a different approach. I've submitted a commit with the thing I've come up with:

* Replaces the manager with a couple of functions that return a Kafka reader and a Kafka writer, which take a `clowder.BrokerConfig` struct as the configuration.

That way you only have to give the `BrokerConfig`, and depending if that `BrokerConfig` has Sasl settings or not, the reader/writer will be created accordingly. I kind of like this refactor, to be honest. Let me know what you think, please.

MikelAlejoBR:
I'm going to request a re-review since I significantly changed the contents of the PR. Sorry about the noise!

lpichler:
it looks that PR contains now 2 major things - one regard to settings and other about replacing kafka manager and title of PR now doesn't cover fully the intent :) (I am not requesting any change regard this)

statuslistener/statuslistener.go
```diff
@@ -45,21 +45,20 @@ func (avs *AvailabilityStatusListener) subscribeToAvailabilityStatus(shutdown ch
...
 	}
 
-	kf := kafka.GetKafkaConsumer(config.KafkaSaslSettings, config.KafkaBrokers, groupID, sourcesStatusTopic)
+	kf, err := kafka.GetReader(&config.KafkaBrokerConfig, groupID, sourcesStatusTopic)
+	if err != nil {
+		l.Log.Errorf(`could not get a Kafka reader: %s`, err)
+		return
+	}
+
+	defer kafka.CloseReader(kf, "suscribe availability status. Shutdown signal received")
```
lpichler: suscribe -> subscribe

MikelAlejoBR: Nice catch! :fishing_pole_and_fish:

MikelAlejoBR:
> it looks that PR contains now 2 major things - one regard to settings and other about replacing kafka manager and title of PR now doesn't cover fully the intent :) (I am not requesting any change regard this)

That is true yeah, I just changed the title! Thank you :smile:

lindgrenj6 (approved):
This looks good. 

@lpichler it looks like one would just always call `kafka.GetReader(clowder.BrokerConfig, topic)` to get a reader, same with writer. Since all of the config settings are in the `clowder.BrokerConfig` struct it looks like it was easier to just make the module a bit more simple that just takes the config, spits out a reader/writer, then have helper functions for producing/consuming. This makes it easier to be used from multiple projects like superkey or another team that asked about using it. 

Reading/writing is largely the same - except instead of instantiating a manager you just create the reader/writer then pass it to the helper functions `kafka.Consume(reader, func(Message))`/`kafka.Produce(writer, Message)`

lpichler:
> This looks good.
> 

Yes this is definitely simpler to use - I asked for examples of usage previously but then I deleted that comment because for us it is probably obvious how to use - question is - whether also for other teams and if this package would deserve own repo - I guess it would be more visible for others.

MikelAlejoBR:
> > This looks good.
> 
> Yes this is definitely simpler to use - I asked for examples of usage previously but then I deleted that comment because for us it is probably obvious how to use - question is - whether also for other teams and if this package would deserve own repo - I guess it would be more visible for others.

To be honest I also thought of extracting this module to its own repository too, since this code is very similar to what the [cloud-connector](https://github.com/RedHatInsights/cloud-connector/pull/156/files) team is doing. It's true though that in the current state of the PR there is a dependency with `v1.BrokerConfig`, so you'd need to use that to be able to use the library.

However I think this is better to do it in a different PR, in my opinion.

lindgrenj6:
I'd take it a step further - rather than an entirely separate repo we could make it a sub-package of the app-common-go repository. Since we're using that to parse the config in the first place then the user(s) could just import the subpackage and use it from there. 

---

I'll talk to the owners of that repo and see if that's something they're looking to do, I know there was an initiative to share middleware/code/etc as much as possible before and this is something that others definitely have been just rolling their own etc.

MikelAlejoBR:
> I'd take it a step further - rather than an entirely separate repo we could make it a sub-package of the app-common-go repository. Since we're using that to parse the config in the first place then the user(s) could just import the subpackage and use it from there.
> 
> I'll talk to the owners of that repo and see if that's something they're looking to do, I know there was an initiative to share middleware/code/etc as much as possible before and this is something that others definitely have been just rolling their own etc.

I like the idea. However I would personally merge this one, update the Superkey worker too with this code, and then if we get the approval from the `app-common-go` team then I don't mind refactoring this yet again. That way we can cross ourselves from the "apps that need to implement support for managed Kafka" list, with enough room to fix any potential issues that we might run into.

What do you think?

lindgrenj6:
I think there is a problem with the golangci-lint runner - it's failed a bunch of times. Merging now.

PR #404, opened 2022-07-12
Add source_type/application_type skip list env vars to clowdapp

lindgrenj6:
Must have missed adding these when doing the cutover. This is needed to *not* seed ibm/oracle in prod.


syncrou (approved):
LGTM 👍

PR #405, opened 2022-07-12
[RHCLOUD-20347] Stringer Interface Implementation for Configuration

Ellen-Yi-Dong:
What: within sources-api-go/config.go, added a method String_Integrate() that displays the current settings of the pod and the settings can be logged from the "main" function after the logger has been initialized. 
Why: Create easier logging of the current settings of the pod 
JIRA ticket: [RHCLOUD-20347](https://issues.redhat.com/browse/RHCLOUD-20347)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -53,6 +53,37 @@ type SourcesApiConfig struct {
 	TenantTranslatorUrl       string
 }
 
+//String() returns a string that shows the settings in which the pod is running in
+func (s SourcesApiConfig) String() string {
+	var b strings.Builder
+	fmt.Fprintf(&b, "%s=%s ", "AppName", parsedConfig.AppName)
+	fmt.Fprintf(&b, "%s=%v ", "Hostname", parsedConfig.Hostname)
+	fmt.Fprintf(&b, "%s=%v ", "KafkaBrokers", parsedConfig.KafkaBrokers)
+	fmt.Fprintf(&b, "%s=%v ", "KafkaTopics", parsedConfig.KafkaTopics)
+	fmt.Fprintf(&b, "%s=%v ", "KafkaGroupID", parsedConfig.KafkaGroupID)
+	fmt.Fprintf(&b, "%s=%v ", "MetricsPort", parsedConfig.MetricsPort)
+	fmt.Fprintf(&b, "%s=%v ", "LogLevel", parsedConfig.LogLevel)
+	fmt.Fprintf(&b, "%s=%v ", "LogLevelForMiddlewareLogs", parsedConfig.LogLevelForMiddlewareLogs)
+	fmt.Fprintf(&b, "%s=%v ", "LogGroup", parsedConfig.LogGroup)
+	fmt.Fprintf(&b, "%s=%v ", "LogHandler", parsedConfig.LogHandler)
+	fmt.Fprintf(&b, "%s=%v ", "LogLevelForSqlLogs", parsedConfig.LogLevelForSqlLogs)
+	fmt.Fprintf(&b, "%s=%v ", "MarketplaceHost", parsedConfig.MarketplaceHost)
+	fmt.Fprintf(&b, "%s=%v ", "AwsRegion", parsedConfig.AwsRegion)
+	fmt.Fprintf(&b, "%s=%v ", "DatabaseHost", parsedConfig.DatabaseHost)
+	fmt.Fprintf(&b, "%s=%v ", "DatabasePort", parsedConfig.DatabasePort)
+	fmt.Fprintf(&b, "%s=%v ", "DatabaseName", parsedConfig.DatabaseName)
+	fmt.Fprintf(&b, "%s=%v ", "FeatureFlagsEnvironment", parsedConfig.FeatureFlagsEnvironment)
+	fmt.Fprintf(&b, "%s=%v ", "FeatureFlagsUrl", parsedConfig.FeatureFlagsUrl)
+	fmt.Fprintf(&b, "%s=%v ", "FeatureFlagsService", parsedConfig.FeatureFlagsService)
+	fmt.Fprintf(&b, "%s=%v ", "CacheHost", parsedConfig.CacheHost)
+	fmt.Fprintf(&b, "%s=%v ", "CachePort", parsedConfig.CachePort)
+	fmt.Fprintf(&b, "%s=%v ", "SlowSQLThreshold", parsedConfig.SlowSQLThreshold)
+	fmt.Fprintf(&b, "%s=%v ", "BypassRbac", parsedConfig.BypassRbac)
+	fmt.Fprintf(&b, "%s=%v ", "SecretStore", parsedConfig.SecretStore)
+	fmt.Fprintf(&b, "%s=%v ", "TenantTranslatorUrl", parsedConfig.TenantTranslatorUrl)
+	return b.String()
+}
+
 // Get - returns the config parsed from runtime vars
 func Get() *SourcesApiConfig {
 	// check if we have already parsed the config - otherwise populate it.
--- a/main.go
+++ b/main.go
@@ -45,7 +45,7 @@ func main() {
 		go runServer(shutdown)
 		go runMetricExporter()
 	}
-
+	logging.Log.Info(conf)
 	// wait for a signal from the OS, gracefully terminating the echo servers
 	// if/when that comes in
 	s := <-interrupts


config/config.go
```diff
@@ -53,6 +53,46 @@ type SourcesApiConfig struct {
 	TenantTranslatorUrl       string
 }
 
+//String_Integrate() returns a string that shows the settings in which the pod is running in
+func (s SourcesApiConfig) String_Integrate() string {
```
MikelAlejoBR: In order to satisfy [the stringer interface](https://go.dev/tour/methods/17) you just have to call the function `String()`

config/config.go
```diff
@@ -53,6 +53,46 @@ type SourcesApiConfig struct {
...
+	fmt.Fprintf(&b, "%s: %v\n", "SlowSQLThreshold", parsedConfig.SlowSQLThreshold)
+	fmt.Fprintf(&b, "%s: %v\n", "Psks", parsedConfig.Psks)
+	fmt.Fprintf(&b, "%s: %v\n", "BypassRbac", parsedConfig.BypassRbac)
+	fmt.Fprintf(&b, "%s: %v\n", "StatusListener", parsedConfig.StatusListener)
+	fmt.Fprintf(&b, "%s: %v\n", "BackgroundWorker", parsedConfig.BackgroundWorker)
+	fmt.Fprintf(&b, "%s: %v\n", "Migrations", parsedConfig.MigrationsSetup)
+	fmt.Fprintf(&b, "%s: %v\n", "MigrationsReset", parsedConfig.MigrationsReset)
+	fmt.Fprintf(&b, "%s: %v\n", "SecretStore", parsedConfig.SecretStore)
+	fmt.Fprintf(&b, "%s: %v\n", "TenantTranslatorUrl", parsedConfig.TenantTranslatorUrl)
+	fmt.Println(b.String())
```
MikelAlejoBR: I think you forgo this `Println` here

main.go
```diff
@@ -45,7 +45,7 @@ func main() {
 		go runServer(shutdown)
 		go runMetricExporter()
 	}
-
+	logging.Log.Infof(conf.String_Integrate())
```
MikelAlejoBR: If you call the above function `String()`, this call can end up being like this:

```diff
- logging.Log.Infof(conf.String_Integrate())
+logging.Log.Info(conf)
```

And the logger will call that `String()` function to print all the stuff.

MikelAlejoBR (commented):
Nice first pass! I have just a few comments :)

config/config.go
```diff
@@ -56,39 +56,40 @@ type SourcesApiConfig struct {
...
+	fmt.Fprintf(&b, "%s: %v\n", "KafkaGroupID", parsedConfig.KafkaGroupID)
+	fmt.Fprintf(&b, "%s: %v\n", "MetricsPort", parsedConfig.MetricsPort)
+	fmt.Fprintf(&b, "%s: %v\n", "LogLevel", parsedConfig.LogLevel)
+	fmt.Fprintf(&b, "%s: %v\n", "LogLevelForMiddlewareLogs", parsedConfig.LogLevelForMiddlewareLogs)
+	fmt.Fprintf(&b, "%s: %v\n", "LogGroup", parsedConfig.LogGroup)
+	fmt.Fprintf(&b, "%s: %v\n", "LogHandler", parsedConfig.LogHandler)
+	fmt.Fprintf(&b, "%s: %v\n", "LogLevelForSqlLogs", parsedConfig.LogLevelForSqlLogs)
+	fmt.Fprintf(&b, "%s: %v\n", "MarketplaceHost", parsedConfig.MarketplaceHost)
+	fmt.Fprintf(&b, "%s: %v\n", "AwsRegion", parsedConfig.AwsRegion)
+	//fmt.Fprintf(&b, "%s: %v\n", "AwsAccessKeyID", parsedConfig.AwsAccessKeyID)
```
lpichler: it is not needed to add code which is commented out - only in rare situations if we will need to add such code later

lpichler:
2 and 3 commits don't need to be in separated commits you can put to first one with operation squash in git.
Second commit - adding comment which pertains code in first commit - I believe it can be together in one commit 
Third commit - things like removing code which help during development - also it is not needed to be here (you can also put it to first commit)
Last commit - is ok - could be here - but on the other hand aim of this PR is basically to add receiver function - so it would also look great if everything would be in one commit.

Otherwise it looks good!

lpichler (approved):
LGTM, thanks!

config/config.go
```diff
@@ -53,6 +53,42 @@ type SourcesApiConfig struct {
 	TenantTranslatorUrl       string
 }
 
+//String_Integrate() returns a string that shows the settings in which the pod is running in
```
MikelAlejoBR: I think you forgot to update the comment:

```diff
- //String_Integrate() returns a string that shows the settings in which the pod is running in
+// String() returns a string that shows the settings in which the pod is running in
```

Other than that, the PR looks good to me :+1:  Good job :+1:

lpichler: Screenshot 2022-07-14 at 19 26 47

config/config.go
```diff
@@ -53,6 +53,42 @@ type SourcesApiConfig struct {
...
+	fmt.Fprintf(&b, "%s: %v\n", "DatabaseHost", parsedConfig.DatabaseHost)
+	fmt.Fprintf(&b, "%s: %v\n", "DatabasePort", parsedConfig.DatabasePort)
+	fmt.Fprintf(&b, "%s: %v\n", "DatabaseName", parsedConfig.DatabaseName)
+	fmt.Fprintf(&b, "%s: %v\n", "FeatureFlagsEnvironment", parsedConfig.FeatureFlagsEnvironment)
+	fmt.Fprintf(&b, "%s: %v\n", "FeatureFlagsUrl", parsedConfig.FeatureFlagsUrl)
+	fmt.Fprintf(&b, "%s: %v\n", "FeatureFlagsService", parsedConfig.FeatureFlagsService)
+	fmt.Fprintf(&b, "%s: %v\n", "CacheHost", parsedConfig.CacheHost)
+	fmt.Fprintf(&b, "%s: %v\n", "CachePort", parsedConfig.CachePort)
+	fmt.Fprintf(&b, "%s: %v\n", "SlowSQLThreshold", parsedConfig.SlowSQLThreshold)
+	fmt.Fprintf(&b, "%s: %v\n", "Psks", parsedConfig.Psks)
```
lindgrenj6: We probably don't want to log the psks - they're semi-private and probably shouldn't be shipped to cloudwatch.

config/config.go
```diff
@@ -53,6 +53,42 @@ type SourcesApiConfig struct {
...
+	fmt.Fprintf(&b, "%s: %v\n", "BypassRbac", parsedConfig.BypassRbac)
+	fmt.Fprintf(&b, "%s: %v\n", "StatusListener", parsedConfig.StatusListener)
+	fmt.Fprintf(&b, "%s: %v\n", "BackgroundWorker", parsedConfig.BackgroundWorker)
+	fmt.Fprintf(&b, "%s: %v\n", "Migrations", parsedConfig.MigrationsSetup)
+	fmt.Fprintf(&b, "%s: %v\n", "MigrationsReset", parsedConfig.MigrationsReset)
```
lindgrenj6: I also don't think we need these flags, the first 2 would only be when specifying backgroundworker/listener - and the other two are kind of internal to the migrate/reset flags on the app.

lindgrenj6 (commented):
This looks good, just a few tiny things that should be an easy fix. 

When we print it out at the beginning it appears to look like this:
```
{"@timestamp":"2022-07-14T13:48:49.522Z","@version":1,"app":"source-api-go","caller":"main.main","filename":"/home/jlindgren/git/crc/sources/go-rewrite/main.go:48","hostname":"mewtwo","labels":{"app":"source-api-go"},"level":"info","log_type":"default","message":"AppName: source-api-go\nHostname: mewtwo\nKafkaBrokers: [mewtwo:9092]\nKafkaTopics: map[]\nKafkaGroupID: sources-api-go\nMetricsPort: 9394\nLogLevel: debug\nLogLevelForMiddlewareLogs: DEBUG\nLogGroup: \nLogHandler: \nLogLevelForSqlLogs: DEBUG\nMarketplaceHost: \nAwsRegion: us-east-1\nDatabaseHost: mewtwo\nDatabasePort: 5432\nDatabaseName: sources_api_development\nFeatureFlagsEnvironment: development\nFeatureFlagsUrl: \nFeatureFlagsService: \nCacheHost: mewtwo\nCachePort: 6379\nSlowSQLThreshold: 2\nPsks: [1234 thisMustBeEphemeralOrMinikube]\nBypassRbac: false\nStatusListener: false\nBackgroundWorker: false\nMigrations: false\nMigrationsReset: false\nSecretStore: database\nTenantTranslatorUrl: \n","tags":["source-api-go"]}
```

can we change it to be the pattern `%s=%s ` (with the space added at the end), that way its a k/v style look and it'll look better in the log line since the new lines aren't actually printed in our logger.

lindgrenj6 (approved):

PR #407, opened 2022-07-14
[RHCLOUD-20393] Added check if app exists into ApplicationPause() and ApplicationUnpause()

petracihalova:
in ApplicationPause() and ApplicationUnpause() we need a check if app exists before we pause or unpause the application

**JIRA:** [RHCLOUD-20393](https://issues.redhat.com/browse/RHCLOUD-20393)

### DIFF
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -327,6 +327,17 @@ func ApplicationPause(c echo.Context) error {
 		return err
 	}
 
+	// Check if the application exists
+	appExists, err := applicationDao.Exists(applicationId)
+	if err != nil {
+		return err
+	}
+
+	if !appExists {
+		return util.NewErrNotFound("application")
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (approved):

lpichler (approved):

application_handlers.go
```diff
@@ -327,6 +327,13 @@ func ApplicationPause(c echo.Context) error {
 		return err
 	}
 
+	// Check if the application exists
+	_, err = applicationDao.GetById(&applicationId)
```
lindgrenj6: I think we can use `applicationDao.Exists(id)` instead.

petracihalova: good point 👍🏻 I forgot we have it

PR #409, opened 2022-07-14
[RHCLOUD-19859] fix: pr check fails due to Bonfire not finding "sources-api-go"

MikelAlejoBR:
Due to the latest change in AppInterface from "sources-api-go" to
"sources-api", Bonfire is unable to deploy Sources to ephemeral during
the PR check.

## Links

[[RHCLOUD-19859]](https://issues.redhat.com/browse/RHCLOUD-19859)

PR #411, opened 2022-07-14
default to 1 replica for availability status listener

lindgrenj6:
Leftovers from when we initially developed this.


MikelAlejoBR (approved):

lindgrenj6:
Tests are failing due to #409 I think.

MikelAlejoBR:
Yes they are. Shouldn't be too bad merging this one.

PR #412, opened 2022-07-14
Move user creation to middleware

lpichler:
- This PR moves user creation from bulk create (which was added [here](https://github.com/RedHatInsights/sources-api-go/pull/348) to middleware).
- Users are created from xhr header on bulk create request. (just now - it will on more api requests)  
- `users#id` is set in echo context at key `h.USERID`

### Links
- parent issue https://github.com/RedHatInsights/sources-api-go/issues/356


bulk_handlers.go
```diff
@@ -33,7 +33,14 @@ func BulkCreate(c echo.Context) error {
 		return fmt.Errorf("failed to pull identity from request")
 	}
 
-	user := &m.User{TenantID: tenantID, UserID: id.Identity.User.UserID}
+	userID, ok := c.Get(h.USERID).(int64)
+
+	if !ok {
+		c.Logger().Warnf("failed to pull user id from request")
+		return fmt.Errorf("failed to pull user id from request")
+	}
```
MikelAlejoBR: I think we might receive bulk create requests without the user ID

MikelAlejoBR (approved):
LGTM!

bulk_handlers.go
lpichler: yes that's right when we authorise with PSK we don't have userid, thanks!

dao/user_dao_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"testing"
+
+	"github.com/RedHatInsights/sources-api-go/internal/testutils"
+	"github.com/RedHatInsights/sources-api-go/internal/testutils/fixtures"
+	"github.com/RedHatInsights/sources-api-go/model"
+)
+
+// TestGetOrCreateTenantIDEbsNumberCreate tests that when the EBS account number is not found, a new tenant is created.
+func TestFindOrCreateUser(t *testing.T) {
```
petracihalova: here is not updated comment above the TestFindOrCreateUser()

lpichler: Comment removed.

bulk_handlers.go
lpichler: I did the change to last commit.

petracihalova (approved):

middleware/user.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/RedHatInsights/sources-api-go/dao"
+	h "github.com/RedHatInsights/sources-api-go/middleware/headers"
+	"github.com/labstack/echo/v4"
+	"github.com/redhatinsights/platform-go-middlewares/identity"
+)
+
+func UserCatcher(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		xRhIdentity, ok := c.Get(h.PARSED_IDENTITY).(*identity.XRHID)
```
lindgrenj6, line 14: So this gets interesting. With PSK support we actually won't always have the xrhid. So realistically this function needs to handle either a new header (`x-rh-sources-user-id`?) OR the xrhid. If the xrhid nor that header are present then we can just continue on (it's not required per se since it's only for the special user sources). 

We would throw an error if the user tried to create a user source without either/or present. Other than that it shouldn't return an error.

lindgrenj6 (commented):
Not sure if we want to handle this now or later - right now it would ONLY be require on /bulk_create (i use PSK for bulk-create locally, am not sure about the various IQE tests that run against us so it could break other tests).

Maybe we should just check if x-rh-id is present, and if not just continue along.

middleware/user.go
lpichler: I can add `x-rh-sources-user-id` in follow up PR.

we have parsed xrhid header always - for PSK we generate it to the context. So now it will not fail with PSK.

Do we need to check existence of xrhid  in header despite it ?

lindgrenj6: Well the only problem with that generated header is that it will be the default value:
```
(dlv) p i.Identity.User
github.com/redhatinsights/platform-go-middlewares/identity.User {Username: "", Email: "", FirstName: "", LastName: "", Active: false, OrgAdmin: false, Internal: false, Locale: "", UserID: ""}
(dlv) p i.Identity.User.UserID
""
```

If you don't want to check for existence of the header you will need to check and see if the UserID is `""` before setting it on the context since `""` isn't really a valid UserID.

lpichler: I have it in middleware/user.go:25  or do you reference other place ?

lindgrenj6: Oh apparently I missed that. My bad!

lpichler: That's ok at least I looked more how headers are going through our middlewares, thanks!

PR #391, opened 2022-07-01
[WIP] Populate userid in bulk create 

lpichler:
### Prerequisites 

To make this work we need to set `resource_ownership = user` in app-studio app type.

```diff
diff --git a/dao/seeds/application_types.yml b/dao/seeds/application_types.yml
index b27b4e7..51d1c20 100644
--- a/dao/seeds/application_types.yml
+++ b/dao/seeds/application_types.yml
@@ -19,6 +19,7 @@
       - gitlab-personal-access-token
     quay:
       - quay-encrypted-password
+  resource_ownership: user

 "/insights/platform/catalog":
   display_name: Automation Services Catalog
```

### Description
Changes in this PR adds to creation with bulk create possibility to populate column `userid` in created records of tables:
`sources, applications, application_authentications and authentications`.

`userid` is taken from table `users` where user record here was added from XRHID header - user creation was added [here](https://github.com/RedHatInsights/sources-api-go/pull/348) 

`userid` is populated in records which are associated with `application_type` with column `resource_ownership=user`. App studio application type will have `resource_ownership=user` in future PR.

Examples of bulk create request. (let's count that diff above was applied):

```
POST /api/sources/v3.1/bulk_create

{
  "sources": [
    {
      "name": "B17",
      "source_type_name": "bitbucket"
    }
  ],
  "applications": [
    {
      "source_name": "B17",
      "application_type_name": "app-studio"
    }
  ],
  "authentications": [
    {
      "resource_type": "application",
      "resource_name": "app-studio",
      "username": "myUser",
      "password": "myPass"
    }
  ]
}
```

```
POST /api/sources/v3.1/bulk_create

{
  "sources": [
    {
      "name": "B15",
      "source_type_name": "bitbucket"
    }
  ],
  "applications": [
    {
      "source_name": "B15",
      "application_type_name": "app-studio"
    }
  ],
  "authentications": [
    {
      "resource_type": "source",
      "username": "myUser",
      "password": "myPass"
    }
  ]
}
```

### Depends on
- [x] https://github.com/RedHatInsights/sources-api-go/pull/348

PR #419, opened 2022-07-19
Populate userid in bulk create

lpichler:
This is last PR from series about creation user based resources.

### Prerequisites 

To make this work we need to set `resource_ownership = user` in app-studio app type.

```diff
diff --git a/dao/seeds/application_types.yml b/dao/seeds/application_types.yml
index b27b4e7..51d1c20 100644
--- a/dao/seeds/application_types.yml
+++ b/dao/seeds/application_types.yml
@@ -19,6 +19,7 @@
       - gitlab-personal-access-token
     quay:
       - quay-encrypted-password
+  resource_ownership: user

 "/insights/platform/catalog":
   display_name: Automation Services Catalog
```

## Description

Changes in this PR adds to creation with `bulk_create` possibility to populate column userid in created records of tables:
sources, applications, application_authentications and authentications.

`user_id` is taken from table `users` - Users are  created from xhr header- [added here](https://github.com/RedHatInsights/sources-api-go/pull/412) 

`user_id` is populated in records which are associated with `application_type` with column `resource_ownership=user`. App studio application type will set `resource_ownership=user` in future PR.

## How it works

In bulk create we call `loadUserResourceSettingFromBulkCreateApplication` (It is not added in this PR). We collect application type names which have `resource_ownership=user` to `UserResource#ApplicationTypesNames` and also we collect  source names(to `UserResource#SourceNames`) which are tied with this used based application type `resource_ownership=user` .

With those  `UserResource#ApplicationTypesNames` and `UserResource#SourceNames` we can easily ask whether we should populate userid on resources in parsing methods(parseSources,...).

Examples of bulk create request. (let's count that diff above was applied):

```
POST /api/sources/v3.1/bulk_create

{
  "sources": [
    {
      "name": "B17",
      "source_type_name": "bitbucket"
    }
  ],
  "applications": [
    {
      "source_name": "B17",
      "application_type_name": "app-studio"
    }
  ],
  "authentications": [
    {
      "resource_name": "app-studio",
      "username": "myUser",
      "password": "myPass"
    }
  ]
}
```

```
POST /api/sources/v3.1/bulk_create

{
  "sources": [
    {
      "name": "B15",
      "source_type_name": "bitbucket"
    }
  ],
  "applications": [
    {
      "source_name": "B15",
      "application_type_name": "app-studio"
    }
  ],
  "authentications": [
    {
      "resource_type": "source",
      "username": "myUser",
      "password": "myPass"
    }
  ]
}
```

### DIFF
--- a/service/bulk_create.go
+++ b/service/bulk_create.go
@@ -88,14 +88,26 @@ func BulkAssembly(req m.BulkCreateRequest, tenant *m.Tenant, user *m.User) (*m.B
 			}
 
 			if strings.ToLower(output.Authentications[i].ResourceType) == "application" {
-				output.ApplicationAuthentications = append(output.ApplicationAuthentications, m.ApplicationAuthentication{
+				applicationAuthentication := m.ApplicationAuthentication{
 					// TODO: After vault migration.
 					// VaultPath:         output.Authentications[i].Path(),
 					ApplicationID:    output.Authentications[i].ResourceID,
 					AuthenticationID: output.Authentications[i].DbID,
 					TenantID:         tenant.Id,
 					Tenant:           *tenant,
-				})
+				}
+
+				applicationFromAuthentication := &m.Application{ID: output.Authentications[i].ResourceID}
+				tx.Debug().Preload("ApplicationType").Find(&applicationFromAuthentication)
+
+				applicationType := &m.ApplicationType{Id: applicationFromAuthentication.ApplicationTypeID}
+				if tx.Debug().First(&applicationType).Error == nil {
+					if userResource.OwnershipPresentForApplication(applicationType.Name) {
+						applicationAuthentication.UserID = &userResource.User.Id
+					}
+				}
+
+				output.ApplicationAuthentications = append(output.ApplicationAuthentications, applicationAuthentication)
 			}
 		}
 
--- a/service/bulk_create.go
+++ b/service/bulk_create.go
@@ -48,7 +48,7 @@ func BulkAssembly(req m.BulkCreateRequest, tenant *m.Tenant, user *m.User) (*m.B
 		userResource.User = user
 
 		// parse the sources, then save them in the transaction.
-		output.Sources, err = parseSources(req.Sources, tenant)
+		output.Sources, err = parseSources(req.Sources, tenant, userResource)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


lpichler:
please re-review - it is similar to https://github.com/RedHatInsights/sources-api-go/pull/391 but less commits ;) Thanks!

service/bulk_create.go
```diff
@@ -88,14 +88,26 @@ func BulkAssembly(req m.BulkCreateRequest, tenant *m.Tenant, user *m.User) (*m.B
...
+				}
+
+				applicationFromAuthentication := &m.Application{ID: output.Authentications[i].ResourceID}
+				tx.Debug().Preload("ApplicationType").Find(&applicationFromAuthentication)
+
+				applicationType := &m.ApplicationType{Id: applicationFromAuthentication.ApplicationTypeID}
+				if tx.Debug().First(&applicationType).Error == nil {
+					if userResource.OwnershipPresentForApplication(applicationType.Name) {
+						applicationAuthentication.UserID = &userResource.User.Id
+					}
```
MikelAlejoBR: I would suggest not using structs along with Gorm's `Find` functions to avoid issues like [RHCLOUD-20441](https://issues.redhat.com/browse/RHCLOUD-20441). The TL;DR is:

If the `out.put.Authentications[i].ResourceID` has a default value for any reason, Gorm will assume that it is safe not to include the `where application_id = 0` condition to the query, which will result in a `SELECT * FROM application`, and I guess that the `Find` method will pick up one of the results randomly.

So I'd suggest:

```diff
- applicationFromAuthentication := &m.Application{ID: output.Authentications[i].ResourceID}
+ var applicationFromAuthentication &m.Application
- tx.Debug().Preload("ApplicationType").Find(&applicationFromAuthentication)
+ tx.Debug().Preload("ApplicationType").Where("application_id = ?", output.Authentications[i].ResourceID).Find(&applicationFromAuthentication)
```

MikelAlejoBR (commented):
Just one comment, other than that, looks good to me! :+1:

petracihalova (approved):

service/bulk_create.go
lpichler: That's is right and it is also for usages of gorm Find.

I change code little bit and I avoided first query because we have already fetched  in `output.Applications`


thanks!

lindgrenj6 (approved):

PR #420, opened 2022-07-19
Access UserID in source handler and DAO

lpichler:
This PR access user id (from `users#id`) from echo context (populated in UserCatcher middleware) to DAO layer - so we can filter according it in future PRs.

Middleware UserCatcher is added to primary and subcollection sources routes. (except to POST)

### DIFF
--- a/routes.go
+++ b/routes.go
@@ -13,7 +13,7 @@ var listMiddleware = []echo.MiddlewareFunc{
 	middleware.SortAndFilter, middleware.Pagination,
 }
 
-var tenancyWithListMiddleware = append([]echo.MiddlewareFunc{middleware.Tenancy}, listMiddleware...)
+var tenancyWithListMiddleware = append([]echo.MiddlewareFunc{middleware.Tenancy, middleware.UserCatcher}, listMiddleware...)
 var permissionMiddleware = []echo.MiddlewareFunc{middleware.Tenancy, middleware.PermissionCheck, middleware.RaiseEvent}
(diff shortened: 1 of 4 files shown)


routes.go
```diff
@@ -33,11 +33,11 @@ func setupRoutes(e *echo.Echo) {
 		r.POST("/bulk_create", BulkCreate, append(permissionMiddleware, middleware.UserCatcher)...)
 
 		// Sources
-		r.GET("/sources", SourceList, tenancyWithListMiddleware...)
-		r.GET("/sources/:id", SourceGet, middleware.Tenancy)
+		r.GET("/sources", SourceList, append(tenancyWithListMiddleware, middleware.UserCatcher)...)
```
lindgrenj6: So rather than appending it everywhere - what would you think about sticking it up top next to `middleware.Tenancy`? It is doing something really similar and we'll want to do it for basically every request that we have tenancy on right now as well.

lindgrenj6 (commented):
Just a tiny comment in routes, otherwise :+1:

MikelAlejoBR (approved):
LGTM!

routes.go
lpichler: yes it is much better - done  I change second commit - https://github.com/RedHatInsights/sources-api-go/pull/420/commits/dea0f1c73ac3b5ba36383c84eb160b38fd8b1a78 

thanks!

lindgrenj6 (approved):
Nice :+1:

PR #421, opened 2022-07-19
[RHCLOUD-20440] Return 400 responses when the provided IDs are invalid

MikelAlejoBR:
Essentially this ensures that all the endpoints which are expecting an ID, return a bad request response when that ID is empty or zero. I've also moved all the ID processing to the top of those handlers, since it doesn't make sense to get the DAOs, process the filters and then stop the execution because the ID is invalid.

I initially tried doing it as a middleware, but the code ended up being like the sample below, which in my opinion is not worth having on every handler.

```go
idRaw := c.Get("id")
id, ok := idRaw.(int64)
if !ok {
    return errors.New("could not extract ID from the request")
}
```

### DIFF
--- a/middleware/id_validation.go
+++ b/middleware/id_validation.go
@@ -1,0 +1,72 @@
+package middleware
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/RedHatInsights/sources-api-go/util"
+	"github.com/labstack/echo/v4"
+)
+
+// IdValidation takes all the parameters which end with "id" and checks for their validity. Returns a bad request if
+// they're not valid.
+func IdValidation(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		for _, param := range c.ParamNames() {
+			if strings.HasSuffix(param, "id") {
+				err := validateId(c, param)
+				if err != nil {
+					return util.NewErrBadRequest(err)
+				}
+			}
+		}
+
+		return next(c)
+	}
+}
+
+// UuidValidation checks if the UUID parameter is valid. Returns a bad request if it isn't.
+func UuidValidation(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		err := validateUuid(c)
+		if err != nil {
+			return util.NewErrBadRequest(err)
+		}
+
+		return next(c)
+	}
+}
+
+// validateId checks if the ID stored in specified context parameter isn't empty, is a valid integer and that it is
+// greater than zero.
+func validateId(c echo.Context, idParamName string) error {
+	idRaw := c.Param(idParamName)
+
+	if idRaw == "" {
+		return errors.New("the provided ID cannot be empty or missing")
+	}
+
+	id, err := strconv.ParseInt(idRaw, 10, 64)
+	if err != nil {
+		return fmt.Errorf("could not parse the provided ID: %s", err)
+	}
+
+	if !(id > 0) {
+		return errors.New("the provided ID must be greater than zero")
+	}
+
+	return nil
+}
+
+// validateUuid checks if the UUID is not empty.
+func validateUuid(c echo.Context) error {
+	uid := c.Param("uid")
+
+	if uid == "" {
+		return errors.New("the UUID cannot be empty or missing")
+	}
+
+	return nil
+}
--- a/middleware/id_validation.go
+++ b/middleware/id_validation.go
@@ -3,12 +3,15 @@ package middleware
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
 )
+
+var idValidationRegex = regexp.MustCompile(`^\d+$`)
 
 // IdValidation takes all the parameters which end with "id" and checks for their validity. Returns a bad request if
 // they're not valid.
@@ -44,12 +47,17 @@ func UuidValidation(next echo.HandlerFunc) echo.HandlerFunc {
 func validateId(c echo.Context, idParamName string) error {
 	idRaw := c.Param(idParamName)
 
+	if !idValidationRegex.MatchString(idRaw) {
+		return errors.New("the provided ID must be a greater than zero, positive number")
+	}
+
+	// This check is required for the edge case of an out of range error.
 	id, err := strconv.ParseInt(idRaw, 10, 64)
 	if err != nil {
 		return fmt.Errorf("could not parse the provided ID: %s", err)
 	}
 
-	if !(id > 0) {
+	if id == 0 {
 		return errors.New("the provided ID must be greater than zero")
 	}
 
--- a/middleware/id_validation.go
+++ b/middleware/id_validation.go
@@ -44,10 +44,6 @@ func UuidValidation(next echo.HandlerFunc) echo.HandlerFunc {
 func validateId(c echo.Context, idParamName string) error {
 	idRaw := c.Param(idParamName)
 
-	if idRaw == "" {
-		return errors.New("the provided ID cannot be empty or missing")
-	}
-
 	id, err := strconv.ParseInt(idRaw, 10, 64)
 	if err != nil {
 		return fmt.Errorf("could not parse the provided ID: %s", err)
--- a/middleware/id_validation_test.go
+++ b/middleware/id_validation_test.go
@@ -50,28 +50,54 @@ func TestExtractValidateId(t *testing.T) {
 	}
 }
 
-// TestExtractValidateIdNonParseable tests that a bad request response is returned when a non-parseable ID is set.
-func TestExtractValidateIdNonParseable(t *testing.T) {
+// TestExtractValidateIdInvalidFormats tests that a bad request response is returned when an ID which is in an invalid
+// format is given.
+func TestExtractValidateIdInvalidFormats(t *testing.T) {
+	testValues := []string{"-14", "xxx", "+1232"}
+	paramName := "id"
+
+	for _, tv := range testValues {
+		c, rec := request.CreateTestContext(http.MethodGet, "/", nil, nil)
+
+		c.SetParamNames(paramName)
+		c.SetParamValues(tv)
+
+		err := idValidationFunc(c)
+		if err != nil {
+			t.Errorf(`unexpected error received when a validating an invalidly formatted ID: %s`, err)
+		}
+
+		templates.BadRequestTest(t, rec)
+
+		want := "the provided ID must be a greater than zero, positive number"
+		got := rec.Body.String()
+
+		if !strings.Contains(got, want) {
+			t.Errorf(`unexpected error received when testing for an invalidly formatted ID. Want "%s", got "%s"`, want, got)
+		}
+	}
+}
+
+// TestExtractValidateIdInvalidFormats tests that a bad request response is returned when an ID which is in an invalid
+// format is given.
+func TestExtractValidateIdOutOfRange(t *testing.T) {
 	c, rec := request.CreateTestContext(http.MethodGet, "/", nil, nil)
 
-	paramName := "id"
-	paramValue := "abcde"
-
-	c.SetParamNames(paramName)
-	c.SetParamValues(paramValue)
+	c.SetParamNames("id")
+	c.SetParamValues("10000000000000000000000000000000000000000000000000")
 
 	err := idValidationFunc(c)
 	if err != nil {
-		t.Errorf(`unexpected error received when a validating a non parseable ID: %s`, err)
+		t.Errorf(`unexpected error received when a validating an out of range ID: %s`, err)
 	}
 
 	templates.BadRequestTest(t, rec)
 
-	want := "could not parse the provided ID"
+	want := "value out of range"
 	got := rec.Body.String()
 
 	if !strings.Contains(got, want) {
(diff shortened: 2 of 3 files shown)


service/id_validation.go
```diff
@@ -0,0 +1,49 @@
...
+func ExtractValidateId(c echo.Context, idParamName string) (int64, error) {
+	idRaw := c.Param(idParamName)
+
+	if idRaw == "" {
+		return 0, errors.New("the provided ID cannot be empty or missing")
+	}
+
+	id, err := strconv.ParseInt(idRaw, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf(`could not parse the provided ID: %s`, err)
```
lpichler: can I ask what is difference between \` and " ?

main_test.go
```diff
@@ -104,3 +106,71 @@ func SortByStringValueOnKey(field string, data []interface{}) {
...
 }
+
+// TestInvalidIdsReturn400 tests that when the handlers are given invalid IDs, these return a bad request response back
+// to the client.
+func TestInvalidIdsReturn400(t *testing.T) {
+
+	testData := []struct {
+		Method      string
+		IdParamName string
+		Handler     func(c echo.Context) error
```
petracihalova: I really love this test ... 🤔 not sure if we want to have it here, maybe we can move it somewhere (maybe here middleware/error_handling_test.go)

petracihalova (approved):
nice job !!!

main_test.go
```diff
@@ -104,3 +106,71 @@ func SortByStringValueOnKey(field string, data []interface{}) {
...
+		{http.MethodGet, "application_type_id", ApplicationTypeListMetaData},
+		{http.MethodGet, "id", SourceTypeGet},
+		{http.MethodGet, "source_type_id", SourceTypeListSource},
+		{http.MethodGet, "id", RhcConnectionGetById},
+		{http.MethodPatch, "id", RhcConnectionEdit},
+		{http.MethodDelete, "id", RhcConnectionDelete},
+		{http.MethodGet, "id", RhcConnectionSourcesList},
+	}
+
+	testValues := []string{"", "0"}
```
lpichler: can we add more ? also -1 and XXX

lpichler:
Does this should cover also filtering ?  like `/sources?filter[id][eq]=0` - I guess no - it is other topic.

lpichler (approved):
Very nice! 👏

main_test.go
petracihalova: if you add the value of "xxx" into testValues and you can probably replace this code
```
if rec.Code != http.StatusBadRequest {
	t.Errorf(`unexpected status code received. Want "%d", got "%d". Body: %s`, http.StatusBadRequest, rec.Code, rec.Body)
}
```

with this 
`templates.BadRequestTest(t, rec)`

I believe we can then remove then many bad request tests that only checks situation for id = "xxx"
(maybe this all can be in next PR ... let's not make it too big PR)

service/id_validation.go
MikelAlejoBR: In this particular case it's true that it doesn't make much sense to use backticks, since I'm not wrapping the error message between quotes. This is mainly the reason I use backticks for, or when I want Golang to just not touch the string, and to keep the formatting I specify: https://go.dev/ref/spec#String_literals

I'm going to switch the `` ` `` for `"`.

main_test.go
MikelAlejoBR: I placed it in `main_test.go` because to be honest I didn't know where to put it. However I don't think it belongs to `middleware/`, since I'm testing all the handlers, so that's why I thought it should go on the main package — also the second reason is that I cannot access the handlers from within a submodule.

I don't know. Maybe `id_validation_handlers_test.go`?

main_test.go
MikelAlejoBR: Sure! Done! :+1:

main_test.go
petracihalova: id_validation_handlers_test.go sounds good ⭐️

MikelAlejoBR: Done! Thank you!

MikelAlejoBR:
> Does this should cover also filtering ? like `/sources?filter[id][eq]=0` - I guess no - it is other topic.

Actually this is a great catch — I forgot about it. However I agree with you that maybe we should do it separately. I've created a Jira ticket for it: https://issues.redhat.com/browse/RHCLOUD-20457

service/id_validation.go
lpichler: I see, thanks for explanation!

lindgrenj6 (commented):
(sent this in DM, but duplicating here)

For this I think it would be better if we moved the check to a middleware since it is the same check for every controller. 

here is a POC function i just threw in main. something like this can probably be ran on all handlers since it just loops through all of the path params every time, might catch something in the future if we introduce a new route with a param. 

```go
e.Use(func(next echo.HandlerFunc) echo.HandlerFunc {
	return func(c echo.Context) error {
		// loop through all path params, picking out the ones that have `id` at the end
		for _, param := range c.ParamNames() {
			if strings.HasSuffix(param, "id") {
				err := validateId(c.Param(param)) // validation function
				if err != nil {
					return util.NewErrBadRequest("invalid id passed") 
				}
			}
		}
		return next(c)
	}
})
```

MikelAlejoBR:
All right, so I've refactored this PR and made it as a middleware. It is true that it is way less code than the previous PR. Thanks for the suggestion, @lindgrenj6!

I'm requesting a re-review for this one :sweat_smile: 

If you're missing that big test which called all the handlers: this is not required now, since by testing the middlewares directly — as opposed to having these checks as a service, which required us calling every handler — we are already testing the behaviour for any route that runs the middleware.

lindgrenj6 (approved):
In a follow-up I think we'll have to change how we handle the middleware, right now we have a few specific ones for listing/tenancy/etc but we might need to step back and reorganize them a bit in the future. We're getting a ton of them! (which is great!)

petracihalova:
I have a question ... when the id validation is now solved in the middleware then we not need this in our handlers right?
```
	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
	if err != nil {
	}
```

petracihalova:
and another question ... when we moved this into middleware and we have now test for this in middleware, we dont need to test it in handlers ... I mean the bad request tests that cover situation when id = "xxx"

```
func TestSourceGetBadRequest(t *testing.T) {
	c, rec := request.CreateTestContext(
		http.MethodGet,
		"/api/sources/v3.1/sources/xxx",
		nil,
		map[string]interface{}{
			"tenantID": int64(1),
		},
	)

	c.SetParamNames("id")
	c.SetParamValues("xxx")

	badRequestSourceGet := ErrorHandlingContext(SourceGet)
	err := badRequestSourceGet(c)
	if err != nil {
		t.Error(err)
	}

}
```

middleware/id_validation.go
```diff
@@ -0,0 +1,72 @@
...
+		return next(c)
+	}
+}
+
+// validateId checks if the ID stored in specified context parameter isn't empty, is a valid integer and that it is
+// greater than zero.
+func validateId(c echo.Context, idParamName string) error {
+	idRaw := c.Param(idParamName)
+
+	if idRaw == "" {
```
lpichler: I think this can't happen because when we try it - it ends with other route and in case of sub collection it will be second part of of url `/api/sources/v3.1/sources//application_types`

middleware/id_validation.go
```diff
@@ -0,0 +1,72 @@
...
+	if !(id > 0) {
+		return errors.New("the provided ID must be greater than zero")
+	}
+
+	return nil
+}
+
+// validateUuid checks if the UUID is not empty.
+func validateUuid(c echo.Context) error {
+	uid := c.Param("uid")
```
lpichler, line 69: Do need to validate here just string without special characters ?
`uid==""` is not possible to reach.

lpichler:
Should we consider this as valid url ? 👼 
`/api/sources/v3.1/sources/+2`

MikelAlejoBR:
> I have a question ... when the id validation is now solved in the middleware then we not need this in our handlers right?
> 
> ```
> 	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
> 	if err != nil {
> 		return util.NewErrBadRequest(err)
> 	}
> ```

We still do, because we need to somehow convert the string parameters into integers, since we don't save the parsed ID on the context. Essentially, we're converting the ID twice.

To be honest I don't know what's a better solution to this... On the first pass, when I was writing this as a middleware, I did something like this —plus a very similar one for the UUIDs—:

```go
return func(c echo.Context) error {
	id, err := parseValidateId(c, "id")
	if err != nil {
	}

	c.Set("id", id)
	return next(c)
}
```

For this I changed all the parameter names from the routes to `:id`. But then this forced me to do a type check on every handler:

```go
func SourceGet(c echo.Context) error {
	idRaw := c.Get("id")
	id, ok := idRaw.(int64)
	if !ok {
		return errors.New("could not extract ID from the request")
	}

	// ...
}
```

Which made me feel itchy, because I knew that essentially that typecheck is just "empty code". I know that the middleware had set an `int64` on the context, but still I had to do the typecheck. So that evolved to:

```go
func SourceGet(c echo.Context) error {
	id, _ := c.Get("id").(int64)

	// ...
}

```
However since I don't like ignoring anything it didn't quite feel good for me, so I used the previous "full type check but not necessary" version, which made the code uglier than needed. That plus not being certain about where to put the middleware, if it should go first before any other middleware or not, if I should reorder the middlewares, etc., made me go with the service version of this PR.

---

The service version is the one you reviewed first, which basically was:

```go
func SourceGet(c echo.Context) error {
	id, err := service.ValidateId(c, paramName: "id")
	if err != nil {
		return util.NewBadRequestErr(err)
	}

	// ...
}
```
This made all the middlewares dance go away, but introduced the "bummer" of having to repeat this piece of code on every handler. Although we could keep the parameter names as they were, and since we already were checking for the IDs correctness on the handlers before, I thought "this might be it".

---

Jacob suggested using a middleware, so I went back to option 1, but now we face "the repetition again". As you stated, the following error check is redundant:

```go
	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
	if err != nil {
	}
```

We could simply do `id, _ := strconv.ParseInt(c.Param("id"), 10, 64)`, or go back to the option 1 and store the integer on the context and do something like:

```go
func IdValidationExtraction(next echo.HandlerFunc) echo.HandlerFunc {
	return func(c echo.Context) error {
		for _, param := range c.ParamNames() {
			if strings.HasSuffix(param, "id") {
				id, err := extractValidateId(c, param)
				if err != nil {
				}
				
				c.Set(param, id)
			}
		}

		return next(c)
	}
}
```

And then:

```go
func SourceGet(c echo.Context) error {
	// The middleware has validated the ID, so it is safe to ignore the type check's "ok" return value.
	id, _ := c.Get("source_id").(int64)

	// ...
}
```

I still don't like ignoring anything, and we'd still have to copy&paste that on every handler... To be honest I don't know. What do you all think?


> and another question ... when we moved this into middleware and we have now test for this in middleware, we dont need to test it in handlers ... I mean the bad request tests that cover situation when id = "xxx"
> 
> ```
> func TestSourceGetBadRequest(t *testing.T) {
> 	c, rec := request.CreateTestContext(
> 		http.MethodGet,
> 		"/api/sources/v3.1/sources/xxx",
> 		nil,
> 		map[string]interface{}{
> 			"tenantID": int64(1),
> 		},
> 	)
> 
> 	c.SetParamNames("id")
> 	c.SetParamValues("xxx")
> 
> 	badRequestSourceGet := ErrorHandlingContext(SourceGet)
> 	err := badRequestSourceGet(c)
> 	if err != nil {
> 		t.Error(err)
> 	}
> 
> 	templates.BadRequestTest(t, rec)
> }
> ```

That is true, yes. I should remove all these tests on this PR...

middleware/id_validation.go
MikelAlejoBR: You are right, yes. If we send a request to `/sources/` it will just match the collection's route. Thanks!

MikelAlejoBR:
> Should we consider this as valid url ? angel `/api/sources/v3.1/sources/+2`

Wow, I didn't really think about this. Although it makes sense when using an integer for the IDs. I've another regex check to be 100% sure that the format is correct, and that the number doesn't come with any signs.

petracihalova:
> We still do, because we need to somehow convert the string parameters into integers, since we don't save the parsed ID on the context. Essentially, we're converting the ID twice.

I forgot that we need to convert the string ... but thanks for this explanation/thoughts
and I don't have any new idea how to solve it ... I am +1 to not ignoring err
😂 lot of fun with this task 😂

middleware/id_validation.go
MikelAlejoBR: I was thinking on sticking to the UUID specification, but I don't know why a "we might not receive UUIDs sometimes" crossed my mind, and I didn't implement it, since it might be too strict...

I'm unsure about what to do :thinking:

lpichler: OK - let's leave it as it is now I don't know either.

petracihalova:
> That is true, yes. I should remove all these tests on this PR...

just thinking ... when we let the parsing from string into integer in our handlers, then we probably need to let the bad request tests too because they are testing these handlers and handlers contains parsing from string into integer ... hmm 🤔

petracihalova (approved):

lpichler (approved):

lpichler:
@MikelAlejoBR thanks for addressing comments!

lindgrenj6 (approved):

PR #423, opened 2022-07-21
[RHCLOUD-20398] App auth tests update

petracihalova:
tests update for application authentication handlers PART I.
this PR covers:

- ApplicationAuthenticationList()

main goal is to check the tenancy and add missing tests
tests update for other handlers will be covered in next PRs

**JIRA:** [RHCLOUD-20398](https://issues.redhat.com/browse/RHCLOUD-20398)

PR #425, opened 2022-07-21
[RHCLOUD-20398] App auth tenancy tests2

petracihalova:
tests update for application authentication handlers PART II.
(Part I. #423)
this PR covers:

- ApplicationAuthenticationGet()

main goal is to check the tenancy and add missing tests
tests update for other handlers will be covered in next PRs

**JIRA:** [RHCLOUD-20398](https://issues.redhat.com/browse/RHCLOUD-20398)

### DIFF
--- a/application_authentication_handlers_test.go
+++ b/application_authentication_handlers_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/RedHatInsights/sources-api-go/dao"
 	"net/http"
 	"strconv"
 	"testing"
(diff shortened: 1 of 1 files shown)


petracihalova:
this PR will be rebased after merge #423

petracihalova:
rebased and ready for review

application_authentication_handlers_test.go
```diff
@@ -217,6 +220,71 @@ func TestApplicationAuthenticationGet(t *testing.T) {
 	if out.ApplicationID != "1" {
 		t.Error("ghosts infected the return")
 	}
+
+	// Check the tenancy of returned app auth
+	for _, appAuth := range fixtures.TestApplicationAuthenticationData {
+		if fmt.Sprintf("%d", appAuth.ID) == out.ApplicationID {
```
MikelAlejoBR: I think that here it should be `out.ID`, right?

MikelAlejoBR (approved):
LGTM!

application_authentication_handlers_test.go
petracihalova: that is really awesome catch !!!! sharp eye ... I changed the app auth id that is used for this test to have better test data (previous app auth id = 1 was relation between app id = 1 and auth id = 1 and that is the reason why it worked with` out.ApplicationID`)

lindgrenj6 (approved):

PR #424, opened 2022-07-21
Unify [Source&Authentication]DaoParams to RequestParams struct

lpichler:
Partially addressed [comment](https://github.com/RedHatInsights/sources-api-go/pull/397#discussion_r917926822).

I am doing this change now because there is no reason to follow old pattern with adding [Model]DaoParams.

Feel free to bring better name for the struct Request Params.

### Links

https://github.com/RedHatInsights/sources-api-go/issues/356

PR #426, opened 2022-07-21
Extract tenant and user id to request params constructor

lpichler:
- this moves functions getTenantFromEchoContext and getUserFromEchoContext to util package - so it could be user in other places
- add constructor (I am sure if there is other terminology ) ` NewRequestParamsFromContext` where we fetch tenant_id and user_id from echo context and it could be used in  get[model]DaoWithTenant function. In this PR it is done for  `getSourceDaoWithTenant` and `getAuthenticationDaoWithTenant`.

### Links
Depends on
- [x] https://github.com/RedHatInsights/sources-api-go/pull/424
- issue https://github.com/RedHatInsights/sources-api-go/issues/356


petracihalova (approved):
LGTM

util/context.go
```diff
@@ -0,0 +1,47 @@
+package util
```
lindgrenj6: [tiny]: can we rename this file `echo_context.go`? Context overlaps a bit with the standard library context package.

lindgrenj6 (commented):
Just one tiny thing, otherwise lgtm.

util/context.go
lpichler: sure, done, thanks 👍

PR #427, opened 2022-07-21
[RHCLOUD-20478] chore: improve logging for the availability check process

MikelAlejoBR:
The idea is to make it easier for us to identify issues with the whole availability check process when debugging the logs. By adding relevant IDs to the log messages it will be easier to identify the particular messages we're looking for.

I came up with this after debugging the linked issue, and after trying to follow all the steps of the availability check flow to see if anything failed with it. By including all this extra information it will be easier for us to debug it the next time.

## Links
[[RHCLOUD-20478]](https://issues.redhat.com/browse/RHCLOUD-20478)

### DIFF
--- a/service/availability_check.go
+++ b/service/availability_check.go
@@ -47,7 +47,7 @@ var (
 
 // requests both types of availability checks for a source
 func RequestAvailabilityCheck(source *m.Source, headers []kafka.Header) {
-	l.Log.Infof("Requesting Availability Check for Source [%v]", source.ID)
+	l.Log.Infof("[source_id: %d] Requesting availability check for source", source.ID)
 
 	if len(source.Applications) != 0 {
 		ac.ApplicationAvailabilityCheck(source)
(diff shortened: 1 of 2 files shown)


lpichler (approved):

petracihalova (approved):
👍🏻

lpichler:
Do we also need to add some logs status listener ?

lpichler:
Do we also need to add some logs to status listener ?

lindgrenj6 (approved):

MikelAlejoBR:
Do you mean that we should add even more logs? I only added more info regarding the errors for now, since I think that it might be useful to see what happened if anything fails. However I'm tunnel visioning here so I'm open to suggestions :)

lpichler:
> Do you mean that we should add even more logs? I only added more info regarding the errors for now, since I think that it might be useful to see what happened if anything fails. However I'm tunnel visioning here so I'm open to suggestions :)

ah yes - I think it is ok now I understand more aim of those logs :) thanks!

PR #428, opened 2022-07-21
Add user id header for PSK

lpichler:
This adds possibility to pass user id when PSK is used.
Header has name `x-rh-sources-user-id`.

I also added some test for `UserCatcherMiddleware`.

### Links
https://github.com/RedHatInsights/sources-api-go/issues/356

### DIFF
--- a/middleware/headers_test.go
+++ b/middleware/headers_test.go
@@ -48,7 +48,7 @@ func TestParseAll(t *testing.T) {
 	}
 
 	if c.Get(h.PSK_USER).(string) != "55555" {
-		t.Errorf("%v was set as psk-account instead of %v", c.Get(h.ACCOUNT_NUMBER).(string), "9876")
+		t.Errorf("%v was set as x-rh-sources-user-id instead of %v", c.Get(h.PSK_USER).(string), "55555")
 	}
 
 	if c.Get(h.ORGID).(string) != "abcde" {
@@ -170,6 +170,6 @@ func TestOnlyPskHeaders(t *testing.T) {
 	}
 
 	if c.Get(h.PSK_USER).(string) != "555555" {
-		t.Errorf("%v was set as psk-account instead of %v", c.Get(h.ACCOUNT_NUMBER).(string), "9876")
+		t.Errorf("%v was set as x-rh-sources-user-id instead of %v", c.Get(h.PSK_USER).(string), "555555")
 	}
 }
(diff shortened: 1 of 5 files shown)


lindgrenj6 (approved):

middleware/user.go
```diff
@@ -11,21 +11,35 @@ import (
...
-
 		tenantId, ok := c.Get(h.TENANTID).(int64)
 		if !ok {
 			return fmt.Errorf("failed to pull tenant from request")
 		}
 
-		userID := xRhIdentity.Identity.User.UserID
-		if userID != "" {
-			user, err := dao.GetUserDao(&tenantId).FindOrCreate(userID)
+		userIDFromContext := ""
```
MikelAlejoBR: Just an optional minor comment but:

```go
userIDFromContext := ""
```

and

```go
var userIDFromContext string
```

are equivalent.

middleware/user.go
```diff
@@ -11,21 +11,35 @@ import (
...
+
+		switch {
+		case c.Get(h.PSK_USER) != nil:
+			userID, ok := c.Get(h.PSK_USER).(string)
+			if !ok {
+				return fmt.Errorf("failed to pull tenant from request")
```
MikelAlejoBR: I think in this case the error would be `failed to pull user id from request`

MikelAlejoBR (approved):
Looking good :+1: !

middleware/user.go
lpichler: good eye! thanks

middleware/user.go
lpichler: that's right 👍 I used var variant

middleware/headers_test.go
```diff
@@ -46,6 +47,10 @@ func TestParseAll(t *testing.T) {
 		t.Errorf("%v was set as psk-account instead of %v", c.Get(h.ACCOUNT_NUMBER).(string), "9876")
 	}
 
+	if c.Get(h.PSK_USER).(string) != "55555" {
```
petracihalova, line 50: the message in t.Errorf() is not updated for PSK_USER

middleware/headers_test.go
```diff
@@ -162,4 +168,8 @@ func TestOnlyPskHeaders(t *testing.T) {
 	if c.Get(h.ACCOUNT_NUMBER).(string) != "9876" {
 		t.Errorf("%v was set as psk-account instead of %v", c.Get(h.ACCOUNT_NUMBER).(string), "9876")
 	}
+
+	if c.Get(h.PSK_USER).(string) != "555555" {
```
petracihalova, line 172: the message in t.Errorf() is not updated for PSK_USER

petracihalova (changes requested):

middleware/headers_test.go
lpichler: thanks I updated both

petracihalova (approved):
👍🏻

lpichler:
there are 3 approvals and from that time there were only changes in error messages and one change in definition/declaration variable so I am merging it now.

PR #429, opened 2022-07-21
Clean up `logger/` package in preparation for context logging through request

lindgrenj6:
This PR might _look_ big but it really isn't that much of a change. I'll spell out the big changes here below.

## First: the "why"
This is in preparation of logging (by context) throughout the entire pipeline - the main point being "I would like to see the application_id/source_id/request_id on each log message through kibana so it is easier to debug"

#### Removal of `haberdasher` support
This one is pretty self-explanatory, haberdasher isn't going to be the solution going forward so I removed the check of sending to stdout/stderr based on the LOG_MODE env var. 

#### Change default LOG_LEVEL from WARN -> INFO
Got this as a request for the eph env - we're overriding it everywhere but when app teams run our application in the Ephemeral Environment they get no output unless something goes horribly wrong, INFO should be enough for them (which skips the database queries).

#### Removal of `LOG_LEVEL_FOR_MIDDLEWARE_LOGS` and `LOG_LEVEL_FOR_SQL_LOGS`
Since under the hood we're using the single logger implementation - even if one of those was set to debug (which it is by default unless we override it) the logs will get discarded unless the "main" log_level is set that low too. So really it doesn't make sense to have a level for each and run a switch against that log level, instead we just always log the `sql` logs as DEBUG, and the middleware logs will be INFO.

#### Removal of External EchoLogger package
I looked at the wrapper and it worked, but it was yet another layer that we don't really need so I made our own wrapper that just overrides the missing methods. Basically it was just the `(Debug|Info|Print|Error|Warn|Fatal)j` methods and then the "special" level methods. I implemented those just printing the `map[string]any` since that is what they're designed to print. We don't use them at all though so it's more just so we can plug our logger into echo.

Since we're using one logger under the hood - I made the level methods just panic (since we should realistically just be setting the level on the "main" logger). The remaining methods just panic as well, but we don't use them.

#### Removal of `LogByLevel` methods from LogWriter & GormLogger
So this kind of ties in with the extra log level removals - instead of switching through lets just log all the things of a certain type at a certain level.

Echo's middleware logger is always going to be INFO
GORM is always going to be using the Trace method, which is going to be DEBUG

### DIFF
--- a/logger/gorm_logger.go
+++ b/logger/gorm_logger.go
@@ -6,36 +6,31 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"gorm.io/gorm"
 	gormLogger "gorm.io/gorm/logger"
 	"gorm.io/gorm/utils"
 )
 
-type CustomGORMLogger struct {
+type GormLogger struct {
 	Logger                  *logrus.Logger
 	SlowThreshold           time.Duration
 	SkipErrorRecordNotFound bool
-	LogLevelForSqlLogs      string
 }
 
-func (l *CustomGORMLogger) LogMode(gormLogger.LogLevel) gormLogger.Interface {
-	return l
-}
+func (l *GormLogger) LogMode(gormLogger.LogLevel) gormLogger.Interface { return l }
 
-func (l *CustomGORMLogger) Info(_ context.Context, logMessage string, data ...interface{}) {
-	l.Logger.Info(logMessage, data)
-}
+// Functions implementing the rest of the gorm logger interface - not used
+// unless you were to call `DB.Logger.Info(...)`, which we don't use. Everything
+// else goes through Trace
+func (l *GormLogger) Debug(_ context.Context, logMessage string, data ...interface{}) {}
+func (l *GormLogger) Info(_ context.Context, logMessage string, data ...interface{})  {}
+func (l *GormLogger) Warn(_ context.Context, logMessage string, data ...interface{})  {}
+func (l *GormLogger) Error(_ context.Context, logMessage string, data ...interface{}) {}
 
-func (l *CustomGORMLogger) Warn(_ context.Context, logMessage string, data ...interface{}) {
-	l.Logger.Warn(logMessage, data)
-}
-
-func (l *CustomGORMLogger) Error(_ context.Context, logMessage string, data ...interface{}) {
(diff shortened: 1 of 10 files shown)


lpichler:
fyi there is  https://issues.redhat.com/browse/RHCLOUD-19049 for this effort.

lindgrenj6:
Ok this is ready for review, this is the "base" for #455 where I basically just merged the loggers a bit more.

The "actual" logging changes are happening in #455.

logger/log_writer.go
```diff
@@ -2,40 +2,22 @@ package logger
...
 	if err != nil {
-		Log.Warn("Unmarshalling Error in LogWriter: " + err.Error())
+		lw.Logger.Warn("Unmarshalling Error in LogWriter: " + err.Error())
 		return 0, err
 	}
 
-	lw.logByLevel(logFields)
+	lw.Logger.WithFields(logFields).Info()
 	return len(data), nil
 }
```
MikelAlejoBR, line 23: I saw this is used in the middleware logger. Will we always log it at the `Info()` level?

logger/gorm_logger.go
```diff
@@ -6,36 +6,31 @@ import (
...
+// else goes through Trace
+func (l *GormLogger) Debug(_ context.Context, logMessage string, data ...interface{}) {}
+func (l *GormLogger) Info(_ context.Context, logMessage string, data ...interface{})  {}
+func (l *GormLogger) Warn(_ context.Context, logMessage string, data ...interface{})  {}
+func (l *GormLogger) Error(_ context.Context, logMessage string, data ...interface{}) {}
 
+// Trace runs a SQL query and logs how long it took as well as the sql executed.
+// By default the log entry is debug, but if the SQL is very slow it will log as
+// warn.
+func (l *GormLogger) Trace(_ context.Context, begin time.Time, fc func() (sql string, rowsAffected int64), err error) {
```
MikelAlejoBR, line 33: Regarding this one: so, does Gorm always call trace when logging SQL queries? On the comment above you say that in order to use any other level you must manually call `Debug|Info|Warn|Error`, so I am assuming that the answer is "yes", but...

MikelAlejoBR (commented):
Looks good to me! I just have a couple of doubts...

lpichler (approved):

PR #433, opened 2022-07-22
[RHCLOUD-20441] refactor: replace query structs with explicit "where" conditions

MikelAlejoBR:
In order to avoid hitting the issue of selecting unexpected data due to
Gorm treating default values as "should not include as where condition",
this commit removes all the structs that were used in queries, and has
replaced them with explicit "where" conditions instead.

## Links

[[RHCLOUD-20441]](https://issues.redhat.com/browse/RHCLOUD-20441)

### DIFF
--- a/dao/application_authentication_dao.go
+++ b/dao/application_authentication_dao.go
@@ -119,13 +119,20 @@ func (a *applicationAuthenticationDaoImpl) List(limit int, offset int, filters [
 }
 
 func (a *applicationAuthenticationDaoImpl) GetById(id *int64) (*m.ApplicationAuthentication, error) {
-	appAuth := &m.ApplicationAuthentication{ID: *id}
-	result := DB.Debug().
-		Where("tenant_id = ?", a.TenantID).First(&appAuth)
-	if result.Error != nil {
+	var appAuth m.ApplicationAuthentication
+
+	err := DB.Debug().
+		Model(&m.ApplicationAuthentication{}).
+		Where("id = ?", *id).
+		Where("tenant_id = ?", a.TenantID).
+		First(&appAuth).
+		Error
+
+	if err != nil {
 		return nil, util.NewErrNotFound("application authentication")
 	}
-	return appAuth, nil
+
+	return &appAuth, nil
 }
 
 func (a *applicationAuthenticationDaoImpl) Create(appAuth *m.ApplicationAuthentication) error {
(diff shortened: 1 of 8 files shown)


dao/application_authentication_dao.go
```diff
@@ -119,12 +119,19 @@ func (a *applicationAuthenticationDaoImpl) List(limit int, offset int, filters [
 }
 
 func (a *applicationAuthenticationDaoImpl) GetById(id *int64) (*m.ApplicationAuthentication, error) {
-	appAuth := &m.ApplicationAuthentication{ID: *id}
-	result := DB.Debug().
-		Where("tenant_id = ?", a.TenantID).First(&appAuth)
-	if result.Error != nil {
+	var appAuth *m.ApplicationAuthentication
```
lindgrenj6: So this is me being a bit pedantic - but I was _very_ confused at first why this was working (GORM does not allocate memory for us). 

Declaring a nil pointer and scanning into it doesn't quite make sense but it looks like GORM handles that and under the hood that pointer (I think it's actually a double pointer?) is getting reassigned to the anonymous struct during the `(*gorm.DB).Model(interface{})` call, here: https://github.com/go-gorm/gorm/blob/6e03b97e266f30db994d8bc24bca2afd74a106b9/callbacks.go#L100

That makes me a bit nervous...could we switch it to declaring the memory here in the DAO and passing the reference to both `.Model` and `.First`?

lindgrenj6 (commented):
Looks good just one thing - I think we should alloc the memory before passing to GORM that way the magic doesn't save us, more details in comment.

dao/application_authentication_dao.go
MikelAlejoBR: Actually, they seem to be using reflection to check the type of the pointer and then, by looking at their model cache, they're able to return which model belongs to: https://github.com/go-gorm/gorm/blob/6e03b97e266f30db994d8bc24bca2afd74a106b9/schema/schema.go#L106-L121

I'll change it though...

lpichler (approved):

dao/application_authentication_dao.go
lindgrenj6: Yes that is true it can figure out the table fine - but I guess what I was trying to say is GORM will return an error if both the `.Model()` and `.First()/.Scan()/.Find()` calls have a nil pointer - under the hood they still expect you to make an allocation from where the call is happening. 

The reason this was working is because it was assigning the anonymous struct from the Model() call to the finisher call, which is a bit funky since the ORM is doing it and not us. It's easier to follow if we explicitly make the allocation and pass it to the finisher ourselves.

PR #438, opened 2022-07-24
[RHCLOUD-20513] Source validation badrequest

petracihalova:
we use function `ValidateSourceCreationRequest()` in bulk create and in single source create handlers

in `SourceCreate()` handler we return Bad Request err in case this validation fails

this PR adds the same for bulk create

**JIRA:** [RHCLOUD-20513](https://issues.redhat.com/browse/RHCLOUD-20513)

### DIFF
--- a/bulk_handlers_test.go
+++ b/bulk_handlers_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/internal/testutils"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/fixtures"
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/request"
+	"github.com/RedHatInsights/sources-api-go/internal/testutils/templates"
 	"github.com/RedHatInsights/sources-api-go/kafka"
 	"github.com/RedHatInsights/sources-api-go/middleware"
 	h "github.com/RedHatInsights/sources-api-go/middleware/headers"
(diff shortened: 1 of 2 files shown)


bulk_handlers_test.go
```diff
@@ -244,6 +245,69 @@ func TestBulkCreate(t *testing.T) {
...
+	badRequestBulkCreate := ErrorHandlingContext(BulkCreate)
+	err = badRequestBulkCreate(c)
+	if err != nil {
+		t.Error(err)
+	}
+
+	templates.BadRequestTest(t, res)
+
+	// Delete created source
+	deletedSource, err := sourceDao.Delete(&src.ID)
```
lpichler, line 302: do we need `cleanSourceForTenant` or sourceDAO#Delete handles run also DeleteCascade ? (I guess there were created more objects in DB with using bulk create )

petracihalova: I believe we don't because in this test only one source is created via simple source create and then I try to create the same source via bulk create but this attempt leads to bad request err because this source already exists ... so that why is not needed the cascade delete

lpichler: I see, thanks!

lpichler (approved):

service/bulk_create.go
```diff
@@ -171,7 +171,7 @@ func parseSources(reqSources []m.BulkCreateSource, tenant *m.Tenant, userResourc
 		// validate the source request
 		err = ValidateSourceCreationRequest(dao.GetSourceDao(&dao.RequestParams{TenantID: &tenant.Id}), &source.SourceCreateRequest)
 		if err != nil {
-			return nil, err
+			return nil, util.NewErrBadRequest(fmt.Sprintf("Validation failed: %s", err.Error()))
```
lindgrenj6: I think you can just use `%v` instead of calling `err.Error()`: https://goplay.space/#RO_ix5fIoIJ

MikelAlejoBR (approved):

service/bulk_create.go
petracihalova: 👍🏻 thanks ... fixed

lindgrenj6 (approved):

PR #450, opened 2022-07-25
Update clowdapp to expose "sources" as apiPath

BlakeHolifield:
By default, the route exposes `sources-api-svc` since Clowder publishes the service name as the route termination point. We can override this with `apiPath: sources`. This will fix an issue where the sources frontend is reaching out to the wrong endpoint since `sources-api-svc/version/graphql` isn't valid.


lindgrenj6 (approved):
Thanks Blake!

PR #451, opened 2022-07-26
[RHCLOUD-20548] chore: bump the "platform-go-middlewares" dependency version

MikelAlejoBR:
Even thought we don't use the library's utility functions to parse and
process the identity header, we think it's a good idea to have it up to
date just in case, before the official "account number shutdown"
arrives.

## Links

[[RHCLOUD-20548]](https://issues.redhat.com/browse/RHCLOUD-20548)


lindgrenj6 (approved):
Interesting that the `System` key went from being a map to actually fleshed out, I like it.

PR #455, opened 2022-07-26
Log all Request Information with each entry to the logger

lindgrenj6:
Continuing on the work I started in #429 

- [x] based on #429, so that needs to go first.

This changes up quite a bit of how the logging works throughout the request flow - the logger is now "unified" so there is no more `log_type` field. 

There is one commit added which displays the work it'll take on the DAO side, from there it should be as easy as just making sure the `RequestParamsFromContext` function just alway populated ctx field if it is non-nil. 

A few notable changes:
- There is now a "custom" context we're using instead of just the plain `echo.Context`, this allows us to override any methods we want. Currently just overriding `Logger()` since we just want to return our own logger with the fields for the current request set on it. 
- As long as `DB.WithContext(dao.ctx)` is called -> the labels will be picked up on the DB
- Similarly, in the handlers we should _always_ be using `c.Logger().<Level>(...)` since that will pull the logger event with the fields set
- No more echo middleware logger necessary - it is handled in the timing middleware now.

--- 

\# TODO:
- [x] Return `request_id` on 5xx errors so we can trace it easier, maybe 400s too? 

----

Examples:
SQL Log on Update, with request information inline:
```json
{"@timestamp":"2022-07-26T10:12:22.47Z","@version":1,"account_number":"1234","app":"source-api-go","caller":"github.com/RedHatInsights/sources-api-go/logger.(*GormLogger).Trace","duration":10.725929,"filename":"/home/jlindgren/cloud/git/crc/sources/go-rewrite/dao/source_dao.go:162","hostname":"mewtwo","labels":{"app":"source-api-go"},"level":"debug","message":"UPDATE \"sources\" SET \"availability_status\"='unavailable',\"created_at\"='2022-07-21 12:05:11.495',\"updated_at\"='2022-07-26 10:12:22.46',\"name\"='yeet',\"uid\"='eb79e9c8-f485-44ee-bf73-6aa984b3e205',\"app_creation_workflow\"='account_authorization',\"source_type_id\"=4,\"tenant_id\"=1 WHERE \"id\" = 9","method":"PATCH","org_id":"1234","request_id":"ef42dafa-c11a-426f-a4b3-c3ebf0b85c32","rows":1,"tags":["source-api-go"],"uri":"/api/sources/v3.1/sources/9","user_agent":"insomnia/2022.3.0"}
```

Event Raising with Request Info:
```json
{"@timestamp":"2022-07-26T10:12:22.47Z","@version":1,"account_number":"1234","app":"source-api-go","caller":"github.com/RedHatInsights/sources-api-go/middleware.RaiseEvent.func1","filename":"/home/jlindgren/cloud/git/crc/sources/go-rewrite/middleware/event_stream.go:52","hostname":"mewtwo","labels":{"app":"source-api-go"},"level":"info","message":"Raising Event Source.update","method":"PATCH","org_id":"1234","request_id":"ef42dafa-c11a-426f-a4b3-c3ebf0b85c32","tags":["source-api-go"],"uri":"/api/sources/v3.1/sources/9","user_agent":"insomnia/2022.3.0"}
```

Timing call (after request is done):
```json
{"@timestamp":"2022-07-26T10:16:36.126Z","@version":1,"account_number":"1234","app":"source-api-go","caller":"github.com/RedHatInsights/sources-api-go/middleware.Timing.func1.1","filename":"/home/jlindgren/cloud/git/crc/sources/go-rewrite/middleware/timings.go:24","hostname":"mewtwo","labels":{"app":"source-api-go"},"latency":1766244,"latency_human":"1.766244ms","level":"info","message":"","method":"DELETE","org_id":null,"request_id":"c92cbe32-d24b-4350-b9c0-2e8d622d4034","tags":["source-api-go"],"uri":"/api/sources/v3.1/sources/9","user_agent":"insomnia/2022.3.0"}
```


lindgrenj6:
On a 500 the request looks like this now:
```json
{
    "errors": [
        {
            "detail": "Internal Server Error: account number or OrgId not present in x-rh-identity",
            "request_id": "d6bd5555-8834-478e-b5df-adbcf6075e8f",
            "status": "500"
        }
    ]
}
```

lindgrenj6:
Ok this is ready to go now - I pulled out the commit that updates the sourceDAO to include a context. The helper function that creates a dao.RequestParam struct now pulls the context if it's present.

util/echo_context.go
```diff
@@ -3,10 +3,34 @@ package util
...
+	// Otherwise just return the base echo "empty" logge
+	if sc.Get("logger") != nil {
+		if entry, ok := sc.Get("logger").(*logrus.Entry); ok {
+			return l.EchoLogger{Entry: entry}
+		} else {
+			return sc.Echo().Logger
+		}
+	} else {
+		return sc.Echo().Logger
+	}
```
MikelAlejoBR: I think you can simplify this to:

```go
if sc.Get("logger") != nil {
	if entry, ok := sc.Get("logger").(*logrus.Entry); ok {
		return l.EchoLogger{Entry: entry}
	}
} 
	
return sc.Echo().Logger
```

MikelAlejoBR (approved):
LGTM! This one is going to help us a ton!

util/echo_context.go
lindgrenj6: Always the hardest part, I sort of wish casting was safe on `nil`, but I get why it isn't.

lpichler (approved):
Great, LGTM👍 :shipit:

lpichler:
I realised this after merging - should not we add `middleware.LoggerField` also for internal endpoints ?

lpichler:
@lindgrenj6 Another thing  - we should look also for account number and org id from `PARSED_IDENTITY` in LoggerFields - now it looks there are not filled.

PR #456, opened 2022-07-26
Add user based filtering for ApplicationAuthenticationDAO struct

lpichler:
- Extract DB.Debug() to func in ApplicationAuthentication Dao struct
- Move tenancy condition to getDB()
- Add user filtering to applicationAuthenticationDAO

Reason why  getDb has parameter is that statement `.Model(&m.ApplicationAuthentication{}).` doesn't work for some usages like DB.Debug().Updates(...) - so it is reason why I leave it optional.

### Links

- issue: https://github.com/RedHatInsights/sources-api-go/issues/356

### DIFF
--- a/dao/application_authentication_dao.go
+++ b/dao/application_authentication_dao.go
@@ -43,7 +43,14 @@ func (a applicationAuthenticationDaoImpl) getDb() *gorm.DB {
 		panic("nil tenant found in applicationAuthentication db DAO")
 	}
 
-	return DB.Debug()
+	query := DB.Debug()
+	query = query.Where("tenant_id = ?", a.TenantID)
+
+	if a.UserID != nil {
+		query = query.Where("user_id IS NULL OR user_id = ?", a.UserID)
+	}
+
+	return query
 }
 
 func (a applicationAuthenticationDaoImpl) getDbWithModel() *gorm.DB {
--- a/dao/application_authentication_dao.go
+++ b/dao/application_authentication_dao.go
@@ -48,6 +48,8 @@ func (a applicationAuthenticationDaoImpl) getDb() *gorm.DB {
 
 	if a.UserID != nil {
 		query = query.Where("user_id IS NULL OR user_id = ?", a.UserID)
+	} else {
+		query = query.Where("user_id IS NULL")
 	}
 
 	return query
--- a/dao/application_authentication_dao.go
+++ b/dao/application_authentication_dao.go
@@ -6,6 +6,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/config"
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
+	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 )
 
@@ -37,6 +38,18 @@ type applicationAuthenticationDaoImpl struct {
 	UserID   *int64
 }
 
+func (a applicationAuthenticationDaoImpl) getDb() *gorm.DB {
+	if a.TenantID == nil {
+		panic("nil tenant found in applicationAuthentication db DAO")
+	}
+
+	return DB.Debug()
+}
+
+func (a applicationAuthenticationDaoImpl) getDbWithModel() *gorm.DB {
+	return a.getDb().Model(&m.ApplicationAuthentication{})
+}
+
 func (a *applicationAuthenticationDaoImpl) ApplicationAuthenticationsByApplications(applications []m.Application) ([]m.ApplicationAuthentication, error) {
 	var applicationAuthentications []m.ApplicationAuthentication
 
@@ -45,8 +58,7 @@ func (a *applicationAuthenticationDaoImpl) ApplicationAuthenticationsByApplicati
 		applicationIDs = append(applicationIDs, value.ID)
 	}
 
-	err := DB.
-		Debug().
+	err := a.getDb().
 		Preload("Tenant").
 		Where("application_id IN ?", applicationIDs).
 		Where("tenant_id = ?", a.TenantID).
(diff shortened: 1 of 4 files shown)


dao/application_authentication_dao.go
```diff
@@ -37,6 +38,21 @@ type applicationAuthenticationDaoImpl struct {
 	UserID   *int64
 }
 
+func (a applicationAuthenticationDaoImpl) getDb(applicationAuthentication *m.ApplicationAuthentication) *gorm.DB {
+	if a.TenantID == nil {
+		panic("nil tenant found in applicationAuthentication db DAO")
+	}
```
lindgrenj6, line 44: I like this - but I also think we should throw a `.Where("tenant_id = ?", a.TenantID` on this query before returning it, same with the `UserID` perhaps. 

---

Sidenote: This is where we're going to be adding the `DB.WithContext(a.ctx)` per #455 after that one gets merged. It's shaping up nicely to have all of the "setup" tasks in this method.

lindgrenj6 (commented):
>Reason why getDb has parameter is that statement .Model(&m.ApplicationAuthentication{}). doesn't work for some usages like DB.Debug().Updates(...) - so it is reason why I leave it optional.

For that - I think rather than making it a parameter we could just use `.Model(&m.ApplicationAuthentication{})` in the actual methods - that way we don't have to check it. This way the `getDb()` function can basically just return the "prepared" DB instance with debugging + tenancy + the current context. 

Great first pass though its mostly here!

dao/application_authentication_dao.go
lindgrenj6: Also thinking about this some more we could maybe have `getDb()` which just contains the setup functions as well as a `getDbWithTable()` which returns the `getDb().Model(&m.ApplicationAuthention{})`. Can use whichever one needed per method. Curious your thoughts on that?

MikelAlejoBR: I would personally vote for having a single `getDb()` with:

```go
db := DB.Debug().
    Model(&applicationAuthentication).
    Where("tenant_id = ?", tenantId)

if userResource {
    db.Where("user_id = ?", userId)
}

return db
```

Maybe I missed a query, but I think we're safe in this DAO if we specify the model for all the queries, right?

MikelAlejoBR (approved):

dao/application_authentication_dao.go
lpichler: > Also thinking about this some more we could maybe have `getDb()` which just contains the setup functions as well as a `getDbWithTable()` which returns the `getDb().Model(&m.ApplicationAuthention{})`. Can use whichever one needed per method. Curious your thoughts on that?

@lindgrenj6 yes I did it initially now I can't remember why did not used - anyway I added it as you suggested (I just name method with getDbWithModel)

lpichler: > I would personally vote for having a single `getDb()` with:
> 
> ```go
> db := DB.Debug().
>     Model(&applicationAuthentication).
>     Where("tenant_id = ?", tenantId)
> 
> if userResource {
>     db.Where("user_id = ?", userId)
> }
> 
> return db
> ```
> 
> Maybe I missed a query, but I think we're safe in this DAO if we specify the model for all the queries, right?

yes that's right I wanted to split it to more PR - but I did at once -I believe it will be still easy to review.

lpichler:
@MikelAlejoBR @lindgrenj6 Thanks for you insights - I added tenancy to getDB method and also userid filtering + added tests for user based filtering.

dao/application_authentication_dao.go
```diff
@@ -37,6 +38,25 @@ type applicationAuthenticationDaoImpl struct {
...
+func (a applicationAuthenticationDaoImpl) getDb() *gorm.DB {
+	if a.TenantID == nil {
+		panic("nil tenant found in applicationAuthentication db DAO")
+	}
+
+	query := DB.Debug()
+	query = query.Where("tenant_id = ?", a.TenantID)
+
+	if a.UserID != nil {
+		query = query.Where("user_id IS NULL OR user_id = ?", a.UserID)
```
MikelAlejoBR, line 50: The `IS NULL` bit maybe is a leftover? Because I think that would match the following:

| id | tenant_id | app_id | auth_id | user_id |
|----|-----------|--------|---------|---------|
| 1  | 12345     | 999    | 888     | 101010    |

Even if the specified user ID is different from `101010`.

**EDIT**: never mind, was clarified below.

MikelAlejoBR (approved):
LGTM, just a comment

lindgrenj6:
That's actually what we want though - because if the user_id is null that
means it's not a user-based source but still in the tenant. e.g. a
cloud-meter source would show up as well as an app-studio one.
--
Jacob Lindgren

Red Hat

MikelAlejoBR:
> That's actually what we want though - because if the user_id is null that means it's not a user-based source but still in the tenant. e.g. a cloud-meter source would show up as well as an app-studio one.

That's right, yeah, I was tunnel visioning. Thank you!

lindgrenj6 (approved):
Nice, this will make it so easy to add the context to the queries!

lpichler:
I forgot to cover case where UserID is not present in request. 

This can happen for example for requests with PSK - when `x-rh-sources-user-id` header is not present in request.

I believe that in this case we should list records without ownership only (column `user_id==NULL`):
https://github.com/RedHatInsights/sources-api-go/pull/456/commits/7b3f120645f9ddfd7add9f46d65defe7b714b4f3#diff-e44569a4d110df92cdbe0c0dc8f996ce5cee5a46031f0d7dff34ff43c263a25eR52

@lindgrenj6 @MikelAlejoBR Is it OK for you ?

MikelAlejoBR (approved):
LGTM, yes!

petracihalova (approved):
LGTM

lindgrenj6 (approved):
Yep I agree on the side of caution - if there isn't a userID then don't show user based resources.

PR #461, opened 2022-07-27
Remove GraphQL Proxying Capability since the old app is decommissioned

lindgrenj6:
Just pulling this out since the Ruby app has been gone for a while now and the UI is using the new GraphQL implementation.


MikelAlejoBR (approved):
Nice :+1:

lpichler (approved):

PR #462, opened 2022-07-27
[RHCLOUD-19924] Change sql query in GetByName methods in app_type/source_type dao

Ellen-Yi-Dong:
What: Within source_type_dao.go and application_type_dao.go, changed the sql query in GetByName methods and return a bad request error if more than one application type or source type name is found. 
Why: in certain cases the desired app or source type may not be selected if there are multiple rows with the same name. 

JIRA: [RHCLOUD-19924](https://issues.redhat.com/browse/RHCLOUD-19924)

### DIFF
--- a/dao/application_type_dao.go
+++ b/dao/application_type_dao.go
@@ -112,7 +112,6 @@ func (at *applicationTypeDaoImpl) GetByName(name string) (*m.ApplicationType, er
 		return nil, util.NewErrNotFound("application type")
 	}
 	return &appTypes[0], nil
-
 }
 
 func (at *applicationTypeDaoImpl) Create(_ *m.ApplicationType) error {
--- a/dao/application_type_dao.go
+++ b/dao/application_type_dao.go
@@ -98,13 +98,21 @@ func (at *applicationTypeDaoImpl) GetById(id *int64) (*m.ApplicationType, error)
 }
 
 func (at *applicationTypeDaoImpl) GetByName(name string) (*m.ApplicationType, error) {
-	appType := &m.ApplicationType{}
-	result := DB.Debug().Where("name LIKE ?", "%"+name+"%").First(&appType)
+	appTypes := make([]m.ApplicationType, 0)
+	result := DB.Debug().
+		Where("name LIKE ?", "%"+name+"%").
+		Find(&appTypes)
+
 	if result.Error != nil {
+		return nil, result.Error
+	}
+	if result.RowsAffected > int64(1) {
+		return nil, util.NewErrBadRequest("Found more than one of the same application type name")
+	} else if result.RowsAffected == int64(0) {
 		return nil, util.NewErrNotFound("application type")
 	}
+	return &appTypes[0], nil
 
-	return appType, nil
 }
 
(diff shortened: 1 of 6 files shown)


dao/application_type_dao_test.go
```diff
@@ -127,3 +126,22 @@ func TestApplicationTypeGetByNameNotFound(t *testing.T) {
 
 	DropSchema("app_type_by_name")
 }
+func TestApplicationTypeGetByNameBadRequest(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
+	SwitchSchema("app_type_by_name")
+	//wantAppType := m.ApplicationType{Name: "not existing name"}
```
MikelAlejoBR: I think you forgot to remove this comment

MikelAlejoBR (approved):
LGTM! Just a very minor comment

lpichler:
@Ellen-Yi-Dong looks nice - you can also write better commit messages, I like style:


1. line: "briefly describe change and where it is"
2. line: <empty line>
Other lines: more detailed description 

(where it is - could be file, struct or other way how to localise change)


I also like that @MikelAlejoBR adds <type> of change as first word in 1.line. Like refactor, fix, chore. (I should get use on this style and use it also)
@MikelAlejoBR what does `chore` mean ? 

So for your first commit it could be:

```
Handle case when query deals with more application types in applicationTypeDaoImp#GetByName

More more info ..

```

it is because when you looks on commits - you can see quickly what is change about.

Hint: you can write commit message in such way which you can use it in description of PR and then you basically don't need rephrase the description again.

petracihalova (approved):
nice job 🎉

MikelAlejoBR:
> I also like that @MikelAlejoBR adds of change as first word in 1.line. Like refactor, fix, chore. (I should get use on this style and use it also) @MikelAlejoBR what does `chore` mean ?

I try to follow [Angular's commit style convention](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#type), and for all those commits I don't know in which category they fit in, I was using "chore", which is like a "tedious but required thing that has to be done", like upgrading a dependency so that you can refactor some code.

However I see that there's no "chore" category on that commit convention, so I might have to start forgetting about that :eyes:

dao/application_type_dao.go
```diff
@@ -98,13 +98,21 @@ func (at *applicationTypeDaoImpl) GetById(id *int64) (*m.ApplicationType, error)
...
+
 	if result.Error != nil {
+		return nil, result.Error
+	}
+	if result.RowsAffected > int64(1) {
+		return nil, util.NewErrBadRequest("Found more than one of the same application type name")
+	} else if result.RowsAffected == int64(0) {
 		return nil, util.NewErrNotFound("application type")
 	}
+	return &appTypes[0], nil
```
lindgrenj6: [minor]: can you remove this teeny empty line here at the end of the method?

lindgrenj6 (commented):
Looks great! Just one tiny comment.

MikelAlejoBR (approved):

lpichler (approved):
Looks great, Thanks for tests! 👏

PR #466, opened 2022-07-28
Filter by user in sourceDaoImpl

lpichler:
This PR moves `DB.Debug().` to the functions and adds tenancy and user conditions to common place.

Unlike PR https://github.com/RedHatInsights/sources-api-go/pull/456 there is more functions which are used with getDb() functions.

We have to deal with some occurences `DB.Debug().` where we needed specify table for tenancy and user condition or use instances for `*gorm.DB`.

I added some functions which covers it and we have still one common method where tenancy and user, etc have setup.

EDIT: it is just one commit  - I had not luck with rebasing and splittin up to commits.

### DIFF
--- a/dao/source_dao.go
+++ b/dao/source_dao.go
@@ -39,6 +39,49 @@ type sourceDaoImpl struct {
 	UserID   *int64
 }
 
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (approved):
LGTM!

lindgrenj6 (approved):

PR #473, opened 2022-07-29
Add UserCatcher middleware to missing routes

lpichler:
During preparation for DEMO I realised that UserCatcher middleware is missing for bulk create. 

But I believe it should be also in other routes.


MikelAlejoBR (approved):

lindgrenj6 (approved):
We _really really_ need to make a new group for tenancy + user catcher and move idvalidation to the top layer of middleware, but approving for now.

PR #474, opened 2022-07-29
[RHCLOUD-20569] Authentication handlers tests update 3

petracihalova:
authentication handlers tests update PART III.
(part I. #465, part II. #471)

for handler:
- AuthenticationCreate()

tests update for other handlers will be covered in next PR

main goal is to check the tenancy and add missing tests

**JIRA:** [RHCLOUD-20569](https://issues.redhat.com/browse/RHCLOUD-20569)

PR #476, opened 2022-08-01
[RHCLOUD-20638] Add check if source exists into authentication create method for resource type = source

petracihalova:
when we create new authentication for resource type = source, we need to check if the given source exists and if not, then we should return bad request (bad source id was provided)

until now it was possible to create an authentication with resource type = source even if the source didn't exist

for resource type = endpoint and application this works correctly

**JIRA:** [RHCLOUD-20638](https://issues.redhat.com/browse/RHCLOUD-20638)

### DIFF
--- a/authentication_handlers_test.go
+++ b/authentication_handlers_test.go
@@ -425,6 +425,46 @@ func TestAuthenticationCreateBadRequestInvalidResourceType(t *testing.T) {
 	templates.BadRequestTest(t, rec)
 }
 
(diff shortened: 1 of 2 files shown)


petracihalova:
this PR wil be rebased after #474 merge

petracihalova:
rebased and ready for review

lindgrenj6 (approved):

MikelAlejoBR (approved):
LGTM!

lpichler (approved):

PR #471, opened 2022-07-29
[RHCLOUD-20569] Authentication handlers tests update 2

petracihalova:
authentication handlers tests update PART II.
(part I. #465)

for handler:
- AuthenticationGet()

tests update for other handlers will be covered in next PR

main goal is to check the tenancy and add missing tests

I am avare about the TestAuthenticationGet() is updated to be working with secret store vault and database and rest of tests not -> I believe we can update all the authentication tests after we will know if we will support the Vault

**JIRA:** [RHCLOUD-20569](https://issues.redhat.com/browse/RHCLOUD-20569)

PR #477, opened 2022-08-01
[RHCLOUD-20569] Authentication handlers tests update 4

petracihalova:
authentication handlers tests update PART IV.
(part I. #465, part II. #471, part III. #474)

for handler:
- AuthenticationCreate()

added test for the request with wrong field's name

tests update for other handlers will be covered in next PR

**JIRA:** [RHCLOUD-20569](https://issues.redhat.com/browse/RHCLOUD-20569)

### DIFF
--- a/authentication_handlers_test.go
+++ b/authentication_handlers_test.go
@@ -425,6 +425,31 @@ func TestAuthenticationCreateBadRequestInvalidResourceType(t *testing.T) {
 	templates.BadRequestTest(t, rec)
 }
 
+// TestAuthenticationCreateBadRequestInvalidRequest tests that bad request err is returned
+// for request that contains wrong field name
+func TestAuthenticationCreateBadRequestInvalidRequest(t *testing.T) {
+	tenantId := int64(1)
+	rawJSON := `{"wrong_field_name": "psw"}`
+
+	c, rec := request.CreateTestContext(
+		http.MethodPost,
+		"/api/sources/v3.1/authentications",
+		bytes.NewReader([]byte(rawJSON)),
+		map[string]interface{}{
+			"tenantID": tenantId,
+		},
+	)
+	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
+
+	badRequestAuthenticationCreate := ErrorHandlingContext(AuthenticationCreate)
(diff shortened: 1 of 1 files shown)


petracihalova:
this PR will be rebased after #476 merge

petracihalova:
rebased and ready for review

lpichler (approved):

MikelAlejoBR (approved):
Looks good to me, although I'm confused. Here it feels like we are ultimately testing the binder, and don't we already do that in [binder_test.go](https://github.com/RedHatInsights/sources-api-go/blob/main/binder_test.go)?

petracihalova:
> Looks good to me, although I'm confused. Here it feels like we are ultimately testing the binder, and don't we already do that in [binder_test.go](https://github.com/RedHatInsights/sources-api-go/blob/main/binder_test.go)?

well, yes and no ... my main idea was to write another test for AuthenticationCreate() that ends with bad request err on line 94 (see the picture). This should be part of PR #474 but I forgot to include it. But maybe for this part of AuthenticationCreate() it is enough to have the test in binder_test.go. 🤔
image

@lindgrenj6, @lpichler what do you think?

lindgrenj6:
Hmm. Yeah that would just be falling under a similar test to the binding test. In order to get that error to trigger it would have to be invalid json in the `extra` field.

I'm not actually sure we'll be able to effectively test that since _technically_ if extra is invalid json the bind will fail - but I'm not entirely sure. In my local testing using something like this:

```
curl --header "x-rh-sources-psk: 1234" \
     --header "x-rh-sources-account-number: 1234" \
     murkrow:8000/api/sources/v3.1/authentications  \
     --data-raw '{"resource_type": "Source", "resource_id": "4", "extra": {"thing":a}}'
```
with the extra being invalid it gets caught by the initial bind. In my opinion this is acceptable - I assume the only time that error would happen is if something went wrong with Go itself which we still need to handle.

petracihalova:
> with the extra being invalid it gets caught by the initial bind. In my opinion this is acceptable - I assume the only time that error would happen is if something went wrong with Go itself which we still need to handle.

agree ... so I think I can close this PR because for err returned from Bind() function we have tests in different file ... thanks for this discussion

PR #481, opened 2022-08-01
Move around middleware to try and make it a bit less duplicative

lindgrenj6:
#### First Commit
Just an idea - since the idvalidation is going to be ran basically everywhere we should always run it - involved a teeny change to make sure the param isn't `uid` but that allows us to skip where necessary. 

#### Second Commit
Moves `middleware.Tenancy` and `middleware.UserCatcher` into a group - this way we don't have to duplicate that everywhere.

### DIFF
--- a/middleware/id_validation.go
+++ b/middleware/id_validation.go
@@ -18,7 +18,7 @@ var idValidationRegex = regexp.MustCompile(`^\d+$`)
 func IdValidation(next echo.HandlerFunc) echo.HandlerFunc {
 	return func(c echo.Context) error {
 		for _, param := range c.ParamNames() {
-			if strings.HasSuffix(param, "id") {
+			if strings.HasSuffix(param, "id") && param != "uid" {
 				err := validateId(c, param)
 				if err != nil {
 					return util.NewErrBadRequest(err)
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (approved):
Looks good to me, yeah. It indeed looks a little bit cleaner

lindgrenj6:
@Fynardo it looks like some of the tests are getting caught by the id validation now! It's expecting a 404 when it's a 400. Maybe we can get that updated?

Fynardo:
@lindgrenj6  In every case? numeric values (-1, 0) as well?

lindgrenj6:
Actually good point - I think it's only failing on the authentication tests. We change how we validate the ID's whether we're using the postgres db store (ints) or vault (strings). I'll validate thats the case.

lpichler:
This is definitely improvement. Other thing what came to my mind is to create some configurable way, for example in yaml file:


```yaml
---
all:
  middlewares:
   - timing
   - handle_errors
  routes:
     "/bulk_create":
        handler:
           - BulkCreate
        middlewares:
          - permission_check
          - raise_event
    "/sources":
        handler:
           - BulkCreate
        middlerwares:
          - permission_check
          - raise_event

..
```
 
I can't imaginate code which would process this and adds routes like we now in `routes.go` but it might be more readable.

Anyway I think it is enough as we have it now and in this PR.

petracihalova (approved):
LGTM

lpichler (approved):

PR #485, opened 2022-08-03
[RHCLOUD-20666] fix: tenants are created empty when processing anemic tenants

MikelAlejoBR:
Whenever we received an "x-rh-identity" header we were saving both the
account number and the org id in the context as well. After that, due to
the account number being first in the tenancy middleware's switch case,
that code block would get executed to create tenants.

Therefore, both in anemic tenants or in tenants with both an EBS account
number and org ids, the tenants were only being created with an EBS
account number.

## Links

[[RHCLOUD-20666]](https://issues.redhat.com/browse/RHCLOUD-20666)


lpichler (approved):
LGTM! Good catch!

petracihalova (approved):
LGTM

lindgrenj6 (approved):

PR #486, opened 2022-08-03
[RHCLOUD-20671] refactor: centralize tenancy lookups on the identity struct

MikelAlejoBR:
Until now, we were "individually" looking up the tenants in three ways:

1. By creating an identity struct from the EBS account number.
2. By creating an identity struct from the OrgId.
3. By passing the identity struct directly.

Since the third way includes the two above, I thought it might be a good
idea to just use the identity struct for everything, and to forget about
pulling specific data for the tenancy.

This will make the "x-rh-identity" header the favoured in case we
receive it along the "ebs account number" and "orgid" headers.

## Links

[[[RHCLOUD-20671]](https://issues.redhat.com/browse/RHCLOUD-20671)

### DIFF
--- a/middleware/headers.go
+++ b/middleware/headers.go
@@ -6,6 +6,7 @@ import (
 	h "github.com/RedHatInsights/sources-api-go/middleware/headers"
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
+	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
 /*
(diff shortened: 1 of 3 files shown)


middleware/headers.go
```diff
@@ -42,44 +43,43 @@ func ParseHeaders(next echo.HandlerFunc) echo.HandlerFunc {
...
+			id = genId
+		} else {
+			xRhIdentity, err := util.ParseXRHIDHeader(xRhIdentityRaw)
+			if err != nil {
+				return fmt.Errorf("could not extract identity from header: %w", err)
+			}
 
-			// store the account number + org_id for easy usage later
-			c.Set(h.ACCOUNT_NUMBER, xRhIdentity.Identity.AccountNumber)
-			c.Set(h.ORGID, xRhIdentity.Identity.OrgID)
```
lindgrenj6, line 60: We will need to update the LoggerFields middleware to pull from the identity struct too since these fields aren't on the request.

lindgrenj6 (commented):
I like this refactor - if we can only focus on _one_ field in the context that would be better. I think the ultimate refactor in this line is combining this + our internal PKs for tenant/user into a struct. But this is a good first step.

---

One minor comment for the logging - otherwise I'm good with it.

lpichler:
This needs rebase.

middleware/headers.go
MikelAlejoBR: In fact I'm recovering these in the tenancy middleware, because I rather not do a typecheck on the logger middleware...

MikelAlejoBR:
> I like this refactor - if we can only focus on _one_ field in the context that would be better. I think the ultimate refactor in this line is combining this + our internal PKs for tenant/user into a struct. But this is a good first step.

Yep, initially I developed it that way, but I think I was going off rails with the purpose of the PR, so I decided to just stick to this little refactor.

lpichler (approved):

lindgrenj6 (approved):
LGTM!

PR #492, opened 2022-08-04
[RHCLOUD-19375] Add delete test for user based sources 

lpichler:
It is good that check whether user is "user based" is in 
`SourceDelete` handler before we call `service.DeleteCascade`
and also before we call `service.DeleteCascade` in SuperKeyDestroySource 
middleware in method IsSuperkey - it means that we don't have to
solve user based scope in `service.DeleteCascade` whether we can
delete user based source or not.

So this PR adds only tests.

API Test for delete actions:
- Add handler test with disallowed deletion of user based source
- Add handler test with disallowed destroy of super key user based source
- Add handler test for deletion source with ownership

### Links
https://issues.redhat.com/browse/RHCLOUD-20678
https://issues.redhat.com/browse/RHCLOUD-19375

### DIFF
--- a/source_handlers_test.go
+++ b/source_handlers_test.go
@@ -2661,6 +2661,161 @@ func TestSourceEditPausedUnit(t *testing.T) {
 	}
 }
 
+func TestSourceDeleteWithOwnershipWhenUserIsNotAllowedToDelete(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
+	testutils.SkipIfNotSecretStoreDatabase(t)
+
+	accountNumber := "112567"
+	userIDWithOwnRecords := "user_based_user"
+	otherUserIDWithoutOwnRecords := "other_user"
+	applicationTypeID := fixtures.TestApplicationTypeData[3].Id
+	sourceTypeID := fixtures.TestSourceTypeData[2].Id
+	recordsWithUserID, _, err := dao.CreateSourceWithSubResources(sourceTypeID, applicationTypeID, accountNumber, &userIDWithOwnRecords)
+	if err != nil {
+		t.Errorf("unable to create source: %v", err)
+	}
+
+	src := recordsWithUserID.Sources[0]
+	tenantID := src.TenantID
+
+	otherUser, err := dao.CreateUserForUserID(otherUserIDWithoutOwnRecords, tenantID)
+	if err != nil {
+		t.Errorf("unable to create user: %v", err)
+	}
+
+	id := fmt.Sprintf("%d", src.ID)
+	c, rec := request.CreateTestContext(
+		http.MethodDelete,
+		"/api/sources/v3.1/sources/"+id,
+		nil,
+		map[string]interface{}{
+			"tenantID": tenantID,
+			"userID":   otherUser.Id,
+		},
+	)
+
+	c.SetParamNames("id")
+	c.SetParamValues(id)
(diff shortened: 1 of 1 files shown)


MikelAlejoBR (approved):
Nice!

source_handlers_test.go
```diff
@@ -1305,6 +1305,57 @@ func TestSourceEditBadRequest(t *testing.T) {
 	templates.BadRequestTest(t, rec)
 }
 
+func TestSourceDeleteWithOwnershipWhenUserIsNotAllowedToDelete(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
+	testutils.SkipIfNotSecretStoreDatabase(t)
+
+	accountNumber := "112567"
+	userIDWithOwnRecords := "user_based_user"
+	otherUserIDWithOwnRecords := "other_user"
```
petracihalova: for this user you dont create own records (I know that it is not needed), but then this user is not "WithOwnRecords", or?)

source_handlers_test.go
```diff
@@ -1356,6 +1356,70 @@ func TestSourceDeleteWithOwnershipWhenUserIsNotAllowedToDelete(t *testing.T) {
...
+	)
+
+	c.SetParamNames("id")
+	c.SetParamValues(id)
+	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
+
+	superKeyDestroySource := middleware.SuperKeyDestroySource(SourceDelete)
+	err = superKeyDestroySource(c)
+	if err == nil {
+		t.Error(err)
```
petracihalova: I think you can use here Is() method for compare the err object
```
	err = superKeyDestroySource(c)
	if !errors.Is(err, util.ErrNotFoundEmpty) {
		t.Errorf("improper error occurred for super key destroy source operation %v", err)
	}
```

source_handlers_test.go
```diff
@@ -1356,6 +1356,70 @@ func TestSourceDeleteWithOwnershipWhenUserIsNotAllowedToDelete(t *testing.T) {
...
 	}
 }
+
+func TestSuperKeyDestroyWithOwnershipWhenUserIsNotAllowedToDelete(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
+	testutils.SkipIfNotSecretStoreDatabase(t)
+
+	accountNumber := "112567"
+	userIDWithOwnRecords := "user_based_user"
+	otherUserIDWithOwnRecords := "other_user"
```
petracihalova: same comment as for first test

source_handlers_test.go
```diff
@@ -1420,6 +1420,44 @@ func TestSuperKeyDestroyWithOwnershipWhenUserIsNotAllowedToDelete(t *testing.T)
...
+	)
+
+	c.SetParamNames("id")
+	c.SetParamValues(id)
+	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
+
+	err = SourceDelete(c)
+	if err != nil {
+		t.Error(err)
+	}
```
petracihalova: just thinking 🤔 in this case new user was created for this test? so after deletion of the source the expected result it that the source and all subresources are deleted but the user is still in database? if this test creates new user, maybe you should delete this user in the end of the test ... but maybe it is not needed ...but I'm just thinking out loud 😇

lpichler:
@petracihalova I addressed your comments, thanks!

lindgrenj6 (approved):

PR #497, opened 2022-08-08
[RHCLOUD-19948] feature: validate "endpoint edit" requests

MikelAlejoBR:
Makes sure that no endpoint can be edited with an availability status set to empty or null.

## Links

### DIFF
--- a/service/endpoint_validation.go
+++ b/service/endpoint_validation.go
@@ -22,7 +22,13 @@ var fqdnRegexp = regexp.MustCompile(`^(?:[a-zA-Z\d](?:[a-zA-Z\d-]*[a-zA-Z\d])?\.
 
 // schemeRegexp matches a valid scheme, as per RFC 3986
 var schemeRegexp = regexp.MustCompile(`^[a-zA-Z][a-zA-Z\d+\-.]*$`)
-var validAvailabilityStatuses = []string{"", model.Available, model.Unavailable}
+
+// validAvailabilityStatuses is a map containing the valid availability statuses. It has this form because a map is
+// faster for these lookups.
+var validAvailabilityStatuses = map[string]struct{}{
+	model.Available:   {},
+	model.Unavailable: {},
+}
 
 const (
 	defaultScheme    = "https"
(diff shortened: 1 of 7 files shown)


service/endpoint_validation.go
```diff
@@ -97,8 +103,72 @@ func ValidateEndpointCreateRequest(dao dao.EndpointDao, ecr *model.EndpointCreat
 		return fmt.Errorf("the certificate authority cannot be empty")
 	}
 
-	if !util.SliceContainsString(validAvailabilityStatuses, ecr.AvailabilityStatus) {
-		return fmt.Errorf("invalid availability status")
+	// The team decided that the availability statuses will default to "in_progress" whenever they come empty, since
+	// setting them as "unavailable" by default may lead to some confusion to the calling clients.
+	if ecr.AvailabilityStatus == "" {
+		ecr.AvailabilityStatus = model.InProgress
```
lindgrenj6, line 101: I see we don't have InProgress in the validAvailabilityStatuses map above - I also think we should just have one instance of that if we could (probably in the model package).

MikelAlejoBR: Yes, I checked the OpenApi spec and I saw that we only allow the "available" and "unavailable" statuses for the endpoint. That's why I borrowed the "in_progress" one from the model package, and created a custom map for the endpoint's valid statuses.

Do I simply merge "all the structs" into one and let the endpoints have those two values too?

service/endpoint_validation.go
```diff
@@ -22,7 +22,13 @@ var fqdnRegexp = regexp.MustCompile(`^(?:[a-zA-Z\d](?:[a-zA-Z\d-]*[a-zA-Z\d])?\.
...
 // schemeRegexp matches a valid scheme, as per RFC 3986
 var schemeRegexp = regexp.MustCompile(`^[a-zA-Z][a-zA-Z\d+\-.]*$`)
-var validAvailabilityStatuses = []string{"", model.Available, model.Unavailable}
+
+// validAvailabilityStatuses is a map containing the valid availability statuses. It has this form because a map is
+// faster for these lookups.
+var validAvailabilityStatuses = map[string]struct{}{
+	model.Available:   {},
+	model.Unavailable: {},
+}
```
petracihalova: sorry if I missed some discussions/explanations about this ... but why you define here the same struct as is in model/availability_status.go? 🤔🤔🤔

service/endpoint_validation_test.go
```diff
@@ -416,3 +465,397 @@ func TestInvalidAvailabilityStatuses(t *testing.T) {
...
+	editRequest, sourceId := setUpEndpointEditRequest()
+
+	// The "longHostname" variable holds a 256 char hostname
+	tmp := `aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeee
+		aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeee
+		aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeee
+		aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeee
+		aaaaaaaaaabbbbbbbbbbccccccccccddddddddddeeeeeeeeee
+		aaaaaa
+	`
```
petracihalova: you can use here Repeat() method too
`strings.Repeat("a", 256)`

service/endpoint_validation_test.go
```diff
@@ -416,3 +465,397 @@ func TestInvalidAvailabilityStatuses(t *testing.T) {
...
+}
+
+// TestEditEmptyHost tests if no error is returned even when an empty host is given. Tested on an "edit endpoint"
+// payload.
+func TestEditEmptyHost(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
+
+	editRequest, sourceId := setUpEndpointEditRequest()
+	tmp := ""
+	editRequest.Host = &tmp
```
petracihalova: in cases like this you can find handy the StringRef() method from util package
`editRequest.Host = util.StringRef("")`

petracihalova (approved):
nice job, I have only tiny comments

service/endpoint_validation.go
MikelAlejoBR: Yeah, we were discussing it Jacob and I in [the above thread](https://github.com/RedHatInsights/sources-api-go/pull/497#discussion_r951782242). This comes from the endpoints only allowing those two availability statuses, since the "in_progress" and "partially_available" don't make sense for the endpoints. I'm going to close this one if you don't mind to keep discussing this on a single thread :smile:

petracihalova: great thanks

service/endpoint_validation_test.go
MikelAlejoBR: TIL! This is very helpful! Thanks!

service/endpoint_validation_test.go
MikelAlejoBR: Uugggh!!! I keep forgetting about the utility. Thanks!

petracihalova: but I think it is not needed to change it here ... I wanted only let you know that we have it

MikelAlejoBR: Well I did it anyway, it was just a quick change :)

service/endpoint_validation.go
lindgrenj6: I guess I more meant that we have this map here in this file: https://github.com/MikelAlejoBR/sources-api-go/blob/acedd66999d31f744f5475dc3c8186318ea2869f/service/endpoint_validation.go#L26-L31

And also this map over in model: https://github.com/MikelAlejoBR/sources-api-go/blob/acedd66999d31f744f5475dc3c8186318ea2869f/model/availability_status.go#L20-L25

Could we just change this line below to pull from the model package rather than duplicate the map? 
```diff
diff --git a/service/endpoint_validation.go b/service/endpoint_validation.go
index 5afa823..141c125 100644
--- a/service/endpoint_validation.go
+++ b/service/endpoint_validation.go
@@ -107,7 +107,7 @@ func ValidateEndpointCreateRequest(dao dao.EndpointDao, ecr *model.EndpointCreat
 	if ecr.AvailabilityStatus == "" {
 		ecr.AvailabilityStatus = model.InProgress
 	} else {
-		if _, ok := validAvailabilityStatuses[ecr.AvailabilityStatus]; !ok {
+		if _, ok := model.ValidEndpointAvailabilityStatuses[ecr.AvailabilityStatus]; !ok {
 			return fmt.Errorf("invalid availability status")
 		}
 	}
```

MikelAlejoBR: Oops, I thought one of them had the 4 availability statuses... I pushed the change!

service/endpoint_validation.go
```diff
@@ -97,8 +95,71 @@ func ValidateEndpointCreateRequest(dao dao.EndpointDao, ecr *model.EndpointCreat
...
 	}
 
-	if !util.SliceContainsString(validAvailabilityStatuses, ecr.AvailabilityStatus) {
-		return fmt.Errorf("invalid availability status")
+	// The team decided that the availability statuses will default to "in_progress" whenever they come empty, since
+	// setting them as "unavailable" by default may lead to some confusion to the calling clients.
+	if ecr.AvailabilityStatus == "" {
+		ecr.AvailabilityStatus = model.InProgress
+	} else {
+		if _, ok := model.ValidAvailabilityStatuses[ecr.AvailabilityStatus]; !ok {
```
lindgrenj6: Should this be `model.ValidEndpointAvailabilityStatuses[ecr.AvailabilityStatus]`?

service/endpoint_validation.go
```diff
@@ -97,8 +95,71 @@ func ValidateEndpointCreateRequest(dao dao.EndpointDao, ecr *model.EndpointCreat
...
+		tmp := defaultVerifySsl
+		editRequest.VerifySsl = &tmp
+	}
+
+	if *editRequest.VerifySsl && (editRequest.CertificateAuthority == nil || *editRequest.CertificateAuthority == "") {
+		return fmt.Errorf("the certificate authority cannot be empty")
+	}
+
+	if editRequest.AvailabilityStatus != nil {
+		if _, ok := model.ValidAvailabilityStatuses[*editRequest.AvailabilityStatus]; !ok {
```
lindgrenj6: Same here.

lpichler (approved):
LGTM 👍

service/endpoint_validation.go
MikelAlejoBR: Ooops, you're right. Thanks!

service/endpoint_validation.go
MikelAlejoBR: Done!

lindgrenj6 (approved):

PR #498, opened 2022-08-08
[RHCLOUD-19948] feature: validate "edit source" requests

MikelAlejoBR:
Makes sure that no source can be edited with an availability status set to empty or null.

## Links

### DIFF
--- a/service/source_validation.go
+++ b/service/source_validation.go
@@ -36,8 +36,14 @@ func ValidateSourceCreationRequest(sourceDao dao.SourceDao, req *model.SourceCre
 		req.AppCreationWorkflow = model.ManualConfig
 	}
 
-	if !util.SliceContainsString(model.AvailabilityStatuses, req.AvailabilityStatus) {
-		return fmt.Errorf("invalid status")
+	// The team decided that the availability statuses will default to "in_progress" whenever they come empty, since
+	// setting them as "unavailable" by default may lead to some confusion to the calling clients.
+	if req.AvailabilityStatus == "" {
+		req.AvailabilityStatus = model.InProgress
+	} else {
+		if _, ok := model.ValidAvailabilityStatuses[req.AvailabilityStatus]; !ok {
+			return fmt.Errorf("invalid status")
+		}
 	}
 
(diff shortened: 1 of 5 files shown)


service/source_validation.go
```diff
@@ -55,3 +61,22 @@ func ValidateSourceCreationRequest(dao dao.SourceDao, req *model.SourceCreateReq
...
 }
+
+func ValidateSourceEditRequest(dao dao.SourceDao, editRequest *model.SourceEditRequest) error {
+	if editRequest.Name == nil || *editRequest.Name == "" {
+		return errors.New("name cannot be empty")
+	}
+
+	if dao.NameExistsInCurrentTenant(*editRequest.Name) {
+		return errors.New("name already exists in tenant")
+	}
```
petracihalova: I know that this PR is not ready for review but I found out that you solve here the same name validation as was solved here https://github.com/RedHatInsights/sources-api-go/pull/507 (already merged) but the name of function is slightly different so I wanted to let you know
+
for edit request I believe the "Name" field can be nil (when you dont want to edit the name) so the logic should be updated here ... we need to cover situation when the "name" field is not part of body request and when the "name" field is there and is an empty string 🤔

MikelAlejoBR: In this case I think it is better to just run the validations if the incoming name isn't `nil` — if it is nil, we know it wasn't included in the payload, so we don't need to modify it. If it is not `nil`, even if it was an accidental empty string, we enforce the validation.

petracihalova: yes and only for info: this is solved in @Ellen-Yi-Dong #544 too

MikelAlejoBR: Mmmmmmmmmmm... I'm doubting on whether waiting until that PR is merged and then rebasing this one...

Ellen-Yi-Dong (approved):

Ellen-Yi-Dong (commented):
LGTM!

petracihalova (approved):
LGTM

lindgrenj6 (approved):

lindgrenj6:
Looks like we got a conflict, will merge once it's solved!

MikelAlejoBR:
All right everything's green!

PR #499, opened 2022-08-08
[RHCLOUD-19948] feature: not null constraint and default value for "availability_status"

MikelAlejoBR:
This migration adds a not null contraint and a default value for all the
tables which have an "availability_status" column. The default value is
"in_progress", a value that was agreed as the default one by the team.

It also takes care of all the current NULL or empty values and sets them
to the default value too.

## Links

### DIFF
--- a/db/migrations/20220905120000_availability_status_not_null_constraint_default_value.go
+++ b/db/migrations/20220905120000_availability_status_not_null_constraint_default_value.go
@@ -1,0 +1,103 @@
+package migrations
+
+import (
+	_ "embed"
+	"fmt"
+
+	logging "github.com/RedHatInsights/sources-api-go/logger"
+	"github.com/go-gormigrate/gormigrate/v2"
+	"gorm.io/gorm"
+)
+
+// AvailabilityStatusColumnsNotNullConstraintDefaultValue adds a NOT NULL constraint to the "availability_status"
+// columns from all the tables that have it, and it also adds a default value "in_progress" for new insertions. The
+// migration takes care of the NULL or empty values from those columns and sets them to "in_progress".
+func AvailabilityStatusColumnsNotNullConstraintDefaultValue() *gormigrate.Migration {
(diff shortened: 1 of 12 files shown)


lpichler (approved):
Nice improvement, looks good!

Ellen-Yi-Dong (approved):
LGTM!

db/migrations/20220905120000_availability_status_not_null_constraint_default_value.go
```diff
@@ -0,0 +1,103 @@
...
+						Where(`availability_status IS NULL OR availability_status = ''`).
+						Update("availability_status", "in_progress").
+						Error
+
+					if err != nil {
+						return fmt.Errorf(`unable to set the NULL or empty availability statuses of table "%s" to "in_progress": %w`, table, err)
+					}
+
+					// Add a default value to the column.
+					defaultSql := fmt.Sprintf(`ALTER TABLE "%s" ALTER COLUMN "availability_status" SET DEFAULT 'in_progress'`, table)
```
lpichler, line 47: there is also set `gorm:"default:in_progress;not null`  at models - which it looks that duplicates this functionality for default setting. 

I would prefer to have it only at application level (in models) - because we can just forgot that we have it on db level - we would probably lose control of it from app side - (if we would need to change it we would need other migration)

or are you OK with this ?
@lindgrenj6 @MikelAlejoBR 

Thanks!

MikelAlejoBR: The reasoning for this was that if we, for any chance, miss validating an availability status, and we try to insert it as a NULL value, the database would simply default it to `in_progress`. In my case I don't have any problem having this duplicated, so that the issue doesn't happen again.

lindgrenj6 (approved):
Looks good, going to wait to merge until sending out a communication to at least the IQE folks since this could effect their test suites potentially (if they're checking for "" or unknown instead of in_progress on "bad" sources).

PR #502, opened 2022-08-09
[DISSCUSSION] feature: application user for the sources application

MikelAlejoBR:
The idea is to have two users in the sources application: a fully
privileged one, which will solely be used to perform database migrations
and the seeding.

The application user, on the other hand, is granted just the required
CRUD permissions on the tables for the application to work.

There are two goals:

* Restrict what the database user can do when operating normally.
* Pave the way to enable row level security via session variables in a
  future update.
  
  ## Note

I'm still waiting on AppSre's response according to what they think about having another user on the database. Also, this is just a draft to see what you think about this approach.

The idea would be to store the secrets on Vault, and to pull them from it to access the database. Although since these are database credentials, should they also go in a file on the pod's filesystem? :thinking:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,41 +18,43 @@ var parsedConfig *SourcesApiConfig
 
 // SourcesApiConfig is the struct for storing runtime configuration
 type SourcesApiConfig struct {
-	AppName                 string
-	Hostname                string
-	KafkaBrokerConfig       clowder.BrokerConfig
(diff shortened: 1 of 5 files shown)


MikelAlejoBR:
I'm closing this PR since it seems that we are not going to be able to do it with the restrictions that are in place... More information is available on https://issues.redhat.com/browse/RHCLOUD-19788

PR #503, opened 2022-08-10
feature: run SonarQube scanner containerized

MikelAlejoBR:
This change containerizes the keystore creation and runs the scanner on
a container too.

The pipeline's runner has a Java 1.8 as the Java runtime environment,
and even though the scanner has its own JRE embedded, it might
eventually complain if we feed it a keystore generated with an old JRE.

That's why the keystore creation is containerized. And since we're at
it, we also containerized running the scanner.

### DIFF
--- a/scripts/keystore/generate_keystore.bash
+++ b/scripts/keystore/generate_keystore.bash
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+#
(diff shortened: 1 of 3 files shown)


lpichler (approved):
Nice!

lindgrenj6 (approved):

PR #505, opened 2022-08-11
Memoize kafka SSL related variables to prevent memory from leaking

lindgrenj6:
This appears to be the cause of the memory leak when SSL is involved - I read through the docs that the Transport/Dialer can be used concurrently so we only need to create them once at runtime.


lpichler:
It looks that CI failure is related with this PR.

lindgrenj6:
I should have marked this as WIP I didn't get time to actually update the tests haha, they're updated now but the IQE tests are failing for weird reasons.

MikelAlejoBR (approved):

lpichler (approved):

PR #506, opened 2022-08-11
Remove dependency on sources' logger package for kafka

lindgrenj6:
So during managed kafka testing it was noticed that managed kafka likes to disconnect consumers around every 30 seconds which obviously results in an error. The problem here is that if another package is using this library (superkey-worker in this case) the entire program panics since the sources logger package isn't initialized (and why would it be, that isn't sources!).

This PR pulls out all deps from sources itself and might be a good candidate for upstreaming as part of the app-common-go starter app for kafka. 

---

While scouring the docs for more stuff I did notice that you can pass logging functions to the readers/writers so that internal kafka-ey things get logged like raw publich info, disconnects, rebalances etc which is really nice. So I included some helper functions and added them to the config struct. This way users can opt in/out of logging kafka internals as well as our log messages with the context of the listener.  

---
edit: also moved the kafka.CloseReader() call in the status listener to be before signaling back for shutdown, that way we for sure wait for the listener to shut down. 

edit again: ended up going through and implementing a wrapper around the reader/writer structs so we don't expose any implementation details.


lindgrenj6:
Ok I think this is ready - probably worth going through the commits. 

The first commit just moves the closeReader call to by synchronous rather than deferred so it can commit properly before exiting. The second commit basically pulls the sources logger requirement and introduces an `Options` struct so we can pass any logger through, the third commit wraps `kafka.Reader/Writer` so they don't expose implementation details to any end-users, it also changes the way the logger gets passed so users don't have to pass individual functions they can just pass any logger that implements `Debugf` and `Errorf`.

MikelAlejoBR (approved):
Nice! LGTM!

lpichler:
There is conflict :/

lindgrenj6:
Fixed, and tests are green now.

lpichler (approved):
very interesting! LGTM!

PR #513, opened 2022-08-16
[RHCLOUD-20615] fix: return status 400 on user error on Bulk Create

MikelAlejoBR:
Until now most of the errors were returned as 500s, even if it was an
error related with an invalid input from the user. This commit fixes
that.

## Links

[[RHCLOUD-20615]](https://issues.redhat.com/browse/RHCLOUD-20615)

### DIFF
--- a/bulk_handlers_test.go
+++ b/bulk_handlers_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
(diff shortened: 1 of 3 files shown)


lindgrenj6 (approved):
Thanks for this, wanted to do this but never got time.

lpichler (approved):
LGTM!

PR #515, opened 2022-08-16
[RHCLOUD-18655] feature: translate EBS account numbers to "orgIds"

MikelAlejoBR:
After checking for untranslated EBS accout numbers, I found that there
are quite a few that are not translated. However, I was able to manually
translate them with the translation tool, so I am suspecting that we
might get rid of quite a bit of untranslated tenants if we perform a
second run of this migration.

After this, I'd like to check for untranslated EBS account numbers again
to discuss a proper move for these.

## Links

[[RHCLOUD-18655]](https://issues.redhat.com/browse/RHCLOUD-18655)

### DIFF
--- a/db/migrations/20220811170000_translate_ebs_org_ids_second_run.go
+++ b/db/migrations/20220811170000_translate_ebs_org_ids_second_run.go
@@ -1,0 +1,87 @@
+package migrations
+
+import (
+	"context"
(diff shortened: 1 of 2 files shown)


MikelAlejoBR:
I'm closing this one since we're taking a different approach. Check the Jira ticket for more details.

PR #516, opened 2022-08-16
Switch to x-rh-insights-request-id instead of generating our own uuid

lindgrenj6:
I totally forgot that 3scale includes a request-id on each request. This way we can use that and follow it through the system more effectively.


MikelAlejoBR (approved):
This is getting better and better!

lpichler (approved):

PR #519, opened 2022-08-17
[RHCLOUD-18655] feature: add tenant translation endpoints

MikelAlejoBR:
We were initially going to perform another migration to translate the
leftover tenants which don't have an associated "OrgId" value, but after
a stand up meeting we decided upon creating a couple of endpoints to be
able to do it on demand.

These endpoints will help at:

1. Fetching all the "translatable" tenants.
2. Initiating the "EBS account number to OrgId" process.

Of course, these endpoints go on the internal routes, so they will only
be able to be hit from inside the cluster.

## Links

[[RHCLOUD-18655]](https://issues.redhat.com/browse/RHCLOUD-18655)

### DIFF
--- a/dao/tenant_dao.go
+++ b/dao/tenant_dao.go
@@ -113,7 +113,7 @@ func (t *tenantDaoImpl) GetUntranslatedTenants() ([]m.Tenant, error) {
 
 	err := DB.Debug().
 		Model(&m.Tenant{}).
-		Where(`("external_tenant" IS NOT NULL OR "external_tenant" = '') AND ("org_id" IS NULL OR "org_id" = '')`).
+		Where(`("external_tenant" IS NOT NULL OR "external_tenant" != '') AND ("org_id" IS NULL OR "org_id" = '')`).
 		Find(&tenants).
 		Error
 
@@ -126,7 +126,7 @@ func (t *tenantDaoImpl) GetUntranslatedTenants() ([]m.Tenant, error) {
 
 func (t *tenantDaoImpl) TranslateTenants() (int64, uint64, uint64, []m.TenantTranslation, error) {
 	// The where condition to fetch only the untranslated tenants.
-	untranslatedTenantsWhereCondition := `("external_tenant" IS NOT NULL OR "external_tenant" = '') AND ("org_id" IS NULL OR "org_id" = '')`
+	untranslatedTenantsWhereCondition := `("external_tenant" IS NOT NULL OR "external_tenant" != '') AND ("org_id" IS NULL OR "org_id" = '')`
 
 	// Count the total number of translatable tenants.
 	var translatableTenants int64
--- a/dao/tenant_dao.go
+++ b/dao/tenant_dao.go
@@ -212,8 +212,8 @@ func (t *tenantDaoImpl) TranslateTenants() (int64, uint64, uint64, []m.TenantTra
 					continue
 				}
 
-				if err != nil {
-					translationError := fmt.Errorf(`could no translate to "org_id": %w`, err)
+				if dbResult.Error != nil {
+					translationError := fmt.Errorf(`could no translate to "org_id": %w`, dbResult.Error)
 
 					logger.Log.WithField("external_tenant", *result.EAN).WithField("org_id", result.OrgID).Error(translationError)
 
--- a/dao/tenant_dao.go
+++ b/dao/tenant_dao.go
@@ -225,7 +225,9 @@ func (t *tenantDaoImpl) TranslateTenants() (int64, uint64, uint64, []m.TenantTra
 					continue
 				}
 
-				logger.Log.WithField("external_tenant", *result.EAN).WithField("org_id", result.OrgID).Info("successful tenant translation")
+				logger.Log.WithFields(
+					logrus.Fields{"external_tenant": *result.EAN, "org_id": result.OrgID, "tenant_id": tenant.Id},
+				).Info("successful tenant translation")
 
 				tenantTranslation.Id = tenant.Id
 				tenantTranslation.ExternalTenant = *result.EAN
--- a/dao/tenant_dao.go
+++ b/dao/tenant_dao.go
@@ -199,6 +199,19 @@ func (t *tenantDaoImpl) TranslateTenants() (int64, uint64, uint64, []m.TenantTra
 						"org_id": result.OrgID,
 					})
 
+				if dbResult.Error != nil {
+					translationError := fmt.Errorf(`could no translate to "org_id": %w`, dbResult.Error)
+
+					logger.Log.WithField("external_tenant", *result.EAN).WithField("org_id", result.OrgID).Error(translationError)
+
+					tenantTranslation.ExternalTenant = *result.EAN
+					tenantTranslation.OrgId = result.OrgID
+					tenantTranslations = append(tenantTranslations, tenantTranslation)
+
+					untranslatedTenants++
+					continue
+				}
+
 				// If the update yielded 0 affected rows that means that no tenant could be found with that
 				// "external_tenant".
 				if dbResult.RowsAffected == 0 {
@@ -208,19 +221,6 @@ func (t *tenantDaoImpl) TranslateTenants() (int64, uint64, uint64, []m.TenantTra
 
 					tenantTranslation.ExternalTenant = *result.EAN
 					tenantTranslation.Error = translationError
-					tenantTranslations = append(tenantTranslations, tenantTranslation)
-
-					untranslatedTenants++
-					continue
-				}
-
-				if dbResult.Error != nil {
-					translationError := fmt.Errorf(`could no translate to "org_id": %w`, dbResult.Error)
-
-					logger.Log.WithField("external_tenant", *result.EAN).WithField("org_id", result.OrgID).Error(translationError)
-
-					tenantTranslation.ExternalTenant = *result.EAN
-					tenantTranslation.OrgId = result.OrgID
 					tenantTranslations = append(tenantTranslations, tenantTranslation)
 
 					untranslatedTenants++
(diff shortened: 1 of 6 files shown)


dao/tenant_dao.go
```diff
@@ -68,3 +77,141 @@ func (t *tenantDaoImpl) TenantByIdentity(id *identity.Identity) (*m.Tenant, erro
...
+					tenantTranslations = append(tenantTranslations, tenantTranslation)
+
+					untranslatedTenants++
+					continue
+				}
+
+				// Grab the updated tenant for the "translated tenant" struct.
+				var tenant m.Tenant
+				// Update the tenant with the translated "orgId".
+				dbResult := DB.Debug().
```
lpichler: I did not play with transaction in gorm but should not be here `tx` instead of `DB.Debug()` ?

MikelAlejoBR: Ooof you're right, yes. I wrapped everything on a transaction and then I forgo to use it :joy: Thanks!

dao/tenant_dao.go
```diff
@@ -70,3 +77,141 @@ func (t *tenantDaoImpl) TenantByIdentity(id *identity.Identity) (*m.Tenant, erro
...
+					Model(&tenant).
+					Clauses(clause.Returning{Columns: []clause.Column{{Name: "id"}}}).
+					Where("external_tenant = ?", result.EAN).
+					Updates(map[string]interface{}{
+						"org_id": result.OrgID,
+					})
+
+				// If the update yielded 0 affected rows that means that no tenant could be found with that
+				// "external_tenant".
+				if dbResult.RowsAffected == 0 {
```
lpichler: - EAN has to exist in sources DB because of condition in `untranslatedTenantsWhereCondition`.
- IF EAN exists in sources DB but doesn't exist in translation service I guess - it is caught on line 148.
- `org_id` is not populated yes in tenant DB for the EAN

what has to happen to have this condition true ?

but probably it doesn't hurt to have it here.

dao/tenant_dao.go
```diff
@@ -70,3 +77,141 @@ func (t *tenantDaoImpl) TenantByIdentity(id *identity.Identity) (*m.Tenant, erro
...
+
+					tenantTranslation.ExternalTenant = *result.EAN
+					tenantTranslation.Error = translationError
+					tenantTranslations = append(tenantTranslations, tenantTranslation)
+
+					untranslatedTenants++
+					continue
+				}
+
+				if err != nil {
```
lpichler: it looks that this `err != nil` is covered on 134 or am I missing something ?

dao/tenant_dao.go
```diff
@@ -70,3 +77,141 @@ func (t *tenantDaoImpl) TenantByIdentity(id *identity.Identity) (*m.Tenant, erro
...
+					tenantTranslations = append(tenantTranslations, tenantTranslation)
+
+					untranslatedTenants++
+					continue
+				}
+
+				// Grab the updated tenant for the "translated tenant" struct.
+				var tenant m.Tenant
+				// Update the tenant with the translated "orgId".
+				dbResult := tx.
```
lpichler, line 194: Is in postgres batch update ? I think that yes - but maybe it is not needed here - I guess we don't have too much untranslated EAN.

dao/tenant_dao.go
```diff
@@ -70,3 +77,141 @@ func (t *tenantDaoImpl) TenantByIdentity(id *identity.Identity) (*m.Tenant, erro
...
+		translator := tenantid.NewTranslator(config.Get().TenantTranslatorUrl)
+		for i := int64(0); i < translatableTenants; i += 100 {
+			// Grab all the EBS account numbers to be processed.
+			var ebsAccountNumbers []string
+			err := tx.
+				Model(&m.Tenant{}).
+				Where(untranslatedTenantsWhereCondition).
+				Pluck("external_tenant", &ebsAccountNumbers).
+				Offset(int(i)).
+				Limit(100).
```
lpichler, line 163: I guess we need to add here Order BY clause as order of records is not ensured.

lpichler: Oh but maybe it will not causes any issues as there is `Where(untranslatedTenantsWhereCondition)`, I am taking it back.

dao/tenant_dao.go
```diff
@@ -70,3 +77,141 @@ func (t *tenantDaoImpl) TenantByIdentity(id *identity.Identity) (*m.Tenant, erro
...
+	if err != nil {
+		return nil, err
+	}
+
+	return tenants, err
+}
+
+func (t *tenantDaoImpl) TranslateTenants() (int64, uint64, uint64, []m.TenantTranslation, error) {
+	// The where condition to fetch only the untranslated tenants.
+	untranslatedTenantsWhereCondition := `("external_tenant" IS NOT NULL OR "external_tenant" != '') AND ("org_id" IS NULL OR "org_id" = '')`
```
lpichler: is it possible to share this condition as with `GetUntranslatedTenants` ?

lpichler:
Who will call this ? Will it be in cronjob ?

lpichler:
OT: It looks that CI failure is sporadic and unrelated - It is about improper matching same arrays.

dao/tenant_dao.go
MikelAlejoBR: Since we are using a transaction that code should be "dead code", you're right. Do I leave it just in case? :thinking:

dao/tenant_dao.go
MikelAlejoBR: You're right. Actually this was meant to be `dbResult.Error`, to be able to log a potential error with the update statement.

dao/tenant_dao.go
MikelAlejoBR: Mmmmmmmm... I don't follow. Why do we want ordered records here? :thinking: I included the clause anyways :+1:

dao/tenant_dao.go
MikelAlejoBR: Sure, I created a `const` with that string.

dao/tenant_dao.go
MikelAlejoBR: > Who will call this ? Will it be in cronjob ?

More like CronJacob, since he'll be the one sending these requests for the translations

MikelAlejoBR:
> Who will call this ? Will it be in cronjob ?

More like CronJacob, since he'll be the one sending the requests internally to run the translations

dao/tenant_dao.go
lpichler: Idea is when order of records is not ensured then it can take first 100 records(for offset=0) and then it can pick also some same records for next 100 records - 

is it probably edge case and most likely it will not cause issue - tenants from batch are processed and org_id is set and then those tenants are excluded from this query thanks to the  `untranslatedTenantsWhereCondition`
- so it pertains just records which remain after translation processing without `org_id`.

hopefully it makes sense.

Thanks for adding order by!

dao/tenant_dao.go
lpichler: yes, I agree 👍

dao/tenant_dao.go
MikelAlejoBR: Thanks an interesting point, yeah. Thanks for the explanation! :+1:

dao/tenant_dao.go
MikelAlejoBR: I thought Gorm supported batch updating, but I am facing a really weird "unsupported data" when fetching the tenants and then calling `db.Updates(&tenants)`. I've looked on the internet to actually see if you can do a batch update when you're trying to update multiple rows with different values and... things are a bit complicated.

However, aside from this, the implementation I came up with was a little bit weird:

1. I fetched 100 tenants per loop.
2. Created a `map[string]*m.Tenant` to store in memory the tenants by their external tenant.
3. Called the translation service.
4. I used the `map[string]` to update each of the tenants with their corresponding `org_id`.
5. Updated the database with the tenants.
6. Looped through the tenants to generate a proper HTTP response.

Here's the implementation: https://gist.github.com/MikelAlejoBR/54c6f3d3694556cc4344f6f8c3e03f88

I would honestly not refactor the current PR's code to the above, to be honest... plus I would have to check if there's a way to properly batch update with Gorm without hitting an error :eyes:

lpichler: yes, you are right and thank you very much for trying it and for those information! 👏

lpichler (approved):
Looks great! 👍

dao/tenant_dao.go
```diff
@@ -103,3 +111,141 @@ func (t *tenantDaoImpl) TenantByIdentity(id *identity.Identity) (*m.Tenant, erro
...
+				Order(clause.OrderByColumn{Column: clause.Column{Name: "external_tenant"}, Desc: false}).
+				Limit(100).
+				Error
+
+			if err != nil {
+				return fmt.Errorf("unable to fetch the tenants to translate them: %w", err)
+			}
+
+			// Batch process 100 EBS account numbers.
+			results, err := translator.EANsToOrgIDs(context.Background(), ebsAccountNumbers)
```
lindgrenj6, line 171: This will actually send all of the `ebsAccountNumbers` at once, the name of the method is misleading.

We should probably loop over this slice 100 at a time and translate them that way.

MikelAlejoBR: I believe I am doing that in the code, right? Everything is wrapped on a `for i := int64(0); i < translatableTenants; i += 100` loop. Were you actually referring to fetching everything at once and then applying the loop to just the translator part?

lindgrenj6: Yep I see it now. I totally missed it on L154, sorry about that.

lindgrenj6 (approved):

dao/tenant_dao.go
MikelAlejoBR: No problem at all! :)

lpichler:
CronJ(ac)ob can be activated.

PR #521, opened 2022-08-18
[RHCLOUD-20860] Refactor set OrgIds and EBS account numbers from the database

MikelAlejoBR:
Instead of just forwarding the incoming headers, this change ensures
that the "EBS account number" and the "OrgId" are set in the context
using the contents from the database.

This comes from one issue we have been having with the notifications
service: the service was erroring because we were just sending the
incoming "EBS account number" header, even though we had the associated
"OrgId" on the database.

So we have decided to simply use the incoming headers as a validation
—make sure that an EBS account number and/or OrgId are present—, and
then use the database contents to send the headers to the rest of the
services or to the Kafka queues.

## Links

[[RHCLOUD-20860]](https://issues.redhat.com/browse/RHCLOUD-20860)

### DIFF
--- a/dao/interfaces.go
+++ b/dao/interfaces.go
@@ -166,9 +166,9 @@ type RhcConnectionDao interface {
 }
 
 type TenantDao interface {
-	// GetOrCreateTenantID returns the ID of the tenant associated with the provided identity. It tries to fetch the
-	// tenant by its OrgId, and if it is not present, by its EBS account number.
-	GetOrCreateTenantID(identity *identity.Identity) (int64, error)
+	// GetOrCreateTenant returns the tenant associated with the provided identity. It tries to fetch the tenant by its
+	// OrgId, and if it is not present, by its EBS account number.
+	GetOrCreateTenant(identity *identity.Identity) (*m.Tenant, error)
 	// TenantByIdentity returns the tenant associated to the given identity. It tries to fetch the tenant by its OrgId,
 	// and if it is not preset, by its EBS account number.
 	TenantByIdentity(identity *identity.Identity) (*m.Tenant, error)
--- a/dao/tenant_dao.go
+++ b/dao/tenant_dao.go
@@ -1,6 +1,8 @@
 package dao
 
 import (
+	"fmt"
+
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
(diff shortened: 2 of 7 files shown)


lpichler:
This is great, thanks for doing it!

2 things: 

1. in notifier middleware I am passing` XRHID identity` to `EmitAvailabilityStatusNotification` is it also possible to populate `OrgID` in context's `h.PARSED_IDENTITY (*identity.XRHID)` struct instance ? Other way is that we can do is that we can populate OrgID in  `identity.XRHID` struct instance which passed to `EmitAvailabilityStatusNotification` but my preference is to do in tenancy middleware. Let me know what do you think. 
2. This is not passing OrgID(in case we have api request without OrgID but orgID is in DB ) to `LoggerFields` because tenancy middleware is not before `ParseHeaders` in `e.Group` in routes.go. Maybe we can put Tenancy middleware before  `LoggerFields` ? cc @lindgrenj6

lindgrenj6:
Yeah we could throw `LoggerFields` right in the `tenancyMiddleware` group at the top - and add it where needed throughout the app. I think the only place that isn't covered by those is check avialability and get /endpoints/:id from my quick grep.

lindgrenj6 (approved):
Barring the loggerfields part looks good!

MikelAlejoBR:
>     1. in notifier middleware I am passing` XRHID identity` to `EmitAvailabilityStatusNotification` is it also possible to populate `OrgID` in context's `h.PARSED_IDENTITY (*identity.XRHID)` struct instance ? Other way is that we can do is that we can populate OrgID in  `identity.XRHID` struct instance which passed to `EmitAvailabilityStatusNotification` but my preference is to do in tenancy middleware. Let me know what do you think.

You are right, yeah. We talked about it and I forgot implementing it... I've just pushed that!

>     2. This is not passing OrgID(in case we have api request without OrgID but orgID is in DB ) to `LoggerFields` because tenancy middleware is not before `ParseHeaders` in `e.Group` in routes.go. Maybe we can put Tenancy middleware before  `LoggerFields` ? cc @lindgrenj6
>> Yeah we could throw `LoggerFields` right in the `tenancyMiddleware` group at the top - and add it where needed throughout the app. I think the only place that isn't covered by those is check avialability and get /endpoints/:id from my quick grep.

All right, I've rearranged the middlewares!

lpichler (approved):
LGTM This is works great! Thanks again!

PR #522, opened 2022-08-18
[RHCLOUD-20865] Refactor tenancy lookup orgid preference

MikelAlejoBR:
Since the EBS account numbers are deprecated, we want to explicitly find
the tenants by their OrgId first. We are aware that this solution
hurts the performance just a bit, but this will eventually go away.

## Links 

[[RHCLOUD-20865]](https://issues.redhat.com/browse/RHCLOUD-20865)

### DIFF
--- a/dao/tenant_dao.go
+++ b/dao/tenant_dao.go
@@ -4,9 +4,11 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/RedHatInsights/sources-api-go/logger"
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
+	"github.com/sirupsen/logrus"
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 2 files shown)


lindgrenj6 (commented):
I know this is built on #521 - but still looking :eyes: 

I think we should log when we have to fallback to the EBS acct number - that way we can at least hopefully get an idea around how many misses we're getting via kibana.

Other than that looks great!

MikelAlejoBR:
> I know this is built on #521 - but still looking eyes
> 
> I think we should log when we have to fallback to the EBS acct number - that way we can at least hopefully get an idea around how many misses we're getting via kibana.
> 
> Other than that looks great!

Yeah, I was planning on not requesting a review until the other one got merged. However I've implemented your suggestion on this one! Thanks!

lpichler:
https://github.com/RedHatInsights/sources-api-go/pull/521 is merged 👀

MikelAlejoBR:
> #521 is merged eyes

Great! I just rebased this PR!

lpichler:
(I removed wait-to-merge label)

lindgrenj6 (approved):
Looks good. I noticed we don't log not-found requests - which is probably fine but I was puzzled for a minute why I was only seeing the successful queries haha.

MikelAlejoBR:
Mmmm... I've included an `info` log for when the tenants get created. That should help us identify when the tenant was not found by its org id or ebs account numbers.

lpichler (approved):
LGTM

PR #523, opened 2022-08-18
[RHCLOUD-20691] Check err handling from Bind() in handlers

Ellen-Yi-Dong:
Within authentication_handlers.go, instead of util.NewErrBadRequest(err), change to err and check err handling from Bind() in handlers.

Within authentication_handlers_test.go, set binder on each invocation

JIRA: RH-CLOUD 20691[https://issues.redhat.com/browse/RHCLOUD-20691](url)

### DIFF
--- a/authentication_handlers.go
+++ b/authentication_handlers.go
@@ -150,13 +150,13 @@ func AuthenticationEdit(c echo.Context) error {
 		return err
 	}
 
-	updateRequest := &m.AuthenticationEditRequest{}
-	err = c.Bind(updateRequest)
+	updateRequest := m.AuthenticationEditRequest{}
+	err = c.Bind(&updateRequest)
 	if err != nil {
-		return util.NewErrBadRequest(err)
+		return err
 	}
 
-	err = service.ValidateAuthenticationEditRequest(updateRequest)
+	err = service.ValidateAuthenticationEditRequest(&updateRequest)
 	if err != nil {
 		return util.NewErrBadRequest(err)
 	}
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (approved):
Nice one! Looks good to me!

authentication_handlers_test.go
```diff
@@ -417,6 +420,8 @@ func TestAuthenticationCreateBadRequestInvalidResourceType(t *testing.T) {
 	)
 	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
 
+	c.Echo().Binder = echoBinder
+
```
petracihalova, line 424: well I am not sure about this ... yes we need it here but I think we could set Binder() for all our tests on one place ... I started discussion on Slack ... maybe I am wrong but this can probably lead to separate PR where we set Binder() for all our tests (so in this PR we can leave it the way you did it) 🤔 😇

authentication_handlers.go
```diff
@@ -150,13 +150,13 @@ func AuthenticationEdit(c echo.Context) error {
...
 	}
 
-	updateRequest := &m.AuthenticationEditRequest{}
-	err = c.Bind(updateRequest)
+	updateRequest := m.AuthenticationEditRequest{}
+	err = c.Bind(&updateRequest)
 	if err != nil {
-		return util.NewErrBadRequest(err)
+		return err
 	}
```
petracihalova, line 157: we need this change in other handlers too (but can be part of another PR for sure 😇)

PR #524, opened 2022-08-22
[RHCLOUD-20767] feature: return 400 response on empty or null JSON bodies

MikelAlejoBR:
This will let know the clients that basically nothing changed on our
side due to receiving a bad payload.

## Links

[[RHCLOUD-20767]](https://issues.redhat.com/browse/RHCLOUD-20767)

### DIFF
--- a/binder.go
+++ b/binder.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"encoding/json"
-	"net/http"
+	"reflect"
 
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
@@ -19,11 +19,12 @@ type NoUnknownFieldsBinder struct{}
 	`DisallowUnknownFields` Decoder setting for unmarshaling json.
 */
 func (binder *NoUnknownFieldsBinder) Bind(i interface{}, c echo.Context) error {
-	// if there is no body on the request - return early
-	if c.Request().Body == http.NoBody || c.Request().Body == nil {
-		return util.NewErrBadRequest("no body")
-	}
+	// Close the request's body after we're done with it.
 	defer c.Request().Body.Close()
+
+	// is the struct passed in initially empty? then it's pretty easy to tell if
+	// no fields were plucked from the body.
+	isInitiallyEmpty := reflect.ValueOf(i).Elem().IsZero()
 
 	// create a new decoder for the request body
 	dec := json.NewDecoder(c.Request().Body)
@@ -34,6 +35,14 @@ func (binder *NoUnknownFieldsBinder) Bind(i interface{}, c echo.Context) error {
 	if err != nil {
 		c.Logger().Warnf("Failed to decode request: %v", err)
 		return util.NewErrBadRequest(err)
+	}
+
+	// ...is the struct still empty after unmarshaling? if so then we had an
+	// empty JSON body.
+	isEmptyAfterDecoding := reflect.ValueOf(i).Elem().IsZero()
+
+	if isInitiallyEmpty && isEmptyAfterDecoding {
+		return util.NewErrBadRequest("empty body")
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


petracihalova (commented):
it works well 👍🏻 I was testing scenario for request without body and I got in response
```
{
    "errors": [
        {
            "detail": "bad request: bad request: EOF",
            "status": "400"
        }
    ]
}
```

and before we got better description about what is wrong
```
{
    "errors": [
        {
            "detail": "bad request: bad request: no body",
            "status": "400"
        }
    ]
}
```

I think we still need to cover scenarios for requests without body 🤔 😇 ❤️
(it is not needed to solve now this double "bad request" in the message because this will be solved here [RHCLOUD-20691](https://issues.redhat.com/browse/RHCLOUD-20691))

petracihalova (approved):
nice 👍🏻

lindgrenj6:
Now it's time to throw a wrench in the conversation! I agree with what Mikel said a while ago about reading the body twice - I see we're basically reading the reader into another buffer so basically re-allocating the entire body. I'm not a huge fan but that MIGHT be what we need to do.

Unless we dip into reflection!

As long as we're diligent about always passing in zero-value structs (which we are) we can use `reflect.ValueOf(i).Elem().IsZero()` to tell whether the value is zero or not. From there we can check the interface before/after marshaling, if it's zero both times then no fields were there and it was an empty body. 

I checked a few of the cases and it seemed to work:
- `null` -> empty body error
- `{}` -> empty body error
- `[]` -> unmarshal error on individual structs

What do you all think? 

```diff
diff --git a/binder.go b/binder.go
index e4b186a..6bfe522 100644
--- a/binder.go
+++ b/binder.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"net/http"
+	"reflect"
 
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
@@ -25,6 +26,10 @@ func (binder *NoUnknownFieldsBinder) Bind(i interface{}, c echo.Context) error {
 	}
 	defer c.Request().Body.Close()
 
+	// is the struct passed in initially empty? then it's pretty easy to tell if
+	// no fields were plucked from the body.
+	initiallyEmpty := reflect.ValueOf(i).Elem().IsZero()
+
 	// create a new decoder for the request body
 	dec := json.NewDecoder(c.Request().Body)
 	// return an error if any non-declared fields are included in the request
@@ -36,5 +41,12 @@ func (binder *NoUnknownFieldsBinder) Bind(i interface{}, c echo.Context) error {
 		return util.NewErrBadRequest(err)
 	}
 
+	// ...is the struct still empty after unmarshaling? if so then we had an
+	// empty JSON body.
+	emptyStruct := reflect.ValueOf(i).Elem().IsZero()
+	if emptyStruct && initiallyEmpty {
+		return util.NewErrBadRequest("empty body")
+	}
+
 	return nil
 }
```

MikelAlejoBR:
To be honest I like the simplicity of your proposal, and I don't mind using reflection for this one. When I was writing this PR I hoped that the `io.Reader` would have a `.Peek` function like the `io.ReaderClose` has, to take a look at the contents without copying them around, but unfortunately there is not. I could have implemented it myself since it isn't that difficult, but the thing was starting to feel like "a little bit too much"...

binder.go
```diff
@@ -19,12 +19,13 @@ type NoUnknownFieldsBinder struct{}
 	`DisallowUnknownFields` Decoder setting for unmarshaling json.
 */
 func (binder *NoUnknownFieldsBinder) Bind(i interface{}, c echo.Context) error {
-	// if there is no body on the request - return early
-	if c.Request().Body == http.NoBody || c.Request().Body == nil {
-		return util.NewErrBadRequest("no body")
-	}
```
lindgrenj6, line 25: Removing this check makes me nervous but I couldn't get it to trigger - so it's probably fine!

lindgrenj6 (approved):
LGTM!

lpichler (approved):

PR #526, opened 2022-08-23
Remove endpoint object from amazon section for provisioning

lindgrenj6:
Not sure why this was there in the first place - it looks like it was part of the cloud-meter-arn object above but obviously we haven't been sending it for a while so maybe it just isn't used anymore.

It doesn't make sense to begin with since we don't need an endpoint for an application based source. So I'm thinking it's legacy code. 

...lets see if the checks all pass.

### DIFF
--- a/dao/seeds/source_types.yml
+++ b/dao/seeds/source_types.yml
@@ -84,14 +84,6 @@ amazon:
               pattern: "^arnaws:.*"
             - type: min-length
               threshold: 10
-    endpoint:
-      hidden: true
-      fields:
-      - component: text-field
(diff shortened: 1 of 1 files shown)


lpichler (approved):

MikelAlejoBR (approved):
Probably I copy pasted this :man_facepalming:

lindgrenj6:
Nope it was me! I pulled it directly from the ruby code. It's just kinda been chilling there forever.

ezr-ondrej (approved):
Fixed my issue! :+1:

PR #527, opened 2022-08-23
Source dao test TestSourceSubcollectionWithUserOwnership() fix

petracihalova:
without JIRA ticket

sometimes when I run the tests I get this result
```
--- FAIL: TestSourceSubcollectionWithUserOwnership (2.16s)
    source_dao_test.go:1082: Expected source IDS [103 105] are not same with obtained ids: [105 103]
```

I found out that subcollection source IDs can be returned in random order so they have to be sorted before comparing them with the expected result

EDIT:
updated other tests that contains same logic

### DIFF
--- a/dao/application_authentication_dao_test.go
+++ b/dao/application_authentication_dao_test.go
@@ -3,6 +3,7 @@ package dao
 import (
 	"errors"
 	"reflect"
+	"sort"
 	"testing"
 
 	"github.com/RedHatInsights/sources-api-go/internal/testutils"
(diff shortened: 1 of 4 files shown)


lindgrenj6 (approved):
...I can't wait until we have generic sort functions after go 1.18 so `sort.Ints` would work on these int64s.

MikelAlejoBR (approved):

petracihalova:
added same fix for other tests (that contains same logic)

MikelAlejoBR (approved):

PR #532, opened 2022-08-24
[RHCLOUD-20947] middleware function update for rhc connection source list endpoint

petracihalova:
I found out that for existing rhc connection the endpoint `/rhc_connections/:id/sources` returns not found
because we used for this route wrong middleware (tenant was not set from header and that led to not found err)

**JIRA:** [RHCLOUD-20947](https://issues.redhat.com/browse/RHCLOUD-20947)


MikelAlejoBR (approved):
Nice! LGTM!

lpichler (approved):

lindgrenj6 (approved):

PR #535, opened 2022-08-25
[RHCLOUD-20908] Endpoint create - source id validation (tenancy)

petracihalova:
in endpoint create handler we need to check if provided source id belongs to the tenant

I added the check outside `service.ValidateEndpointCreateRequest()` because we are using this function in bulk create where all resources are created in a transaction ... so the source doesnt exist when the bulk create calls this validation for endpoint

**JIRA:** [RHCLOUD-20908](https://issues.redhat.com/browse/RHCLOUD-20908)

### DIFF
--- a/endpoint_handlers_test.go
+++ b/endpoint_handlers_test.go
@@ -689,6 +689,42 @@ func TestEndpointCreateBadRequest(t *testing.T) {
 	templates.BadRequestTest(t, rec)
 }
 
+// TestEndpointCreateInvalidSource tests bad request is returned
+// when create request contains source id that is not owned by tenant
+func TestEndpointCreateInvalidSource(t *testing.T) {
+	testutils.SkipIfNotRunningIntegrationTests(t)
+	tenantId := int64(2)
+	sourceId := int64(1)
+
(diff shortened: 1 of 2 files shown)


lindgrenj6 (approved):

endpoint_handlers_test.go
```diff
@@ -689,6 +689,42 @@ func TestEndpointCreateBadRequest(t *testing.T) {
 	templates.BadRequestTest(t, rec)
 }
 
+// TestEndpointCreateInvalidTenant tests bad request is returned
+// when create request contains source id that is not owned by tenant
+func TestEndpointCreateInvalidTenant(t *testing.T) {
```
MikelAlejoBR: I guess the name should be `TestEndpointCreateInvalidSource` then, right?

MikelAlejoBR (approved):
LGTM!

endpoint_handlers_test.go
petracihalova: 🤔 you're right, that is better name for it 😁

lpichler (approved):

PR #536, opened 2022-08-25
Enable record creation with user ownership for AppStudio application type

lpichler:
1. Remove UserOwnership var from config
2. Replace userResourceOwnership with userIDPresent in UserResource
3. Set user ownership for app-studio in application_type.yml
---
Point 1. and 2. I removed possibility to enable user based feature with config - It looks to me now superfluous.
Point 3. This set `ApplicationType#UserOwnership` to `user` for AppStudio - which enable to populate user_id records during bulk creation for records for Sources tied with AppStudio application type 

### Links
- https://github.com/RedHatInsights/sources-api-go/issues/356

### DIFF
--- a/bulk_handlers_test.go
+++ b/bulk_handlers_test.go
@@ -64,8 +64,6 @@ func TestBulkCreateMissingSourceType(t *testing.T) {
 func TestBulkCreateWithUserCreation(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
 
-	conf.ResourceOwnership = "user"
-
 	testUserId := "testUser"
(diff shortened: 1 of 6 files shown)


MikelAlejoBR (approved):
LGTM!

lindgrenj6 (commented):
We probably need to pull the flag out of the `deploy/clowdapp.yml` file still - other than that looks good.

petracihalova (approved):
LGTM ⭐️

lpichler:
> We probably need to pull the flag out of the `deploy/clowdapp.yml` file still - other than that looks good.

yes, that's right, thanks for noticing it! I removed it.

lindgrenj6 (approved):

PR #538, opened 2022-08-26
[RHCLOUD-20978] App create validation for source id (tenancy)

petracihalova:
when we create new application we have to check if given source id belongs to the tenant

this PR adds check for the source id in create request, fixes the related tests and adds new test for app create handler

**JIRA:** [RHCLOUD-20978](https://issues.redhat.com/browse/RHCLOUD-20978)

### DIFF
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -89,12 +89,17 @@ func ApplicationCreate(c echo.Context) error {
 		return err
 	}
 
+	requestParams, err := dao.NewRequestParamsFromContext(c)
+	if err != nil {
+		return err
+	}
+
 	input := &m.ApplicationCreateRequest{}
 	if err = c.Bind(input); err != nil {
 		return err
 	}
 
-	err = service.ValidateApplicationCreateRequest(input)
+	err = service.ValidateApplicationCreateRequest(requestParams, input)
 	if err != nil {
 		return util.NewErrBadRequest(fmt.Sprintf("Validation failed: %v", err))
 	}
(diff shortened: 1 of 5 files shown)


MikelAlejoBR (approved):
LGTM!

petracihalova:
rebased

application_handlers.go
```diff
@@ -89,12 +89,17 @@ func ApplicationCreate(c echo.Context) error {
...
+		return err
+	}
+
 	input := &m.ApplicationCreateRequest{}
 	if err = c.Bind(input); err != nil {
 		return err
 	}
 
-	err = service.ValidateApplicationCreateRequest(input)
+	err = service.ValidateApplicationCreateRequest(sourceDao, input)
```
lindgrenj6: Rather than pass a DAO - Should we just pass the tenant ID or request params? That way in the sub-function we can get whatever DAO's we would like.

lindgrenj6 (commented):
Just one thought - otherwise looks good.

application_handlers.go
petracihalova: ok, changed

petracihalova:
as suggested, now the validation function takes as input request parameters (tenant, user)

petracihalova:
rebased & conflict solved

lpichler (approved):

MikelAlejoBR (approved):

lindgrenj6 (approved):

PR #544, opened 2022-08-30
[RHCLOUD-20931] [GO] Edit source name to be set nil not empty

Ellen-Yi-Dong:
Within source_validation.go, edit ValidateEditSourceNameRequest() to allow req.Name to be null.

Within source_handlers_test.go, created tests TestSourceEditNameEmptyString and TestSourceEditNoNameRequest in order to test that a bad request is returned if the changing name to an empty string and that the request returns a 200 without name in body of request respectively

LINK: [RHCLOUD-20931](https://issues.redhat.com/browse/RHCLOUD-20931)

### DIFF
--- a/source_handlers_test.go
+++ b/source_handlers_test.go
@@ -1124,8 +1124,8 @@ func TestSourceCreate(t *testing.T) {
 	}
 
 	id, _ := strconv.ParseInt(src.ID, 10, 64)
-	dao, _ := getSourceDao(c)
-	_, err = dao.Delete(&id)
+	SourceDao, _ := getSourceDao(c)
+	_, err = SourceDao.Delete(&id)
 	if err != nil {
 		t.Errorf("Failed to delete source id %v", id)
 	}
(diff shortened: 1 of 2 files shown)


lindgrenj6 (approved):
Nice job! :clap:

source_handlers_test.go
```diff
@@ -1305,6 +1305,110 @@ func TestSourceEditBadRequest(t *testing.T) {
...
+		"/api/sources/v3.1/sources/xxx",
+		bytes.NewReader(body),
+		map[string]interface{}{
+			"tenantID": int64(1),
+		},
+	)
+
+	c.SetParamNames("id")
+	c.SetParamValues("xxx")
+	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
```
petracihalova: I believe you get here Bad request because you set id = "xxx" and not because new source name is empty string (we parse this id in the handler before the validation function in called)

you can use here any valid source id we have in fixtures, e.g. 1

petracihalova (commented):
nice job !!! I have only one comment for tests

petracihalova (approved):
LGTM

lpichler (approved):

PR #546, opened 2022-08-31
[RHCLOUD-21027] Source validation update

petracihalova:
For create source request validation and missing source type id

```
POST /api/sources/v3.1/sources
{
  "name": "source test name"
}
```
we get 400 Bad request + message `bad request: Validation failed: the source type id is not valid`

I think we can improve this and for this case return that source type is missing and cannot be empty

so now we get 400 Bad request + message `bad request: Validation failed: source type id cannot be empty`

**JIRA:** [RHCLOUD-21027](https://issues.redhat.com/browse/RHCLOUD-21027)


MikelAlejoBR (approved):
LGTM!

Ellen-Yi-Dong (approved):
LGTM!

lindgrenj6 (approved):

PR #548, opened 2022-08-31
[RHCLOUD-20914] Add POST /secrets (create action)

lpichler:
This creates secret which is authentication with resource_type="tenant" also
`resource_id` is populated as `tenant_id`. (we were discussing it - but I tend to have consistency data as possible - I don't want potentially break this polymorphic relation by skipping populating `resource_id`)

This doesn't use vault either not optionally.
SecretDao and secret handler created.
New model for secret was not created.

Also I added some other dao functions(DELETE, GetByID) - but in this PR it is only test usage - and they could be changed in other PRs of https://github.com/RedHatInsights/sources-api-go/issues/549.

New route added:

```
POST /api/sources/v3.1/secrets
{
  "extra": {
    "test": {
      "tenant_id": "string"
    }
  },
  "authtype" : "github-token",
  "name": "OpenShift default",
  "password": "NEWPASS",
  "username": "user@example.com"
}
```

Also added possibility to set user ownership: (it is possible with adding `  "user_ownership" : true`)

```
{
  "extra": {
    "azure": {
      "tenant_id": "string"
    }
  },
  "authtype" : "AAAA",
  "name": "OpenShift default",
  "password": "NEWPASS",
  "username": "user@example.com",
  "user_ownership" : true
}

```

### Links

part of https://issues.redhat.com/browse/RHCLOUD-20914
https://github.com/RedHatInsights/sources-api-go/issues/549

### DIFF
--- a/dao/secret_dao.go
+++ b/dao/secret_dao.go
@@ -9,7 +9,7 @@ import (
 	"gorm.io/gorm"
 )
 
-const SecretResourceType = "Tenant"
+const secretResourceType = "Tenant"
 
 var GetSecretDao func(daoParams *RequestParams) SecretDao
 
@@ -47,7 +47,7 @@ func (secret *secretDaoDbImpl) getDb() *gorm.DB {
 	query := DB.Debug().WithContext(secret.ctx)
 	query = query.Where("tenant_id = ?", secret.TenantID)
 
-	query = query.Where("resource_type = ?", SecretResourceType)
+	query = query.Where("resource_type = ?", secretResourceType)
 
 	return query
 }
--- a/dao/secret_dao.go
+++ b/dao/secret_dao.go
@@ -60,7 +60,7 @@ func (secret *secretDaoDbImpl) GetById(id *int64) (*m.Authentication, error) {
 	var secretAuthentication m.Authentication
 
 	err := secret.getDbWithModel().
-		Where("id = ? AND resource_type = ?", id, SecretResourceType).
+		Where("id = ?", id).
 		First(&secretAuthentication).
 		Error
 
--- a/dao/secret_dao.go
+++ b/dao/secret_dao.go
@@ -73,6 +73,8 @@ func (secret *secretDaoDbImpl) GetById(id *int64) (*m.Authentication, error) {
 
 func (secret *secretDaoDbImpl) Create(authentication *m.Authentication) error {
 	authentication.TenantID = *secret.TenantID // the TenantID gets injected in the middleware
+	authentication.ResourceType = secretResourceType
+	authentication.ResourceID = *secret.TenantID
 	if authentication.Password != nil {
 		encryptedValue, err := util.Encrypt(*authentication.Password)
 		if err != nil {
--- a/dao/secret_dao.go
+++ b/dao/secret_dao.go
@@ -1,0 +1,121 @@
+package dao
+
+import (
+	"context"
+	"fmt"
+
+	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/util"
+	"gorm.io/gorm"
+)
+
+const SecretResourceType = "Tenant"
+
+var GetSecretDao func(daoParams *RequestParams) SecretDao
+
+type secretDaoDbImpl struct {
+	TenantID *int64
+	UserID   *int64
+	ctx      context.Context
+}
+
+func getDefaultSecretDao(daoParams *RequestParams) SecretDao {
+	var tenantID, userID *int64
+	var ctx context.Context
+	if daoParams != nil && daoParams.TenantID != nil {
+		tenantID = daoParams.TenantID
+		userID = daoParams.UserID
+		ctx = daoParams.ctx
+	}
+
+	return &secretDaoDbImpl{
+		TenantID: tenantID,
+		UserID:   userID,
+		ctx:      ctx,
+	}
+}
+
+func init() {
+	GetSecretDao = getDefaultSecretDao
+}
+
+func (secret *secretDaoDbImpl) getDb() *gorm.DB {
+	if secret.TenantID == nil {
+		panic("nil tenant found in sourceDaoImpl DAO")
+	}
+
+	query := DB.Debug().WithContext(secret.ctx)
+	query = query.Where("tenant_id = ?", secret.TenantID)
+
+	query = query.Where("resource_type = ?", SecretResourceType)
+
+	return query
+}
+
+func (secret *secretDaoDbImpl) getDbWithModel() *gorm.DB {
+	return secret.getDb().Model(&m.Authentication{})
+}
+
+func (secret *secretDaoDbImpl) GetById(id *int64) (*m.Authentication, error) {
+	var secretAuthentication m.Authentication
+
+	err := secret.getDbWithModel().
+		Where("id = ? AND resource_type = ?", id, SecretResourceType).
+		First(&secretAuthentication).
+		Error
+
+	if err != nil {
+		return nil, util.NewErrNotFound("secret")
+	}
+
+	return &secretAuthentication, nil
+}
+
+func (secret *secretDaoDbImpl) Create(authentication *m.Authentication) error {
+	authentication.TenantID = *secret.TenantID // the TenantID gets injected in the middleware
+	if authentication.Password != nil {
+		encryptedValue, err := util.Encrypt(*authentication.Password)
+		if err != nil {
+			return err
+		}
+
+		authentication.Password = &encryptedValue
+	}
+
+	return DB.
+		Debug().
+		Create(authentication).
+		Error
+}
+
+func (secret *secretDaoDbImpl) Delete(id *int64) error {
+	var authentication m.Authentication
+
+	err := secret.getDb().
+		Where("id = ?", id).
+		First(&authentication).
+		Error
+
+	if err != nil {
+		return util.NewErrNotFound("secret")
+	}
+
+	err = secret.getDb().
+		Delete(&authentication).
+		Error
+
+	if err != nil {
+		return fmt.Errorf(`failed to delete secret with id "%d"`, &id)
+	}
+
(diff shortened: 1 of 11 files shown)


dao/secret_dao.go
```diff
@@ -0,0 +1,121 @@
...
+	return secret.getDb().Model(&m.Authentication{})
+}
+
+func (secret *secretDaoDbImpl) GetById(id *int64) (*m.Authentication, error) {
+	var secretAuthentication m.Authentication
+
+	err := secret.getDbWithModel().
+		Where("id = ? AND resource_type = ?", id, SecretResourceType).
+		First(&secretAuthentication).
+		Error
```
MikelAlejoBR: I think that the `SecretResourceType` is already specified on the `getDb()` function, right? `getDbWithModel` calls `getDb()` which as a `query.Where("resource_type =?", SecretResourceType)`

secret_handlers.go
```diff
@@ -0,0 +1,81 @@
...
+
+func SecretCreate(c echo.Context) error {
+	secretDao, err := getSecretDao(c)
+	if err != nil {
+		return err
+	}
+
+	createInputRequest := struct {
+		m.AuthenticationCreateRequest
+		UserOwnership bool `json:"user_ownership"`
```
MikelAlejoBR: Is the `UserOwnership` thing here for future usage? Because I don't see it being used :eyes:

service/secret_validation.go
```diff
@@ -0,0 +1,31 @@
...
+		}
+	}
+
+	if auth.ResourceIDRaw != nil {
+		return fmt.Errorf("resource_id is not applicable to create secret")
+	}
+
+	if auth.ResourceType != "" && !strings.EqualFold(auth.ResourceType, dao.SecretResourceType) {
+		return fmt.Errorf("invalid resource_type - must be Tenant(default) or empty")
+	}
```
MikelAlejoBR: I think using `EqualFold` here would allow having strings like `TeNaNt` or `tenaNT` or whatever... I think we should enforce it to be `Tenant` all the time.

MikelAlejoBR (approved):
A few comments, other than that I think it's looking good!

dao/secret_dao.go
lpichler: yes that is right, very good eye! 👀  (it is fixed now)

secret_handlers.go
lpichler: it is used on line 71 - https://github.com/RedHatInsights/sources-api-go/pull/548/commits/5e5dbd7e0ef7d491aefa67f92d58255d9dd31658#diff-d4b47aa769b5b534f06c5e9e6bcd67f7a98713cccf1e857610cf76b829cb5b52R71 - it enables to setting `user_id`

service/secret_validation.go
lpichler: yes that is right ! - we definitely want it as you mentioned - I fixed it - thanks!

secret_handlers.go
MikelAlejoBR: Oh my god :man_facepalming: I cannot believe I missed that. Sorry about that :joy:

MikelAlejoBR (approved):

secret_handlers_test.go
```diff
@@ -181,7 +190,15 @@ func createSecretRequest(t *testing.T, requestBody *m.AuthenticationCreateReques
...
-	err = SecretCreate(c)
+
+	tenancy := middleware.Tenancy(SecretCreate)
+	userCatcher := middleware.UserCatcher(tenancy)
+
+	testUserId := "testUser"
+	identityHeader := testutils.IdentityHeaderForUser(testUserId)
+	c.Set("identity", identityHeader)
+
+	err = userCatcher(c)
```
petracihalova, line 215: I believe it is not needed to use here middleware.Tenancy(), tests will pass without it

petracihalova (approved):
LGTM

secret_handlers.go
```diff
@@ -0,0 +1,81 @@
...
+		return util.NewErrBadRequest(`invalid JSON given in "extra" field`)
+	}
+
+	secret := &m.Authentication{
+		Name:         createRequest.Name,
+		AuthType:     createRequest.AuthType,
+		Username:     createRequest.Username,
+		Password:     createRequest.Password,
+		ExtraDb:      extraDb,
+		ResourceType: dao.SecretResourceType,
```
lindgrenj6: Rather than setting this here - what would you think about setting this in the DAO's Create method? Same with the resource_id. 

Edit: we already manually override resource_id it looks like - so I would say definitely pull these out of the create request struct.

service/secret_validation.go
```diff
@@ -0,0 +1,30 @@
...
+		}
+	}
+
+	if auth.ResourceIDRaw != nil {
+		return fmt.Errorf("resource_id is not applicable to create secret")
+	}
+
+	if auth.ResourceType != "" && auth.ResourceType != dao.SecretResourceType {
+		return fmt.Errorf("invalid resource_type - must be Tenant(default) or empty")
+	}
```
lindgrenj6: I think we can remove both of these checks since we should be setting them and not letting the user set them.

secret_handlers.go
```diff
@@ -0,0 +1,81 @@
...
+func SecretCreate(c echo.Context) error {
+	secretDao, err := getSecretDao(c)
+	if err != nil {
+		return err
+	}
+
+	createInputRequest := struct {
+		m.AuthenticationCreateRequest
+		UserOwnership bool `json:"user_ownership"`
+	}{}
```
lindgrenj6: Can we move this to `model/` and give it a name like `SecretCreateRequest`? 

Edit: we also don't want to allow all the same fields like a regular authentication (no resource_type/resource_id) so maybe a new struct would be better for the create struct at least.

lindgrenj6: Also maybe `user_scoped` for the JSON name?

model/authentication.go
```diff
@@ -73,6 +73,17 @@ func (auth *Authentication) ToResponse() *AuthenticationResponse {
 	}
 }
 
+func (auth *Authentication) SecretToResponse() *SecretResponse {
```
lindgrenj6: What would you think about `ToSecretResponse()` to follow our other response names?

dao/secret_dao.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"context"
+	"fmt"
+
+	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/util"
+	"gorm.io/gorm"
+)
+
+const SecretResourceType = "Tenant"
```
lindgrenj6: If we change this to only being set in the create function we should make this private.

dao/secret_dao.go
```diff
@@ -0,0 +1,121 @@
...
+
+var GetSecretDao func(daoParams *RequestParams) SecretDao
+
+type secretDaoDbImpl struct {
+	TenantID *int64
+	UserID   *int64
```
lindgrenj6, line 18: this is kind of a general thing - I was wondering if maybe in the future we should just embed the reqestParams struct in the impl - that way we don't have to copy the fields in the `get**` functions.

lindgrenj6 (commented):
Just a few things mostly around re-using the authentication create request struct and allowing `resource_(type|id)`, otherwise looks good.

secret_handlers_test.go
lpichler: Yes, I probably forgot to add test for it - so I added coverage for it,  thanks!

secret_handlers.go
lpichler: Yes, ResourceID and ResourceType were moved to dao create method.

service/secret_validation.go
lpichler: that is right - removed!

secret_handlers.go
lpichler: It is much better to have SecretCreateRequest with only required field - that's also solve other things like validation as mentioned.

Thanks for suggestions!

> Also maybe user_scoped for the JSON name?

I used this name also for field in struct, I guess this field name and json name should same.

dao/secret_dao.go
lpichler: that would work - great idea - let's do in other PR.

lpichler:
Thank you for review! I added all your suggestions!

routes.go
```diff
@@ -15,8 +15,9 @@ var listMiddleware = []echo.MiddlewareFunc{middleware.SortAndFilter, middleware.
 var tenancyMiddleware = []echo.MiddlewareFunc{middleware.Tenancy, middleware.LoggerFields, middleware.UserCatcher}
 
 var tenancyWithListMiddleware = append(tenancyMiddleware, listMiddleware...)
-var permissionMiddleware = append(tenancyMiddleware, []echo.MiddlewareFunc{middleware.PermissionCheck, middleware.RaiseEvent}...)
+var permissionMiddleware = append(permissionMiddlewareWithoutEvents, middleware.RaiseEvent)
 var permissionWithListMiddleware = append(listMiddleware, middleware.PermissionCheck)
+var permissionMiddlewareWithoutEvents = append(tenancyMiddleware, middleware.PermissionCheck)
```
MikelAlejoBR, line 20: Aren't `permissionMIddleware` and `permissionMIddlewareWithoutEvents` the same in this case?

MikelAlejoBR (approved):
LGTM!

routes.go
lpichler: `permissionMIddlewareWithoutEvents` doesn't have `middleware.RaiseEvent` - it is difference - Is it OK ? or I am missing something ?

MikelAlejoBR: :facepalm: you're right... I swear I thought I had seen `middleware.RaisEvent` in the `permissionMiddlewareWithoutEvents` and that's why I said it... Sorry! :sweat_smile:

lindgrenj6 (approved):

PR #550, opened 2022-09-01
[RHCLOUD-21039] Application create returns 500 instead of 400 for existing combination of app type id and source id

petracihalova:
When we create new application and in db already exists application of same app type + same source id and tenant id, the 500 is returned because of unique constraint in the applications table (unique combination for app type id + source id + tenant id) ... and we are returning in this case every error as 500

```
{
    "errors": [
        {
            "detail": "Internal Server Error: ERROR: duplicate key value violates unique constraint \"applications_app_type_id_source_id_tenant_id_idx\" (SQLSTATE 23505)",
            "status": "500"
        }
    ]
}
```

I think we want in this case return 400 Bad Request because invalid input data was provided

one option is to add new check into create application request validation function - this check should be probably some db query or we could reuse some existing app dao method ... but this increase number of db calls 

so my idea was to distinguish error returned after create application query and if specific error code and message is returned then we send descriptive message nad 400 err type

**JIRA:** [RHCLOUD-21039](https://issues.redhat.com/browse/RHCLOUD-21039)

### DIFF
--- a/dao/application_dao.go
+++ b/dao/application_dao.go
@@ -3,11 +3,14 @@ package dao
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
+	"github.com/jackc/pgconn"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 )
@@ -170,6 +173,15 @@ func (a *applicationDaoImpl) Create(app *m.Application) error {
 	app.TenantID = *a.TenantID
 	result := DB.Debug().Create(app)
 
+	// Check if specific error code is returned
+	var pgErr *pgconn.PgError
+	if errors.As(result.Error, &pgErr) {
+		// unique constraint violation for index (source id + app type id + tenant id)
+		if pgErr.Code == PG_UNIQUE_CONSTRAINT_VIOLATION && strings.Contains(pgErr.Detail, "Key (source_id, application_type_id, tenant_id)") {
+			message := fmt.Sprintf("Application of application type = %d already exists for the source id = %d", app.ApplicationTypeID, app.SourceID)
+			return util.NewErrBadRequest(message)
+		}
+	}
 	return result.Error
 }
 
--- a/dao/db.go
+++ b/dao/db.go
@@ -25,6 +25,12 @@ var (
 	conf = config.Get()
 )
 
+// PostgreSQL Error Codes
+const (
+	// PG_UNIQUE_CONSTRAINT_VIOLATION is PostgreSQL error code for unique index violation (more here: https://www.postgresql.org/docs/current/errcodes-appendix.html)
+	PG_UNIQUE_CONSTRAINT_VIOLATION = "23505"
+)
+
 func Init() {
 	l := &logging.GormLogger{
 		SkipErrorRecordNotFound: true,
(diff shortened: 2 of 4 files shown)


petracihalova:
❗️I believe tests failed because they expect for this case the return code 500 and previous error message

dao/application_dao.go
```diff
@@ -170,6 +172,15 @@ func (a *applicationDaoImpl) Create(app *m.Application) error {
 	app.TenantID = *a.TenantID
 	result := DB.Debug().Create(app)
 
+	// Check if specific error code is returned
+	switch e := result.Error.(type) {
+	case *pgconn.PgError:
+		// unique constraint violation (code 23505)
+		if e.Code == "23505" && strings.Contains(e.Detail, "Key (source_id, application_type_id, tenant_id)") {
```
MikelAlejoBR: I would suggest moving this magic number to a constant like `cost uniqueConstraintViolation`.

MikelAlejoBR (approved):
Just a comment, other than that it looks good!

dao/application_dao.go
petracihalova: that's actually a good point... thanks

MikelAlejoBR (approved):
Nice! LGTM!

dao/application_dao.go
```diff
@@ -170,6 +172,15 @@ func (a *applicationDaoImpl) Create(app *m.Application) error {
...
 
+	// Check if specific error code is returned
+	switch e := result.Error.(type) {
+	case *pgconn.PgError:
+		// unique constraint violation for index (source id + app type id + tenant id)
+		if e.Code == uniqueConstraintViolation && strings.Contains(e.Detail, "Key (source_id, application_type_id, tenant_id)") {
+			message := fmt.Sprintf("Application of application type = %d already exists for the source id = %d", app.ApplicationTypeID, app.SourceID)
+			return util.NewErrBadRequest(message)
+		}
+	}
```
lindgrenj6: So rather than switching on the type - we could just use `errors.As` which trys to cast the error for us and returns true if and only if the error is the right type, this is what it would look like:
```suggestion
	var pgErr *pgconn.PgError
	if errors.As(result.Error, &pgErr) {
		// unique constraint violation for index (source id + app type id + tenant id)
		if pgErr.Code == uniqueConstraintViolation && strings.Contains(pgErr.Detail, "Key (source_id, application_type_id, tenant_id)") {
			message := fmt.Sprintf("Application of application type = %d already exists for the source id = %d", app.ApplicationTypeID, app.SourceID)
			return util.NewErrBadRequest(message)
		}
	}
```

Found here: https://github.com/jackc/pgx/wiki/Error-Handling
I wish there were consts for the error codes in the package but unfortunately it doesn't look like the case since this is a pretty low-level database adapter.

dao/db.go
```diff
@@ -25,6 +25,11 @@ var (
 	conf = config.Get()
 )
 
+// Postgres error codes
+const (
+	uniqueConstraintViolation = "23505"
```
lindgrenj6: Can we make this caps, also maybe `PG_UNIQUE_CONSTRAINT_VIOLATION` instead to indicate it's a PG error code.

lindgrenj6 (commented):
One small suggestion, looks good!

dao/application_dao.go
petracihalova: thanks 😁 done

lpichler (approved):

Ellen-Yi-Dong (approved):
LGTM!

dao/db.go
petracihalova: done

Fynardo:
Can confirm that the regression is due to an unexpected error code, so LGTM

lindgrenj6 (approved):

PR #553, opened 2022-09-07
[WIP][RHCLOUD-21109] Refactor of constants in the code base

petracihalova:
main goals:

- check the occurrence of constant's value in code and use it where is needed
- change names of constants from camel case to upper case style 

this PR contains update for first few constants and rest will be part of next PRs

**JIRA:** [RHCLOUD-21109](https://issues.redhat.com/browse/RHCLOUD-21109)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -12,7 +12,7 @@ import (
 	"github.com/spf13/viper"
 )
 
-const KafkaGroupId = "sources-api-go"
+const KAFKA_GROUP_ID = "sources-api-go"
 
 var parsedConfig *SourcesApiConfig
 
@@ -195,7 +195,7 @@ func Get() *SourcesApiConfig {
 		options.SetDefault("FeatureFlagsEnvironment", "development")
 	}
 
-	options.SetDefault("KafkaGroupID", KafkaGroupId)
+	options.SetDefault("KafkaGroupID", KAFKA_GROUP_ID)
 	options.SetDefault("KafkaTopics", kafkaTopics)
 
 	options.SetDefault("LogLevel", os.Getenv("LOG_LEVEL"))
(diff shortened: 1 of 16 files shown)


MikelAlejoBR:
Hey,

I would like to open up a discussion on this, because the convention for the constants seems to be to use `camelCase` instead of full caps separated by underscores. After [checking StackOverflow](https://stackoverflow.com/questions/22688906/go-naming-conventions-for-const), the reasons are:

*  The standard library uses camel case. See [`net/http`](https://cs.opensource.google/go/go/+/master:src/net/http/http.go;l=26) or the [`fmt`](https://cs.opensource.google/go/go/+/master:src/fmt/print.go;l=19) packages, for instance.
* If we use `CAPITALIZED_CONSTS`, we are automatically exporting those constants everywhere.
* [Effective Go recommends using mixed caps for multiword names](https://go.dev/doc/effective_go#mixed-caps).

I would prefer sticking to the convention, if possible. I come from the Java world so I was initially used to `THAT_CONVENTION`, but...

Please let me know what you all think!

petracihalova:
yes, I also thought we were going to use the camel case for constants and after this comment https://github.com/RedHatInsights/sources-api-go/pull/550#discussion_r961804820 I discussed it with @lindgrenj6 and that is why I started with this upper case style update for constants ... Mikel's arguments seem strong ... what do you think @lindgrenj6, @lpichler

lindgrenj6:
Ah I didn't know there was actually a recommendation for it. In the end of
the day I just prefer CAPS_STYLE consts but if there is a precedent for
camelCase for consts then lets just go with that.
--
Jacob Lindgren

Red Hat

PR #551, opened 2022-09-01
Add GET list /secrets endpoint

lpichler:
supports
- user based filtering 
- filtering by parameters

example:

With filtering:

```
GET /secrets?filter[authtype]=AAAA

{
  "data": [
    {
      "id": "37",
      "name": "OpenShift default6",
      "authtype": "AAAA",
      "username": "user@example.com",
      "extra": {
        "azure": {
          "tenant_id": "string"
        }
      }
    },
...

```

### Links

- [x] https://github.com/RedHatInsights/sources-api-go/pull/548
- part of https://issues.redhat.com/browse/RHCLOUD-20914
- https://github.com/RedHatInsights/sources-api-go/issues/549

Issue #549, opened 2022-08-31
Developer Sources


lpichler:
- [x] https://github.com/RedHatInsights/sources-api-go/pull/548
- [x] https://github.com/RedHatInsights/sources-api-go/pull/551
- [x] https://github.com/RedHatInsights/sources-api-go/pull/555
- [x] https://github.com/RedHatInsights/sources-api-go/pull/560
- [x] https://github.com/RedHatInsights/sources-api-go/pull/558

PR #555, opened 2022-09-08
ADD GET /secrets/:id with internal corresponding endpoint

lpichler:
### Links 
- [x] https://github.com/RedHatInsights/sources-api-go/pull/551

### DIFF
--- a/internal_handlers.go
+++ b/internal_handlers.go
@@ -93,3 +93,22 @@ func TranslateTenants(c echo.Context) error {
 
 	return c.JSON(http.StatusOK, response)
 }
+
+func InternalSecretGet(c echo.Context) error {
+	secretDao, err := getSecretDao(c)
+	if err != nil {
+		return err
+	}
+
+	paramID, err := util.InterfaceToInt64(c.Param("id"))
+	if err != nil {
+		return err
+	}
+
+	secret, err := secretDao.GetById(&paramID)
+	if err != nil {
+		return err
+	}
+
+	return c.JSON(http.StatusOK, secret.ToInternalSecretResponse())
+}
--- a/internal_handlers.go
+++ b/internal_handlers.go
@@ -102,7 +102,7 @@ func InternalSecretGet(c echo.Context) error {
 
 	paramID, err := util.InterfaceToInt64(c.Param("id"))
 	if err != nil {
-		return err
+		return util.NewErrBadRequest(err)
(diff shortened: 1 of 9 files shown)


internal_handlers.go
```diff
@@ -93,3 +93,22 @@ func TranslateTenants(c echo.Context) error {
...
+
+func InternalSecretGet(c echo.Context) error {
+	secretDao, err := getSecretDao(c)
+	if err != nil {
+		return err
+	}
+
+	paramID, err := util.InterfaceToInt64(c.Param("id"))
+	if err != nil {
+		return err
```
MikelAlejoBR: I think this one needs to be a `util.NewErrBadRequest`

secret_handlers.go
```diff
@@ -101,3 +101,22 @@ func SecretList(c echo.Context) error {
...
+
+func SecretGet(c echo.Context) error {
+	secretDao, err := getSecretDao(c)
+	if err != nil {
+		return err
+	}
+
+	paramID, err := util.InterfaceToInt64(c.Param("id"))
+	if err != nil {
+		return err
```
MikelAlejoBR: Same here.

MikelAlejoBR (approved):
Just a couple of minor comments. Other than that looks good!

secret_handlers_test.go
```diff
@@ -489,3 +489,149 @@ func TestSecretListBadRequestInvalidFilter(t *testing.T) {
...
+			map[string]interface{}{
+				"tenantID": tenantIDForSecret,
+			},
+		)
+
+		c.SetParamNames("id")
+		c.SetParamValues(secretID)
+
+		tenancy := middleware.Tenancy(SecretGet)
+		userCatcher := middleware.UserCatcher(tenancy)
```
petracihalova: I think the middleware.Tenancy() is not needed here, so this will work too
`userCatcher := middleware.UserCatcher(SecretGet)`

petracihalova (approved):
LGTM, only one small suggestion

internal_handlers.go
lpichler: That is right, thanks!

lpichler:
I added your suggestion, thanks!

MikelAlejoBR (approved):

lindgrenj6 (approved):

PR #557, opened 2022-09-08
[RHCLOUD-21120] Bind() update - added check for no body

petracihalova:
create an application without body
`POST api/sources/v3.1/applications`
and you get 400 Bad request
```
{
    "errors": [
        {
            "detail": "bad request: EOF",
            "status": "400"
        }
    ]
}
```

in this PR is added check if request contains a body ... otherwise Bad request is returned with more descriptive message
```
{
    "errors": [
        {
            "detail": "bad request: no body",
            "status": "400"
        }
    ]
}
```

**JIRA:** [RHCLOUD-21120](https://issues.redhat.com/browse/RHCLOUD-21120)

### DIFF
--- a/util/echo/binder.go
+++ b/util/echo/binder.go
@@ -2,6 +2,7 @@ package echo
 
 import (
 	"encoding/json"
+	"net/http"
 	"reflect"
 
 	"github.com/RedHatInsights/sources-api-go/util"
@@ -21,6 +22,11 @@ type NoUnknownFieldsBinder struct{}
 func (binder *NoUnknownFieldsBinder) Bind(i interface{}, c echo.Context) error {
 	// Close the request's body after we're done with it.
 	defer c.Request().Body.Close()
+
+	// Check if request contains a body
+	if c.Request().Body == http.NoBody {
+		return util.NewErrBadRequest("no body")
+	}
 
 	// is the struct passed in initially empty? then it's pretty easy to tell if
 	// no fields were plucked from the body.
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (approved):
LGTM!

Ellen-Yi-Dong (approved):
LGTM!

binder.go
```diff
@@ -26,6 +27,11 @@ func (binder *NoUnknownFieldsBinder) Bind(i interface{}, c echo.Context) error {
 	// no fields were plucked from the body.
 	isInitiallyEmpty := reflect.ValueOf(i).Elem().IsZero()
 
+	// Check if request contains a body
+	if c.Request().Body == http.NoBody {
```
lindgrenj6: Can we check if the body is `nil` here too? I think we were checking for both before because iirc a `net/http` request can have either of those options when there isn't a body associated with the request. 

Also can we move this check to the top - no need to run the reflection checks if there isn't a body.

petracihalova: ok, I will change it ... but I dont know how to test the "nil" value here because everything I tried lead to http.NoBody ... any idea? 🤔

lindgrenj6: Turns out I misread the docs, we don't need the check: https://pkg.go.dev/net/http#Request.Body

The first part specifies that a *client* request can have a nil body, whereas a server request will always have a non-nil body (e.g. one that returns `EOF` and/or is `http.NoBody`).

lindgrenj6 (approved):

lpichler (approved):

PR #558, opened 2022-09-08
Add PATCH /sources/:id endpoint

lpichler:
### Links 
- [x] https://github.com/RedHatInsights/sources-api-go/pull/555

### DIFF
--- a/model/authentication_http.go
+++ b/model/authentication_http.go
@@ -102,3 +102,24 @@ func (auth *Authentication) UpdateFromRequest(update *AuthenticationEditRequest)
 
 	return nil
 }
+
+func (auth *Authentication) UpdateSecretFromRequest(update *SecretEditRequest) error {
+	if update.Password != nil {
+		encrypted, err := util.Encrypt(*update.Password)
+		if err != nil {
+			return err
+		}
+		auth.Password = &encrypted
+	}
+
+	if update.Extra != nil {
+		var err error
+		auth.ExtraDb, err = json.Marshal(*update.Extra)
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
--- a/model/authentication_http.go
+++ b/model/authentication_http.go
@@ -119,7 +119,6 @@ func (auth *Authentication) UpdateSecretFromRequest(update *SecretEditRequest) e
 	if update.Extra != nil {
 		var err error
 		auth.ExtraDb, err = json.Marshal(*update.Extra)
-
 		if err != nil {
 			return err
 		}
--- a/model/authentication_http.go
+++ b/model/authentication_http.go
@@ -104,6 +104,10 @@ func (auth *Authentication) UpdateFromRequest(update *AuthenticationEditRequest)
 }
 
 func (auth *Authentication) UpdateSecretFromRequest(update *SecretEditRequest) error {
+	if update.Username != nil {
+		auth.Username = update.Username
+	}
+
 	if update.Password != nil {
 		encrypted, err := util.Encrypt(*update.Password)
 		if err != nil {
--- a/model/secret_http.go
+++ b/model/secret_http.go
@@ -29,6 +29,7 @@ type SecretCreateRequest struct {
 }
 
 type SecretEditRequest struct {
+	Username *string                 `json:"username"`
 	Password *string                 `json:"password,omitempty"`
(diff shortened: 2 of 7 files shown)


model/secret_http.go
```diff
@@ -27,3 +27,8 @@ type SecretCreateRequest struct {
 	Extra      map[string]interface{} `json:"extra,omitempty"`
 	UserScoped bool                   `json:"user_scoped"`
 }
+
+type SecretEditRequest struct {
+	Password *string                 `json:"password,omitempty"`
```
lindgrenj6, line 33: I don't think we want to allow password to be updated, the only thing that should be update-able should be the extra column.

lpichler: I am ok with removing it but if I remember it correctly we were discussing it on meeting and we decided to have possibility to update also password` I also did change in [docs](https://docs.google.com/document/d/1dokYxzbqIAQrG7UYK17zROy4QTSnTwGgQifMWbjQ7VU/edit#) on that meeting.

It looks we allow to update password on Auth. model as well.

Should I still remove it ?

lindgrenj6:
Hmm. I forgot about that. I suppose we should leave it in then - but
probably better add in username too, it doesn't really make sense to allow
password without username.
--
Jacob Lindgren

Red Hat

lindgrenj6 (approved):

model/authentication_http.go
```diff
@@ -102,3 +102,24 @@ func (auth *Authentication) UpdateFromRequest(update *AuthenticationEditRequest)
...
+	}
+
+	if update.Extra != nil {
+		var err error
+		auth.ExtraDb, err = json.Marshal(*update.Extra)
+
+		if err != nil {
+			return err
+		}
+	}
```
petracihalova: I am curious if there is some special reason to not have only
`auth.ExtraDb, err := json.Marshal(*update.Extra)`
(because we usually use short variable declaration in this cases)

secret_handlers.go
```diff
@@ -120,3 +120,38 @@ func SecretGet(c echo.Context) error {
...
+	secretDao, err := getSecretDao(c)
+	if err != nil {
+		return err
+	}
+
+	updateRequest := &m.SecretEditRequest{}
+	err = c.Bind(updateRequest)
+	if err != nil {
+		return util.NewErrBadRequest(err)
+	}
```
petracihalova: here you can return only `err` because our Bind method returns Bad Request err objects

secret_handlers.go
```diff
@@ -120,3 +120,38 @@ func SecretGet(c echo.Context) error {
...
+
+	paramID, err := util.InterfaceToInt64(c.Param("id"))
+	if err != nil {
+		return err
+	}
```
petracihalova: I believe here you should return` util.NewErrBadRequest(err)` because it is a bad request when something wrong is provided as an id

secret_handlers.go
```diff
@@ -120,3 +120,38 @@ func SecretGet(c echo.Context) error {
...
+
+	err = secret.UpdateSecretFromRequest(updateRequest)
+	if err != nil {
+		return util.NewErrBadRequest(err)
+	}
+
+	err = secretDao.Update(secret)
+	if err != nil {
+		return util.NewErrBadRequest(err)
+	}
```
petracihalova: just thinking 🤔 when provided id exists and `secret.UpdateSecretFromRequest()` doesn't return an err, what situation will lead to bad request err here? I think we can have here only `return err`

model/authentication_http.go
lpichler: it looks that it is not possible:
`Syntax error: unexpected end of the statement, expecting ':=','=', or ','`

Probably - it is because you need var on left side instead of `auth.ExtraDb`.
(I just removed empty line)

secret_handlers.go
lpichler: yes 5xx is more relevant here, thanks!

lpichler:
@petracihalova I addressed your suggestions, thanks!

MikelAlejoBR (approved):
LGTM!

model/authentication_http.go
petracihalova: ok, thanks

petracihalova (approved):

lpichler:
3 Approvals I am merging this PR.

PR #559, opened 2022-09-09
[RHCLOUD-21150] Use Red Hat IT's custom sonar scanner container image

MikelAlejoBR:
The IT team has built a custom image which already has the keystore
generated with Red Hat's IT custom certificate in it. This simplifies
the scan process since we only have to mount the code in the container
and run the scanner, instead of having to create the keystore ourselves
and having to download the scanner every time the pipeline ran.

## Links

[[RHCLOUD-21150]](https://issues.redhat.com/browse/RHCLOUD-21150)


lindgrenj6 (approved):
-144 :drooling_face:

lpichler (approved):

PR #560, opened 2022-09-09
Add DELETE /secrets/:id 

lpichler:
### Links 
- [x] https://github.com/RedHatInsights/sources-api-go/pull/558

### DIFF
--- a/routes.go
+++ b/routes.go
@@ -114,6 +114,7 @@ func setupRoutes(e *echo.Echo) {
 		r.GET("/secrets/:id", SecretGet, tenancyMiddleware...)
 		r.POST("/secrets", SecretCreate, permissionMiddlewareWithoutEvents...)
 		r.PATCH("/secrets/:id", SecretEdit, permissionMiddlewareWithoutEvents...)
+		r.DELETE("/secrets/:uid", SecretDelete, permissionMiddleware...)
 
 		// SourceTypes
 		r.GET("/source_types", SourceTypeList, append(listMiddleware, middleware.LoggerFields)...)
--- a/routes.go
+++ b/routes.go
@@ -114,7 +114,7 @@ func setupRoutes(e *echo.Echo) {
 		r.GET("/secrets/:id", SecretGet, tenancyMiddleware...)
 		r.POST("/secrets", SecretCreate, permissionMiddlewareWithoutEvents...)
 		r.PATCH("/secrets/:id", SecretEdit, permissionMiddlewareWithoutEvents...)
-		r.DELETE("/secrets/:uid", SecretDelete, permissionMiddleware...)
+		r.DELETE("/secrets/:id", SecretDelete, permissionMiddleware...)
 
 		// SourceTypes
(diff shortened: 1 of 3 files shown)


routes.go
```diff
@@ -114,6 +114,7 @@ func setupRoutes(e *echo.Echo) {
 		r.GET("/secrets/:id", SecretGet, tenancyMiddleware...)
 		r.POST("/secrets", SecretCreate, permissionMiddlewareWithoutEvents...)
 		r.PATCH("/secrets/:id", SecretEdit, permissionMiddlewareWithoutEvents...)
+		r.DELETE("/secrets/:uid", SecretDelete, permissionMiddleware...)
```
petracihalova: not big deal, but I can see that for GET and PATCH you use "id" and for DELETE "uid" ... maybe we can have it same for all endpoints

secret_handlers.go
```diff
@@ -155,3 +155,22 @@ func SecretEdit(c echo.Context) error {
...
+	secretDao, err := getSecretDao(c)
+	if err != nil {
+		return err
+	}
+
+	secretID, err := util.InterfaceToInt64(c.Param("id"))
+	if err != nil {
+		return err
+	}
+
```
petracihalova: I believe here we can have `return util.NewErrBadRequest(err)`

secret_handlers_test.go
```diff
@@ -879,3 +879,147 @@ func TestSecretEditBadRequest(t *testing.T) {
...
+
+	requestBody := m.SecretEditRequest{
+		Extra:    &secretExtra,
+		Password: util.StringRef(password),
+	}
+
+	body, err := json.Marshal(requestBody)
+	if err != nil {
+		t.Error("Could not marshal JSON")
+	}
```
petracihalova: I think you don't need request body for DELETE the secret

secret_handlers_test.go
```diff
@@ -879,3 +879,147 @@ func TestSecretEditBadRequest(t *testing.T) {
...
+
+		if rec.Code != http.StatusNoContent {
+			t.Error("Did not return 204")
+		}
+
+		secretDao := dao.GetSecretDao(&dao.RequestParams{TenantID: &tenantIDForSecret, UserID: userID})
+		_, err = secretDao.GetById(&secret.DbID)
+		if err.Error() != "secret not found" {
+			t.Error("'secret not found' expected")
+		}
```
petracihalova: maybe here we can check if returned err is the type we want instead of compare the strings, something like: 
```
		if !errors.Is(err, util.ErrNotFoundEmpty) {
			t.Error("'secret not found' expected")
		}
```

petracihalova (changes requested):
LGTM, I have only small comments ⭐️

MikelAlejoBR (approved):
LGTM!

lpichler:
@petracihalova I updated PR according to your suggestions.

petracihalova (approved):

lindgrenj6 (approved):

PR #561, opened 2022-09-12
[RHCLOUD-21159] feature: update source types for the provisioning application

MikelAlejoBR:
The provisioning team is working on making the application compatible with Azure and Google Cloud. They need new authentication types for those clouds and the ability to create sources of that type.

## Links

[[RHCLOUD-21159]](https://issues.redhat.com/browse/RHCLOUD-21159)

### DIFF
--- a/dao/seeds/source_types.yml
+++ b/dao/seeds/source_types.yml
@@ -114,6 +114,17 @@ google:
         isRequired: true
         validate:
         - type: required
+    - type: provisioning_project_id
+      name: Provisioning — project ID
+      fields:
+        - component: text-field
+          name: authentication.authtype
+          hideField: true
+          initializeOnMount: true
+          initialValue: provisioning_project_id
+        - component: text-field
+          name: authentication.username
+          label: Project ID
     endpoint:
       hidden: true
       fields:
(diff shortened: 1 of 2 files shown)


dao/seeds/source_types.yml
```diff
@@ -114,6 +114,24 @@ google:
         isRequired: true
         validate:
         - type: required
+    - type: provisioning_project_id_service_account_json
+      name: Provisioning — project ID and Service Account JSON
```
adiabramovitch: I guess this should be only project id

MikelAlejoBR: Ooops, you're right. Thanks!

dao/seeds/source_types.yml
```diff
@@ -143,6 +154,20 @@ azure:
...
+          name: authentication.authtype
+          hideField: true
+          initializeOnMount: true
+          initialValue: provisioning_lighthouse_subscription_id
+        - component: text-field
+          name: authentication.username
+          label: Subscription ID
+          isRequired: true
+          validate:
+            - type: required
```
lzap, line 170: Can confirm this works fine on our end, assuming there are no other changes from the existing type other than the type/name field :-)

lindgrenj6 (approved):

lpichler (approved):

lzap:
We forgot about OpenAPI enums, here is a follow-up PR: https://github.com/RedHatInsights/sources-api-go/pull/566

PR #562, opened 2022-09-12
[WIP][RHCLOUD-21109] Refactor of constants in middleware/headers/headers.go

petracihalova:
this PR contains:
- rename of constants from headers.go from upper case style into camel case style
- replacing strings with existing constants (only related with constants from headers.go)

refactor of rest of constants will be solved in next PRs

**JIRA:** [RHCLOUD-21109](https://issues.redhat.com/browse/RHCLOUD-21109)

### DIFF
--- a/application_handlers_test.go
+++ b/application_handlers_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/internal/testutils/templates"
 	"github.com/RedHatInsights/sources-api-go/kafka"
 	"github.com/RedHatInsights/sources-api-go/middleware"
+	h "github.com/RedHatInsights/sources-api-go/middleware/headers"
 	m "github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/service"
 	"github.com/RedHatInsights/sources-api-go/util"
@@ -833,7 +834,7 @@ func TestApplicationEdit(t *testing.T) {
 	c.SetParamNames("id")
 	c.SetParamValues(strconv.Itoa(int(applicationID)))
 	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
-	c.Set("identity", &identity.XRHID{Identity: identity.Identity{AccountNumber: fixtures.TestTenantData[0].ExternalTenant}})
+	c.Set(h.ParsedIdentityKey, &identity.XRHID{Identity: identity.Identity{AccountNumber: fixtures.TestTenantData[0].ExternalTenant}})
 
 	appEditHandlerWithNotifier := middleware.Notifier(ApplicationEdit)
 	err = appEditHandlerWithNotifier(c)
--- a/application_handlers_test.go
+++ b/application_handlers_test.go
@@ -834,7 +834,7 @@ func TestApplicationEdit(t *testing.T) {
 	c.SetParamNames("id")
 	c.SetParamValues(strconv.Itoa(int(applicationID)))
 	c.Request().Header.Add("Content-Type", "application/json;charset=utf-8")
-	c.Set(h.ParsedIdentityKey, &identity.XRHID{Identity: identity.Identity{AccountNumber: fixtures.TestTenantData[0].ExternalTenant}})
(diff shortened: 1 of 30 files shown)


MikelAlejoBR (approved):
LGTM, although I would vote for removing the `*key` suffix from the constants.

petracihalova:
> LGTM, although I would vote for removing the `*key` suffix from the constants.

the idea here was that values in these constants are keys, but I am ok with both options 😁😁😁

lindgrenj6 (commented):
I'm kind of with mikel on this one - I would prefer them just `h.ParsedIdentity`.

The main reason I say that is because like in the `net/http` standard library the constant status codes are `http.StatusOK` rather than `http.StatusOKCode` or something like that. It's just implied that they are codes.

petracihalova:
> I'm kind of with mikel on this one - I would prefer them just `h.ParsedIdentity`.
> 
> The main reason I say that is because like in the `net/http` standard library the constant status codes are `http.StatusOK` rather than `http.StatusOKCode` or something like that. It's just implied that they are codes.

🤔 what about to add to these constants the "Xrh" prefix to distinguish between e.g. account number or org id as constants for "x-rh-sources-account-number" or "x-rh-sources-org-id" from headers and fields e.g. in Identity struct and other places where we are using these names

```
// Identity is the main body of the XRHID
type Identity struct {
	AccountNumber         string    `json:"account_number,omitempty"`
	EmployeeAccountNumber string    `json:"employee_account_number,omitempty"`
	OrgID                 string    `json:"org_id"`
...
```

MikelAlejoBR:
> > I'm kind of with mikel on this one - I would prefer them just `h.ParsedIdentity`.
> > The main reason I say that is because like in the `net/http` standard library the constant status codes are `http.StatusOK` rather than `http.StatusOKCode` or something like that. It's just implied that they are codes.
> 
> thinking what about to add to these constants the "Xrh" prefix to distinguish between e.g. account number or org id as constants for "x-rh-sources-account-number" or "x-rh-sources-org-id" from headers and fields e.g. in Identity struct and other places where we are using these names
> 
> ```
> // Identity is the main body of the XRHID
> type Identity struct {
> 	AccountNumber         string    `json:"account_number,omitempty"`
> 	EmployeeAccountNumber string    `json:"employee_account_number,omitempty"`
> 	OrgID                 string    `json:"org_id"`
> ...
> ```

But wouldn't that be covered by the module's name? I mean:

```go
headers.OrgId // I know we're using the x-rh-sources-org-id header here.
```
In the case of the identity struct...

```go
identity.OrgId // I know we're using the identity field's value here.
```

Maybe there is a place where we are using both and it is not clear enough?

petracihalova:
I believe it is covered by the module's name ...  I wanted to make it clear before I started changing in so many places

petracihalova:
constants renamed ("key" suffix removed)

lpichler:
I have to 2 insights:
1) I noticed that you also converted abbreviation in constant for example:
 PSK -> Psk 
 ORGID -> OrgId
 ... 
I would prefer to use capitals for abbreviation like
PSK
OrgID
...

maybe it is nitpicking and I saw that we have in code base `Id` instead of `ID` etc. but what do you think about having capitals in abbreviation in constants ? cc @petracihalova @MikelAlejoBR @lindgrenj6 

2) can we keep PSK suffix in constant  `XrhUserId` - to have `PSKUserID` ?

MikelAlejoBR:
I'm okay with your suggestions :+1: . As per [github.com/golang/go/wiki/CodeReviewComments#initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms):

>Words in names that are initialisms or acronyms (e.g. "URL" or "NATO") have a consistent case. For example, "URL" should appear as "URL" or "url" (as in "urlPony", or "URLPony"), never as "Url". As an example: ServeHTTP not ServeHttp. For identifiers with multiple initialized "words", use for example "xmlHTTPRequest" or "XMLHTTPRequest".
>
> This rule also applies to "ID" when it is short for "identifier" (which is pretty much all cases when it's not the "id" as in "ego", "superego"), so write "appID" instead of "appId".
>
> Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.

PR #564, opened 2022-09-14
[RHCLOUD-20616] add check if ENCRYPTION_KEY is not set into InitializeEncryption

Ellen-Yi-Dong:
Added check for situation when ENCRYPTION_KEY is not set in InitializeEncryption()

if ENCRYPTION_KEY is empty, then send a panic error 

LINK: [RHCLOUD-20616](https://issues.redhat.com/browse/RHCLOUD-20616)


lpichler:
yes, this will save a time!

util/encryption.go
```diff
@@ -26,7 +26,9 @@ func init() {
 // variable again. Useful for testing purposes outside the "util" package.
 func InitializeEncryption() {
 	key = os.Getenv("ENCRYPTION_KEY")
-
+	if key == "" {
+		_ = os.Setenv("ENCRYPTION_KEY", "YWFhYWFhYWFhYWFhYWFhYQ")
+	}
```
petracihalova: I believe this is not working as we need

if key is empty we set there some default value - but the value is set only into environment variables, the "key" variable stays empty 
+
maybe  we can use here some nice private function to set default value, something like
```
if key == "" {
    key = setDefaultEncryptionKey()
}
```

petracihalova (changes requested):

util/encryption.go
```diff
@@ -35,6 +41,14 @@ func InitializeEncryption() {
...
 	keyPresent = true
 }
+func setDefaultEncryptionKey() error {
+	_ = os.Setenv("ENCRYPTION_KEY", "YWFhYWFhYWFhYWFhYWFhYQ")
+	getKey := os.Getenv("ENCRYPTION_KEY")
+	if getKey != "YWFhYWFhYWFhYWFhYWFhYQ" {
+		return fmt.Errorf("Wrong encryption key: did not set correct default value")
+	}
+	return nil
+}
```
MikelAlejoBR: In this case I'd suggest not ignoring the error, and returning early. If the `Setenv` function returns an error, you know for sure that the value you wanted wasn't set.

MikelAlejoBR (commented):
I'm not sure about the use case for this one. I'm guessing this is for development purposes, so that we don't have to set an encryption key to make it work?

My feeling is that it might be better to simply let it panic as it did before, maybe with a more user friendly message. Just because:

1. We're setting a default encryption key which is public to everyone.
2. If we forget to set a proper encryption key on stage or prod —highly unlikely, I know—, I don't foresee myself in the future saying "oh, I need to check if the encryption key was properly set". Probably I would totally forget about it.

I'd like to hear everyone's views on this, please!

util/encryption.go
```diff
@@ -36,6 +44,42 @@ func InitializeEncryption() {
...
+			body, err := ioutil.ReadFile(filepath)
+			if err != nil {
+				return fmt.Errorf("Unable to read file: %v", err)
+			}
+			os.Setenv("ENCRYPTION_KEY", string(body))
+		}
+	} else {
+		return nil
+	}
+	return nil
```
lindgrenj6: This else block can be removed since it's falling through if we're not on stage/prod - and the final return statement should probably be an error right? Since we want to panic and let the user know we couldn't set a default encryption key.

util/encryption.go
```diff
@@ -36,6 +44,42 @@ func InitializeEncryption() {
 	keyPresent = true
 }
 
+func setDefaultEncryptionKey() error {
+	if os.Getenv("SOURCES_ENV") != "stage" && os.Getenv("SOURCES_ENV") != "prod" {
+		_, filename, _, ok := runtime.Caller(1)
+		if !ok {
+			return fmt.Errorf("Not possible to recover the information")
+		}
```
lindgrenj6: ~I'm not following this - can you explain why we do a runtime call to see where we're at in the program?~

This would be a good example where a comment would help in the future for reference - something like 

`// fetch the filename so we know where to get the source encryption_key.dev file from`

util/encryption.go
```diff
@@ -36,6 +44,42 @@ func InitializeEncryption() {
...
+
+		_, error := os.Stat(filepath)
+		if os.IsNotExist(error) {
+			panic("encryption_key.dev file does not exist. Please import encryption_key.dev file.")
+		} else {
+			body, err := ioutil.ReadFile(filepath)
+			if err != nil {
+				return fmt.Errorf("Unable to read file: %v", err)
+			}
+			os.Setenv("ENCRYPTION_KEY", string(body))
```
lindgrenj6: This is exactly right - the only thing I would suggest is also setting the package var `key` to `string(body)` too since we won't be calling this again (other than maybe in tests).

util/encryption.go
```diff
@@ -36,6 +44,42 @@ func InitializeEncryption() {
 	keyPresent = true
 }
 
+func setDefaultEncryptionKey() error {
+	if os.Getenv("SOURCES_ENV") != "stage" && os.Getenv("SOURCES_ENV") != "prod" {
```
lindgrenj6: Here you should be able to use `config.Get().Env` - we parse it at the beginning of the app startup.

util/encryption.go
```diff
@@ -36,6 +44,42 @@ func InitializeEncryption() {
...
+			return fmt.Errorf("Error within symlink")
+		}
+		if _, err := os.Lstat("example.dev"); err == nil {
+			if err := os.Remove("example.dev"); err != nil {
+				return fmt.Errorf("failed to unlink: %v", err)
+			}
+		} else if os.IsNotExist(err) {
+			return fmt.Errorf("failed to check symlink: %v", err)
+		}
+
```
lindgrenj6: Ohhhhh I see why you needed it now. 

I should have maybe been more specific. We don't want to make the symlink for the user in code - we want to just return a nice error that says "hey you need to symlink `encryption_key.dev` to `encryption_key`". 

That way the user themselves have to run `ln -s ...` we could even put the command in the error message.

lindgrenj6 (commented):
Great next pass! I think we're mostly there - we just need to let the user create the symlink rather than doing it for them.

petracihalova (approved):
nice job

PR #566, opened 2022-09-16
OpenAPI: Add provisioning auth types to authtype enum

lzap:
The new provisioning authtypes are missing from the OpenAPI enum. This itself would not be a huge issue, but what is really important is the missing OpenAPI type (string). Go client generated from the OpenAPI does not unmarshal any unknown (`interface{}` or `any`) types, this leads to be this field to be `nil` and unusable.

This patch fixes that too.

### DIFF
--- a/public/openapi-3-v3.1.json
+++ b/public/openapi-3-v3.1.json
@@ -2540,6 +2540,7 @@
             "type": "string"
           },
           "authtype": {
+            "type": "string",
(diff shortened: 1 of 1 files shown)


MikelAlejoBR (approved):
Nice catch! Thank you for the PR!

lpichler (approved):

PR #567, opened 2022-09-16
Force Database SSL if settings present in Clowder

lindgrenj6:
Looks like we were keeping `sslmode=disable` throughout stage/prod and we don't want to do that. 

Luckily our RDS instances force SSL - so we were probably fine except for the fact we were validating with only the public CA rather than the rds CA. 

This PR pulls the CA from clowder, writes it to a file, and passes along the sslmode + cert location to the db connection strings.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -12,7 +12,10 @@ import (
 	"github.com/spf13/viper"
 )
 
-const KafkaGroupId = "sources-api-go"
+const (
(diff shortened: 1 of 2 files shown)


syncrou (approved):
LGTM 👍

lindgrenj6:
Followups #568 and #569 to get permissions right since OCP runs the entrypoint process as a random UID/GID. `777` is about the only permission that would work.

PR #568, opened 2022-09-16
Fix Permission Denied error when writing SSL CA

lindgrenj6:
Getting a permission denied error when writing the CA Cert, this should fix that.


syncrou (approved):
LGTM 👍

PR #577, opened 2022-09-23
Add a key to kafka messages on the event-stream to ensure ordering per-orgid

lindgrenj6:
Ticket: https://issues.redhat.com/browse/RHCLOUD-21370

This has the potential to fix the out-of-order messaging problems that our sub-apps have been seeing.

The tl;dr is that kafka has the ability to set a `key` on each message - this controls which partition the messages get sent to thus controlling ordering. The problem is that when I read about this feature the first time and was looking into it I thought we'd set the key to the same string every time (e.g. `event-stream`) which would lock ALL messages to one partition and we'd lose the ability to scale.

After approaching the problem again and brainstorming with GM it became apparent that we could use the orgID and various other identifiers per-tenant to ensure that each tenant's messages would be sent and arrive in order, so we'd lose the scaling per-tenant, but still be spreading the load across available partitions. 

---

I am going to throw together a program to validate that this fixes the issue before un-wipping it. 

EDIT: Testing program here: https://gist.github.com/lindgrenj6/f463a85267d89f330a736a5d1795ec35

Video demonstrating: 

https://user-images.githubusercontent.com/5856504/192583258-502fae04-6c6c-4ef6-b843-9c66e7099d23.mp4

### DIFF
--- a/internal/events/event_stream_producer.go
+++ b/internal/events/event_stream_producer.go
@@ -53,6 +53,7 @@ func (esp *EventStreamSender) RaiseEvent(eventType string, payload []byte, heade
 
 	m.AddHeaders(headers)
 	m.AddValue(payload)
+	m.SetKeyFromHeaders()
 
 	if err := kafka.Produce(kf, m); err != nil {
 		return err
--- a/kafka/kafka.go
+++ b/kafka/kafka.go
@@ -141,13 +141,7 @@ func Produce(writer *Writer, message *Message) error {
 	}
 
 	if !message.isEmpty() {
-		err := writer.WriteMessages(
-			context.Background(),
-			kafka.Message{
-				Headers: message.Headers,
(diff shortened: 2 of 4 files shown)


syncrou:
@lindgrenj6 do we have Jira tickets / Epic on this change? The teams that are affected by it will probably want an Epic to watch if we don't.

lindgrenj6:
@syncrou I added a link to the ticket in the description, not sure we'll need an epic on our side or not since this is all of the work (minus promotion) and it should be transparent to the end users except the messages will just always come in order now.

MikelAlejoBR (approved):
LGTM!

syncrou:
@lindgrenj6 sorry about missing that important part.  Thanks so much!

lpichler (approved):

PR #586, opened 2022-09-27
Fix formats in comments

lpichler:
Changes were generated by

```
golangci-lint run -E gofmt --fix
```

When I ran `make lint` - it is displaying errors related to gofmt.

I am not sure - why this is not in PR checks - I tried it with go1.17.-1.19.

### DIFF
--- a/dao/application_dao_test.go
+++ b/dao/application_dao_test.go
@@ -326,7 +326,8 @@ func TestApplicationNotExists(t *testing.T) {
 }
 
 // TestApplicationSubcollectionListWithOffsetAndLimit tests that SubCollectionList() in application dao returns
-//  correct count value and correct count of returned objects
+//
+//	correct count value and correct count of returned objects
 func TestApplicationSubcollectionListWithOffsetAndLimit(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
 	SwitchSchema("offset_limit")
--- a/dao/application_dao_test.go
+++ b/dao/application_dao_test.go
@@ -326,8 +326,7 @@ func TestApplicationNotExists(t *testing.T) {
 }
 
 // TestApplicationSubcollectionListWithOffsetAndLimit tests that SubCollectionList() in application dao returns
-//
-//	correct count value and correct count of returned objects
+// correct count value and correct count of returned objects
 func TestApplicationSubcollectionListWithOffsetAndLimit(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
 	SwitchSchema("offset_limit")
--- a/dao/source_dao_test.go
+++ b/dao/source_dao_test.go
@@ -386,9 +386,9 @@ func TestDeleteSourceNotExists(t *testing.T) {
 
 // TestDeleteCascade is a long test function, but very simple in essence. Essentially what it does is:
 //
-//- It creates source with subresources (apps, endpoints, rhc-connections ...).
-//- Cascade deletes the source with the function under test.
-//- Checks that the deleted subresources and source are the ones that have been created in this very same test.
+// - It creates source with subresources (apps, endpoints, rhc-connections ...).
+// - Cascade deletes the source with the function under test.
+// - Checks that the deleted subresources and source are the ones that have been created in this very same test.
 func TestDeleteCascade(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
 	SwitchSchema("delete")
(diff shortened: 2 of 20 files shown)


MikelAlejoBR:
Yeah, this started popping up on me as well when I upgraded to `go 1.19`. However since the GitHub actions are set up with the `1.16` version, I did not submit a PR.

Should we all simply upgrade everything to the latest one? What do you think?

lpichler:
> Yeah, this started popping up on me as well when I upgraded to `go 1.19`. However since the GitHub actions are set up with the `1.16` version, I did not submit a PR.
> 
> Should we all simply upgrade everything to the latest one? What do you think?

yes I agree to upgrade everything to 1.19 - but I am not aware whether there is some issue with it - it looks good.

@lindgrenj6 do you have any insights about upgrading to latest ?

lpichler:
according to @petracihalova - it looks that errors are displayed also because of gci-lint version.

```
> golangci-lint version
```


`golangci-lint has version 1.44.0 built from 617470f on 2022-01-25T11:22:07Z` -> OK

`golangci-lint has version 1.49.0 built from cc2d97f on 2022-08-23T12:03:20Z` -> errors

lindgrenj6:
The nice thing about Go is there is a compatibility guarantee - we can upgrade locally to 1.19 and it _should_ behave the same even if compiled with an older version as long as we don't use new language features. I know the current UBI is either 1.16 or 1.17, our pr_check.sh hardcodes 1.16 still. 

I've been seeing the golangci-lint errors for a while now but just ignored them. The good news is that these are just comment changes so they shouldn't change the binary at all. As long as there aren't any comments that are like `//go:...` that are special comments at least.

logger/echo_logger.go
```diff
@@ -25,7 +25,7 @@ type EchoLogger struct {
 	*logrus.Entry
 }
 
-/// Wrapping _level_j methods
+// / Wrapping _level_j methods
```
lindgrenj6: For this file - can we just make `///` -> `//`? the syntax of space between the final `/` is kind of weird.

dao/application_dao_test.go
```diff
@@ -326,7 +326,8 @@ func TestApplicationNotExists(t *testing.T) {
 }
 
 // TestApplicationSubcollectionListWithOffsetAndLimit tests that SubCollectionList() in application dao returns
-//  correct count value and correct count of returned objects
+//
+//	correct count value and correct count of returned objects
 func TestApplicationSubcollectionListWithOffsetAndLimit(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
```
petracihalova: we have to fix it manually because there are 2 spaces in the second comment line and the linter "fixes" it by adding an extra new empty line but when you remove one of the spaces the linter is happy

logger/echo_logger.go
lpichler: yeah, I covered it but I forgot to push it, good eye, thanks!

dao/application_dao_test.go
lpichler: yea, I have it but I did not push it, thanks for noticing!

lpichler:
> The nice thing about Go is there is a compatibility guarantee - we can upgrade locally to 1.19 and it _should_ behave the same even if compiled with an older version as long as we don't use new language features. I know the current UBI is either 1.16 or 1.17, our pr_check.sh hardcodes 1.16 still.
> 

In follow-up PR(s):
so does this mean that we can upgrade go version in pr_check, docker-compose and GH actions ? 
Are we going to upgrade all directly  to 1.19 ?

dao/source_dao_test.go
```diff
@@ -796,7 +796,8 @@ func TestSourceListInternalOffsetAndLimit(t *testing.T) {
 }
 
 // TestSourceListForRhcConnectionWithOffsetAndLimit tests that ListForRhcConnection() in source dao returns
-//  correct count value and correct count of returned objects
+//
+//	correct count value and correct count of returned objects
 func TestSourceListForRhcConnectionWithOffsetAndLimit(t *testing.T) {
 	testutils.SkipIfNotRunningIntegrationTests(t)
 	SwitchSchema("offset_limit")
```
petracihalova: please here remove the empty line

lpichler: thanks, fixed!

lindgrenj6 (approved):
I'm not actually sure we can upgrade pr_check.sh, will have to see what versions of Go are available on the jenkins nodes. 

We could probably bump the github actions runner though, which is a simple change.

PR #592, opened 2022-09-30
feature: log the events to help debugging issues

MikelAlejoBR:
The goal of this change is to let us know when certain events weren't raised. This should help us debug issues when an application is reporting that they didn't receive some key events.

Also, the debugging line should help us spot how many events have been raised for certain resource types.

### DIFF
--- a/service/events.go
+++ b/service/events.go
@@ -10,6 +10,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
+	"github.com/sirupsen/logrus"
 )
 
 // Producer instance used to send messages - default just an empty instance of the struct.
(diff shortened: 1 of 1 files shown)


dehort (approved):
Looks good to me!

One thing I might watch out for...how big are the events (resource mode.Event) going to be?  It looks like this will print out the entire event if Debug is enabled.  So it might be an issue if the events are "big".

MikelAlejoBR:
> Looks good to me!
> 
> One thing I might watch out for...how big are the events (resource mode.Event) going to be? It looks like this will print out the entire event if Debug is enabled. So it might be an issue if the events are "big".

Thank you for the tip, yeah. I have ended up reverting the change because these logs didn't provide much information: the payload had some references, instead of the dereferenced structs I was expecting. Also, the errors were being logged on the caller functions so the logs were being duplicated. Thanks again!

PR #597, opened 2022-10-03
Add default ordering for collections and subcollections queries

lpichler:
We don't have default ordering on API endpoints which could cause that resulted collection has random order. 

### Links

https://issues.redhat.com/browse/RHCLOUD-20048

### DIFF
--- a/graph/arguments.go
+++ b/graph/arguments.go
@@ -32,6 +32,9 @@ func parseSortBy(sortBy []*generated_model.SortBy) []util.Filter {
 		sorts[i] = filter
 	}
 
+	if len(sorts) == 0 {
+		sorts = append(sorts, util.Filter{Operation: "sort_by", Value: []string{"id ASC"}})
+	}
 	return sorts
 }
 
--- a/middleware/filtering.go
+++ b/middleware/filtering.go
@@ -76,5 +76,5 @@ func parseSorting(c echo.Context) *util.Filter {
 		}
 	}
 
-	return nil
+	return &util.Filter{Operation: "sort_by", Value: []string{"id ASC"}}
 }


lindgrenj6:
For this what would you think about rather than adding that to every DAO we could just add a sort_by filter in the middleware if the user didn't pass a `sort_by` query parameter, like so:
```diff
diff --git a/middleware/filtering.go b/middleware/filtering.go
index 4da82eb..2567dcb 100644
--- a/middleware/filtering.go
+++ b/middleware/filtering.go
@@ -76,5 +76,6 @@ func parseSorting(c echo.Context) *util.Filter {
 		}
 	}
 
-	return nil
+	// no sort_by was passed - so we'll sort_by the id asc by default
+	return &util.Filter{Operation: "sort_by", Value: []string{"id asc"}}
 }
```

ends up with a sql query to `/source_types` looking like this:
```
{"@timestamp":"2022-10-28T13:12:09.652Z","@version":1,"app":"source-api-go","caller":"github.com/RedHatInsights/sources-api-go/logger.(*GormLogger).Trace","duration":3.68183,"filename":"/home/jlindgren/git/crc/sources/go-rewrite/dao/source_type_dao.go:41","hostname":"mewtwo","labels":{"app":"source-api-go"},"level":"debug","message":"SELECT * FROM \"source_types\" ORDER BY source_types.id asc LIMIT 100","rows":13,"tags":["source-api-go"]}
```

lpichler:
> For this what would you think about rather than adding that to every DAO we could just add a sort_by filter in the middleware if the user didn't pass a `sort_by` query parameter, like so:
> 

That's right - it is much simpler and cleaner solution, thanks @lindgrenj6 ! and ready for review again

lindgrenj6 (approved):

PR #599, opened 2022-10-06
Add liveness/readiness probes to AvailabilityStatusListener deployment

lindgrenj6:
@Ellen-Yi-Dong and I paired on this today after @lpichler found out that we were missing the liveness/readiness probes on 2 deployments:
- availability status listener
- background worker

This PR adds the probes for the availability status listener, but not the background worker. That will be done in a followup PR by Ellen.

### DIFF
--- a/statuslistener/statuslistener_test.go
+++ b/statuslistener/statuslistener_test.go
@@ -446,7 +446,10 @@ func TestConsumeStatusMessage(t *testing.T) {
 
 		sender := MockEventStreamSender{TestSuite: t, StatusMessage: testEntry.StatusMessage}
 		esp := &events.EventStreamProducer{Sender: &sender}
-		avs := AvailabilityStatusListener{EventStreamProducer: esp}
+
+		avs := AvailabilityStatusListener{EventStreamProducer: esp, healthcheck: make(chan struct{}, 10)}
+		go avs.healthCheckConsumer()
+
 		message, _ := json.Marshal(testEntry)
 		avs.ConsumeStatusMessage(kafka.Message{Value: message, Headers: testEntry.MessageHeaders})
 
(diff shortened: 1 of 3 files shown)


Ellen-Yi-Dong (approved):

lpichler:
https://issues.redhat.com/browse/RHCLOUD-21533

statuslistener/statuslistener_test.go
```diff
@@ -446,7 +446,10 @@ func TestConsumeStatusMessage(t *testing.T) {
 
 		sender := MockEventStreamSender{TestSuite: t, StatusMessage: testEntry.StatusMessage}
 		esp := &events.EventStreamProducer{Sender: &sender}
-		avs := AvailabilityStatusListener{EventStreamProducer: esp}
+
+		avs := AvailabilityStatusListener{EventStreamProducer: esp, healthcheck: make(chan struct{}, 10)}
```
lpichler, line 450: `healthcheck: make(chan struct{}, 10}`

Does this mean that we cannot send more `struct{}`s than 10 to channel `healthcheck` ?

According to line 84 (`avs.healthcheck <- struct{}{}`) it looks that we send `struct{}` to channel `healthcheck` for each message what process and I guess it could be more than 10.

I am not sure whether this could be issue at least in theory.

If so maybe we can just send to channel only for health check messages.

lindgrenj6: Yeah - a buffered channel allows async production for up to 10 "slots" like a queue. if we didn't have a buffer it would block until the other thread could pick up off of the channel. 

It theoretically could be a problem, but I doubt the sub-thread would get far behind since all it is doing is updating a timestamp, but I'll move the initial call to the `if event_type == "healthcheck"` block to make sure it never happens.

lpichler: thanks!

PR #606, opened 2022-10-12
Availability check 'scheduled' URL flag

lzap:
A new 'scheduled' flag is accepted for the availability flag which is passed to applications over HTTP in the JSON payload. A followup PR for the background scheduler will be created to set this flag for all scheduled checks. Applications can now decide to perform immediate actions or increase checking priority when the flag is missing (or false), meaning the request was not made from the scheduled job.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,14 +17,14 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/hashicorp/vault/api v1.1.1
 	github.com/iancoleman/strcase v0.2.0
-	github.com/jackc/pgconn v1.11.0 // indirect
+	github.com/jackc/pgconn v1.11.0
 	github.com/jackc/pgx/v4 v4.15.0
 	github.com/jinzhu/now v1.1.5 // indirect
 	github.com/labstack/echo-contrib v0.12.0
 	github.com/labstack/echo/v4 v4.6.1
 	github.com/labstack/gommon v0.3.1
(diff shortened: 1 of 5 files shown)


lzap:
Opinions? Thanks.

lindgrenj6:
After thinking about this for a bit I wanted to mainly make it a bit more extendable than "just" a scheduled flag, so I have a proposal:

Since the route is `POST /sources/:id/check_availability`, but we don't currently specify a body, but I think we could add a body that has various options (like `user_initiated`) which we would just pass along in the HTTP body and Kafka. This way if we did want to add another flag we'd just update the body struct to accept the flag.

We'd then update the UI to pass that body with the option, and the checker jobs would just stay the way they are (no extra options passed).

On your end you would just check the body for the `options` map with the `user_initiated` flag and key off of that.

EDIT:
maybe something like
```json
{
	"source_id": 42,
	"options": {
		"user_initiated": true
	}
}
```

---

If you'd like I could do this as part of my work supporting kafka since it's a bit more involved than the query parameter approach. Let me know what you think!

lzap:
Sure that would be great, I guess I will close this proposal for now. Thanks for looking into this.

lindgrenj6:
No problem, I made another ticket to encapsulate this work, we'll keep it updated as we work through it. Shouldn't be more than a sprint or two out. 

https://issues.redhat.com/browse/RHCLOUD-22776

PR #608, opened 2022-10-14
[RHCLOUD-21533]Follow up PR to Liveness and Readiness Probe for Background Worker  

Ellen-Yi-Dong:
What: Adding readiness and liveness probe for background worker and added health check to background worker 

This is a followup PR from this PR: https://github.com/RedHatInsights/sources-api-go/pull/599

JIRA: [RHCLOUD-21533](https://issues.redhat.com/browse/RHCLOUD-21533)

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -64,17 +64,17 @@ objects:
           requests:
             cpu: ${AVAILABILITY_LISTENER_CPU_REQUEST}
             memory: ${AVAILABILITY_LISTENER_MEMORY_REQUEST}
-      readinessProbe:
-        httpGet:
-          path: /health
-          port: 8000
-        initialDelaySeconds: 1
+        readinessProbe:
+          httpGet:
+            path: /health
+            port: 8000
+          initialDelaySeconds: 1
         livenessProbe:
           httpGet:
             path: /health
             port: 8000
           initialDelaySeconds: 10
-        periodSeconds: 30
+          periodSeconds: 30
     - name: availability-status-listener
       minReplicas: ${{AVAILABILITY_MIN_REPLICAS}}
       podSpec:
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -64,6 +64,17 @@ objects:
           requests:
             cpu: ${AVAILABILITY_LISTENER_CPU_REQUEST}
             memory: ${AVAILABILITY_LISTENER_MEMORY_REQUEST}
+      readinessProbe:
+        httpGet:
+          path: /health
+          port: 8000
+        initialDelaySeconds: 1
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8000
+          initialDelaySeconds: 10
+        periodSeconds: 30
     - name: availability-status-listener
       minReplicas: ${{AVAILABILITY_MIN_REPLICAS}}
       podSpec:
--- a/jobs/worker.go
+++ b/jobs/worker.go
@@ -3,7 +3,6 @@ package jobs
 import (
 	"context"
 	"errors"
-	"fmt"
 	"net/http"
 	"time"
 
@@ -130,7 +129,6 @@ func (hc *BackgroundWorkerHealthChecker) healthChecker() {
 func (hc *BackgroundWorkerHealthChecker) healthCheckConsumer() {
 	for range hc.timeChannel {
 		hc.timeStamp = time.Now()
-		fmt.Println("timeStamp", hc.timeStamp)
 	}
 }
 
@@ -140,6 +138,8 @@ func (hc *BackgroundWorkerHealthChecker) healthcheckLoop() {
 		err := redis.Client.Ping(context.Background()).Err()
 		if err == nil {
 			hc.timeChannel <- time.Now()
+		} else {
+			l.Log.Fatalf("Failed to hit redis: %v", err)
 		}
 	}
 }
--- a/jobs/worker.go
+++ b/jobs/worker.go
@@ -3,10 +3,13 @@ package jobs
 import (
 	"context"
 	"errors"
+	"fmt"
+	"net/http"
 	"time"
 
 	l "github.com/RedHatInsights/sources-api-go/logger"
 	"github.com/RedHatInsights/sources-api-go/redis"
+	"github.com/labstack/echo/v4"
 )
 
 // the queue on redis we'll be sending the jobs to
@@ -57,8 +60,13 @@ func Run(shutdown chan struct{}) {
 		}
 	}()
 
+	//create struct for healthcheck struct and run the healthcheck thread
+	healthCheck := BackgroundWorkerHealthChecker{timeStamp: time.Now(), timeChannel: make(chan time.Time, 10)}
+	go healthCheck.healthChecker()
+
 	<-shutdown
 	shutdown <- struct{}{}
+
 }
 
 // Run a Job with a delay but in the background so it does _not_ block the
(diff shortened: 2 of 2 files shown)


deploy/clowdapp.yaml
```diff
@@ -64,6 +64,17 @@ objects:
...
+          httpGet:
+            path: /health
+            port: 8000
+          initialDelaySeconds: 1
+          livenessProbe:
+            httpGet:
+              path: /health
+              port: 8000
+            initialDelaySeconds: 10
+            periodSeconds: 30
```
lindgrenj6: This just needs indented 2 spaces to the left!

jobs/worker.go
```diff
@@ -57,8 +59,13 @@ func Run(shutdown chan struct{}) {
 		}
 	}()
 
+	//create struct for healthcheck struct and run the healthcheck thread
+	healthCheck := healthCheckStruct{timeStamp: time.Now(), timeChannel: make(chan time.Time, 10)}
```
lindgrenj6: We should probably rename the struct to not include `struct` in the name. What do you think of `BackgroundWorkerHealthChecker`?

jobs/worker.go
```diff
@@ -91,3 +98,45 @@ func RunJobNow(j Job) {
...
+	timeChannel chan time.Time
+}
+
+func (hc *healthCheckStruct) healthChecker() {
+	//run healthcheck loop and consumer in the background
+	go hc.healthcheckLoop()
+	go hc.healthCheckConsumer()
+
+	//create new GET request through echo with the same path name "/health"
+	e := echo.New()
```
lindgrenj6, line 114: Can we also tack
```go
	e.HideBanner = true
	e.HidePort = true
```
After this? So we don't get the ASCII art output to keep our logs pure JSON.

jobs/worker.go
```diff
@@ -91,3 +98,45 @@ func RunJobNow(j Job) {
...
+	e.GET("/health", func(c echo.Context) error {
+		//health should return a 500 if there hasn't been a response from ping request within 30 seconds
+		if hc.timeStamp.Before(time.Now().Add(-30 * time.Second)) {
+			return c.String(http.StatusInternalServerError, "500")
+		}
+		return c.String(http.StatusOK, "OK")
+
```
lindgrenj6: I think this empty line can go after the bracket.

jobs/worker.go
```diff
@@ -91,3 +98,45 @@ func RunJobNow(j Job) {
...
+	})
+	l.Log.Fatal(e.Start(":8000"))
+}
+
+// healthcheckConsumer runs a loop against the healthcheck time channel and updates the shared timestamp with the last time we received a response
+func (hc *healthCheckStruct) healthCheckConsumer() {
+	for {
```
lindgrenj6: We can probably change this to `for range hc.timeChannel` since we don't have to manually pull from the channel anymore.

jobs/worker.go
```diff
@@ -91,3 +98,45 @@ func RunJobNow(j Job) {
...
+		<-hc.timeChannel
+		hc.timeStamp = time.Now()
+	}
+}
+
+func (hc *healthCheckStruct) healthcheckLoop() {
+	//every 30 seconds, send a ping request to redis server. If ping request returns a <nil> error, then update the time
+	for range time.NewTicker(30 * time.Second).C {
+		redisErr := redis.Client.Ping(context.Background()).Err()
```
lindgrenj6: Can we rename this to `err`? All err's should be the same name.

redis/client.go
```diff
@@ -2,7 +2,6 @@ package redis
 
 import (
 	"fmt"
-
```
lindgrenj6: I think both of these changes could be removed - since we didn't have to touch this file.

lindgrenj6 (changes requested):
Nice work! I just have a bunch of teeny style changes :grin:

deploy/clowdapp.yaml
```diff
@@ -64,6 +64,17 @@ objects:
...
+        httpGet:
+          path: /health
+          port: 8000
+        initialDelaySeconds: 1
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 8000
+          initialDelaySeconds: 10
+        periodSeconds: 30
```
lindgrenj6:
```suggestion
        readinessProbe:
          httpGet:
            path: /health
            port: 8000
          initialDelaySeconds: 1
        livenessProbe:
          httpGet:
            path: /health
            port: 8000
          initialDelaySeconds: 10
          periodSeconds: 30
```

sorry, this is super complex. this _should_ work. this way you can click apply and it should take the diff for you.

jobs/worker.go
```diff
@@ -91,3 +99,47 @@ func RunJobNow(j Job) {
...
+	}
+}
+
+func (hc *BackgroundWorkerHealthChecker) healthcheckLoop() {
+	//every 30 seconds, send a ping request to redis server. If ping request returns a <nil> error, then update the time
+	for range time.NewTicker(30 * time.Second).C {
+		err := redis.Client.Ping(context.Background()).Err()
+		if err == nil {
+			hc.timeChannel <- time.Now()
+		}
```
lindgrenj6, line 143: I should have mentioned this earlier - can we add an `else` here that logs a failure to hit redis?

jobs/worker.go
```diff
@@ -91,3 +99,47 @@ func RunJobNow(j Job) {
...
+	})
+
+	l.Log.Fatal(e.Start(":8000"))
+}
+
+// healthcheckConsumer runs a loop against the healthcheck time channel and updates the shared timestamp with the last time we received a response
+func (hc *BackgroundWorkerHealthChecker) healthCheckConsumer() {
+	for range hc.timeChannel {
+		hc.timeStamp = time.Now()
+		fmt.Println("timeStamp", hc.timeStamp)
```
lindgrenj6: We probably don't need this print anymore.

lindgrenj6 (commented):
Getting closer!

jobs/worker.go
```diff
@@ -91,3 +98,48 @@ func RunJobNow(j Job) {
...
+}
+
+func (hc *BackgroundWorkerHealthChecker) healthcheckLoop() {
+	//every 30 seconds, send a ping request to redis server. If ping request returns a <nil> error, then update the time
+	for range time.NewTicker(30 * time.Second).C {
+		err := redis.Client.Ping(context.Background()).Err()
+		if err == nil {
+			hc.timeChannel <- time.Now()
+		} else {
+			l.Log.Fatalf("Failed to hit redis: %v", err)
```
lindgrenj6: This should be a `Warnf`, not a `Fatalf`. We don't want the process to exit on failure, the k8s liveness probe will do it for us. 

This will also allow the log to be uploaded to cloudwatch so we can see it in kibana and know "huh we started failing to hit redis at x:xx pm"

jobs/worker.go
```diff
@@ -91,3 +98,48 @@ func RunJobNow(j Job) {
...
+	go hc.healthCheckConsumer()
+
+	//create new GET request through echo with the same path name "/health"
+	e := echo.New()
+	e.HideBanner = true
+	e.HidePort = true
+	e.GET("/health", func(c echo.Context) error {
+		//health should return a 500 if there hasn't been a response from ping request within 30 seconds
+		if hc.timeStamp.Before(time.Now().Add(-30 * time.Second)) {
+			return c.String(http.StatusInternalServerError, "500")
```
lindgrenj6: And one last comment here - can we change the `"500"` message to `"Failed to hit redis for more than 30 seconds"`?

lindgrenj6 (commented):
Last comments!!

lindgrenj6 (approved):
Great work!

PR #609, opened 2022-10-17
[RHCLOUD-21270} Update openapi spec

Fynardo:
OpenAPI client checks are really strict, so adding some changes to normalize schemas definition


lindgrenj6 (approved):
Thanks Diego!

PR #615, opened 2022-10-19
Reorganize various Authentication DAOs

lindgrenj6:
Based on #607, so that one will need to go first. 

This PR does a few things:
1. Renames `authenticationDaoImpl` -> `authenticationDaoVaultImpl` since that is what it is
2. Moves `VaultClient` itnerface (and creation) to `dao/vault` since it is a separate package
3. Creates `dao/amazon` for the secrets manager client
4. Initial empty secrets manager authentication dao impl

PR #616, opened 2022-10-24
Add AWS SecretsManager client package

lindgrenj6:
based on #615, so that and #607 need to go first.

This PR is just adding the secrets-manager API client which I've abstracted to an interface so we can easily sub out localstack for testing. 

---

Ok this is ready for looking, basically this PR does a couple things:
1. Creates an interface `SecretsManagerClient` which abstracts away the actual secret manager, and implements one using the default aws sdk. The code there is pretty easy since the sdk is pretty good.
2. Adds deployment options pulling from the secret provided by AppSRE, it also includes the ephemeral setup for a localstack URL. I have to add an ephemeral env deployment for localstack still - but the code looks for localstack and points the aws sdk at that if it's there.

The next PR will be filling out the DAO to actually use this client.


Ellen-Yi-Dong (approved):

lpichler (approved):
LGTM!

PR #618, opened 2022-10-27
[RHCLOUD-22705] Only send Receptor based availability check if we don't have RHC Connnections

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-22705

Running into an issue on prod where RHC connection based sources that _also_ have an endpoint aren't accurately getting checked - we're sending a request to the worker too which overwrites anything that comes back from RHC. 

So we get into a weird state:
- RHC Conn is available
- Source is unavailable
- on availability check, the rhc conn is available (the same status) so we don't update either the source or rhc conn availability status
- ...which would get overridden anyway by the satellite worker

This pr mitigates that by only sending a request to the receptor worker if and only if there isn't an RHC Connection attached to the source. I also check both RHC + the Source's availability status on update so hopefully they wouldn't get out of sync like this again.

### DIFF
--- a/service/availability_check.go
+++ b/service/availability_check.go
@@ -68,12 +68,13 @@ func RequestAvailabilityCheck(c echo.Context, source *m.Source, headers []kafka.
 		ac.ApplicationAvailabilityCheck(source)
 	}
 
-	if len(source.Endpoints) != 0 {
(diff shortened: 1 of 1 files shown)


syncrou (approved):
LGTM 👍

PR #619, opened 2022-10-31
[RHCLOUD-21086] Implement SecretsManager DAO(s)

lindgrenj6:
https://issues.redhat.com/browse/RHCLOUD-21086

Based on #616, so that needs to go first.

This PR fleshes out the implementation for the secrets-manager Authentication DAOs.

The biggest changes are checking the authentication object's password field and reaching out to amazon to do changes there if necessary, or fetching it, etc.


dao/authentication_secrets_manager_dao.go
```diff
@@ -18,23 +21,157 @@ type authenticationSecretsManagerDaoImpl struct {
...
+			return err
+		}
+
+		err = auth.SetPassword(arn)
+		if err != nil {
+			return err
+		}
+	}
+
+	return a.authenticationDaoDbImpl.Create(auth)
```
lpichler: it looks that this can fail but secret in secret manager will be created.

lpichler:
When we will use secret manager we will still use database (I mean based on `authenticationDaoDbImpl` ) ?

dao/authentication_secrets_manager_dao.go
lindgrenj6: I will get that fixed across all the methods, thanks for catching it.

lindgrenj6:
> When we will use secret manager we will still use database (I mean based on authenticationDaoDbImpl ) ?

Yes, that's correct. 

Rather than storing the entire authentication record in secrets-manager (like we do for vault), we're only going to store the password field if it's present. This way we don't have to re-implement sorting/filtering and run into the performance issues like we would have with vault. It also keeps things simpler (the password was the only "secret" field, so why store the rest?).

It also saves us/the customer money - if they don't use superkey at this rate or developer sources, there won't have to be a secrets-manager record costing 40 cents per month.

---

I want to redo the vault DAO eventually to do the same thing - then we won't have to have all the extra id/extra fields since those fields will always be in the DB, it will just change where we store/fetch the password from.

lpichler:
> > When we will use secret manager we will still use database (I mean based on authenticationDaoDbImpl ) ?
> 
> Yes, that's correct.
> 


thanks for the info about reasons to do it with this approach - definitely it is better!

lpichler (approved):

PR #624, opened 2022-11-08
[RHCLOUD-21270] Update openapi bulk_create schema

Fynardo:
- Add app_creation_workflow to bulk_create payload
- Update auth schema to bulk_create response

### DIFF
--- a/public/openapi-3-v3.1.json
+++ b/public/openapi-3-v3.1.json
@@ -3753,8 +3753,7 @@
               },
               "required": [
                 "name",
-                "source_type_name",
-                "app_creation_workflow"
+                "source_type_name"
               ]
             }
           },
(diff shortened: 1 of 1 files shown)


public/openapi-3-v3.1.json
```diff
@@ -3741,11 +3741,20 @@
...
+                    "account_authorization"
+                  ],
+                  "type": "string"
                 }
               },
               "required": [
                 "name",
-                "source_type_name"
+                "source_type_name",
+                "app_creation_workflow"
```
lindgrenj6: I don't think we need to mark this as required - it will default to `manual_configuration` in the backend.

Fynardo: Makes sense, I'll remove it

lindgrenj6 (approved):

PR #625, opened 2022-11-21
Add support for FedRAMP environment

lindgrenj6:
Also includes a tiny refactor to constants for stage/prod/fedramp env vars. 
Also includes a method on the config struct checking if we're in a development env (e.g. NOT stage/prod/fedramp).


lindgrenj6:
Closing as not needed.

PR #627, opened 2022-11-28
RHCLOUD-23148 Instrument for Distributed Tracing

pilhuhn:
Instrument some parts of sources as example on how to do distributed tracing.

Bildschirmfoto 2022-11-28 um 10 58 11
Bildschirmfoto 2022-11-28 um 10 57 53


pilhuhn:
I think the checks may fail because the go version on test is older

lindgrenj6:
We actually already pass the [http request context](https://github.com/RedHatInsights/sources-api-go/blob/51dee1ae5ebb54d9187127e258aa5640d5d3f251/middleware/logging.go#L34) through to the db layer [when constructing the request params](https://github.com/RedHatInsights/sources-api-go/blob/51dee1ae5ebb54d9187127e258aa5640d5d3f251/dao/request_params.go#L27-L39), which gets set every time `<dao impl>.getDb()` [gets called](https://github.com/RedHatInsights/sources-api-go/blob/51dee1ae5ebb54d9187127e258aa5640d5d3f251/dao/source_dao.go#L64). Not sure if it means anything but the GORM logger has a [Trace](https://github.com/RedHatInsights/sources-api-go/blob/51dee1ae5ebb54d9187127e258aa5640d5d3f251/logger/gorm_logger.go#L25) method which logs how long stuff took - not sure if that would help here at all. 

Hopefully that makes things a bit easier - not having to mess with the DAO interfaces much due to that!

pilhuhn:
I am all ears for simplifying how that works. :)

lzap:
What is the status? We would love seeing that in.

Also I have a request - can you make sure that "trace_id" gets into logs too? That is a great way how to correlate log entries. We simply grab it from the context and put that into the main app logger:

https://github.com/RHEnVision/provisioning-backend/blob/24919edf2fbbc3d79ce7dc82c2fb63c00233ff0b/internal/middleware/logger.go#L25

lindgrenj6:
The status is that I'm over on FedRAMP fulltime currently - and this PR is going to require some work before it goes in. I haven't had a lot of time for it at all yet but plan on getting to this once I'm "back" on Sources at least part time.

Sometime around the end of March or shortly thereafter is what the current plan is. 

---

The insights request_id is a good way to follow a req through the system too since it gets added by 3scale.

gmcculloug:
Thanks @lindgrenj6.  Yes, the work to fully enable this tracing has a way to go before it become a real thing.  This was part of a proof-of-concept that we do intend to get into production, but we do not have a timeline for that at the moment.

One option with this specific PR is to leave the calls in place but make the code essentially inactive until we can get the required infrastructure in place to support this feature.

MikelAlejoBR:
I'm closing this PR for now. We can always reopen it if we end up switching to OpenTelemetry.

PR #634, opened 2023-01-17
Pass through x-rh-id when sending an http availability check request

lindgrenj6:
Per: https://redhat-internal.slack.com/archives/C0246P60U8H/p1673944289746899

Seems we should include the `Type` field since platform-go-middleware actively checks against it. I set it to `System` since this codepath only gets ran if its a PSK request (e.g. in-cluster)

I also changed the availability check operation to pass through the xrhid rather than generating one, since that seems to be "the right way" to do it.

### DIFF
--- a/service/availability_check.go
+++ b/service/availability_check.go
@@ -114,9 +114,16 @@ func (acr availabilityCheckRequester) httpAvailabilityRequest(source *m.Source, 
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


ezr-ondrej (approved):
I'd guess the parent request should always have the XRHID, so looks reasonable to me :+1: :orange_heart:

lpichler (approved):

PR #635, opened 2023-01-19
[RHCLOUD-23401] Removed enums from authentication validation

Fynardo:
Since the implementation of "Sources Secrets", this list is not "exhaustive" anymore, because authtype in Secrets is a regular string so it can store any value that may not match one of those values. This impacts listing authentications from an openAPI autogenerated client (like the Sources QE plugin), there is a validation that fails whenever it encounters one secret authentication (for example: notifications-secret-token)

### DIFF
--- a/public/openapi-3-v3.1.json
+++ b/public/openapi-3-v3.1.json
@@ -2722,29 +2722,8 @@
           },
           "authtype": {
             "type": "string",
-            "description": "The type of the authentication",
-            "enum": [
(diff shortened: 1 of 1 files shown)


petracihalova (approved):

lindgrenj6 (approved):
I like this change. I'm pretty sure we don't even do any validations on the backend for it anyway so it would have just been the clients sending this. But either way :+1:

PR #636, opened 2023-01-25
Add LaunchTemplate permissions to provisioning

lzap:
For a new feature, we need to be able to create and list Launch Templates on EC2. This adds those permissions. We would like this to get merged earlier before we go into production beta so there is no second version of sources in prod.

Edit: This patch also adds four IAM policy permissions which allows our app to validate permissions against expected state. We can then report exactly which permissions are missing to the user. There is no field we could return this to sources via availability check, that would be nice if you could add a string we would send back to sources to give meaningful errors to the user, that is something for the future. In the meantime, we will use this in our UI only.

There is an associated PR where all the AWS CLI commands for such permission check are shown if you are interested in more details. Thanks!


petracihalova (approved):

lindgrenj6 (approved):
Looks good, tested with superkey locally.

PR #637, opened 2023-02-10
Remove endpoint object from azure and google section

ezr-ondrej:
Remove the sections of endpoint, so Provisioning Source is created as application not an Endpoint.

I've replicated what we've done in #526 as we are facing the same issue for Azure (Source creating Endpoint resource instead of App resource).


ezr-ondrej:
@lindgrenj6 could you take a look pls? :)

lindgrenj6 (approved):
sorry - been swamped the last few days. LGTM, this will be on stage shortly.

PR #639, opened 2023-02-19
recognize X-Rh-Edge-Request-Id

lzap:
Sources currently recognizes `X-Rh-Insights-Request-Id`, however, there is also `X-Rh-Edge-Request-Id` which was recently recommended to us as the primary correlation id. This patch makes this HTTP header recognized and then stores it in the Echo context under the InsightsRequestID constant.

Reason: We proxy this header when calling sources so we can correlate log entries in ELK with our logs for each Insights Platform HTTP request.

### DIFF
--- a/middleware/headers.go
+++ b/middleware/headers.go
@@ -43,6 +43,10 @@ func ParseHeaders(next echo.HandlerFunc) echo.HandlerFunc {
 			c.Set(h.PSKUserID, c.Request().Header.Get(h.PSKUserID))
 		}
 
+		if c.Request().Header.Get(h.EdgeRequestID) != "" {
+			c.Set(h.EdgeRequestID, c.Request().Header.Get(h.EdgeRequestID))
+		}
+
 		if c.Request().Header.Get(h.InsightsRequestID) != "" {
 			c.Set(h.InsightsRequestID, c.Request().Header.Get(h.InsightsRequestID))
 		}
--- a/middleware/headers/headers.go
+++ b/middleware/headers/headers.go
@@ -7,6 +7,7 @@ const (
 	PSKUserID         = "x-rh-sources-user-id"
 	XRHID             = "x-rh-identity"
 	InsightsRequestID = "x-rh-insights-request-id"
+	EdgeRequestID     = "x-rh-edge-request-id"
 	ParsedIdentity    = "identity"
 	TenantID          = "tenantID"
 	UserID            = "userID"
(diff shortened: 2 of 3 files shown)


middleware/headers.go
```diff
@@ -43,6 +43,10 @@ func ParseHeaders(next echo.HandlerFunc) echo.HandlerFunc {
 			c.Set(h.PSKUserID, c.Request().Header.Get(h.PSKUserID))
 		}
 
+		if c.Request().Header.Get(h.EdgeRequestID) != "" {
+			c.Set(h.InsightsRequestID, c.Request().Header.Get(h.EdgeRequestID))
```
lindgrenj6: I think this needs to be `c.Set(h.EdgeRequestId ... `

lindgrenj6 (commented):
Looks good - just need one tiny piece to get this to show up on the log messages down the stack: 

```diff
diff --git a/middleware/logging.go b/middleware/logging.go
index 91f215c..3e91d8e 100644
--- a/middleware/logging.go
+++ b/middleware/logging.go
@@ -17,6 +17,7 @@ func LoggerFields(next echo.HandlerFunc) echo.HandlerFunc {
                acct := c.Get(h.AccountNumber)
                orgid := c.Get(h.OrgID)
                uuid := c.Get(h.InsightsRequestID)
+               edgeId := c.Get(h.EdgeRequestID)
 
                baseFields := logrus.Fields{
                        "uri":            uri,
@@ -25,6 +26,7 @@ func LoggerFields(next echo.HandlerFunc) echo.HandlerFunc {
                        "org_id":         orgid,
                        "user_agent":     agent,
                        "request_id":     uuid,
+                       "edge_id":        edgeId,
                }
 
                baseLoggerEntry := l.Log.WithFields(baseFields)
```

change the field to whichever you'd prefer - could even reuse the same header name.

lzap:
So the idea was to replace your current `request_id` with edge ID when present. But if you don’t like it, I am rebasing this PR to add a new field called `edge_id` like you propose. The problem is, this is only useful when all services use the same field name so in Kibana you can easily search this. But I guess in Kibana you can simply search by the unique string and it will search through all the fields so that should do it too.

lindgrenj6:
>So the idea was to replace your current request_id with edge ID when present.

hmm.. that might be ok. I need to check with the frontend folks to see what we return to the user, last I knew it was still just the insights-request-id. If they changed it to the edge id then I'm ok with replacing it when present!

lzap:
Actually, I just confirmed that if you enter a string in Kibana without any field name, it would search all field names. So instead entering `request_id = XXXX` you just enter `XXXX` and it will return you logs. So I don’t think it is a problem. It can be just slightly less intuitive :-)

lindgrenj6:
Ok, I'll merge it as is for now and ping you if I change it to the default `request_id` field!

PR #641, opened 2023-03-07
RHCLOUD-22885: Add source type entries for Azure scope and export_name addtions

chambridge:
* Add dao seeds

Related to: https://issues.redhat.com/browse/RHCLOUD-22885


lindgrenj6 (approved):
image


Tested out and it works!

PR #644, opened 2023-03-14
Cost Management - Add `bucket_region` for AWS sources

samdoran:
A new optional field specifying the bucket region is available for AWS sources in Cost Management. This field needs to be available during creation and editing of AWS sources.

### DIFF
--- a/dao/seeds/source_types.yml
+++ b/dao/seeds/source_types.yml
@@ -37,6 +37,12 @@ amazon:
         - type: required
         - type: pattern
           pattern: "^[A-Za-z0-9]+[A-Za-z0-9_-]*$"
+      - component: text-field
+        name: application.extra.bucket_region
+        label: S3 bucket region
+        validate:
+        - type: pattern
+          pattern: "^[a-z]+[a-z0-9-]*$"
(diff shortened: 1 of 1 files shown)


dao/seeds/source_types.yml
```diff
@@ -37,6 +37,12 @@ amazon:
         - type: required
         - type: pattern
           pattern: "^[A-Za-z0-9]+[A-Za-z0-9_-]*$"
+      - component: text-field
+        name: application.extra.bucket_region
+        label: S3 bucket region
+        validate:
+          - type: pattern
+            pattern: "^[a-z]+[a-z0-9-]*$"
```
lindgrenj6:
```suggestion
        - type: pattern
          pattern: "^[a-z]+[a-z0-9-]*$"
```

These just need to be moved in by one indentation (array of objects)

samdoran: No problem. YAML allows leading white space for lists but I'll change it to match the rest of the file.

lindgrenj6 (approved):

samdoran:
Thank you for merging.

PR #645, opened 2023-03-15
Pass through requestParams to RHC DAO and omit all associations when updating various models

lindgrenj6:
Pods are getting oomkilled on prod seemingly randomly - but it's always during an availability check. 

I'm wondering if it's something weird with GORM's association handling in the RHC availability check code - but need more logging to see which source it is exactly (I can see which tenant, but there are hundreds of sources underneath it).

### DIFF
--- a/dao/application_dao.go
+++ b/dao/application_dao.go
@@ -171,7 +171,7 @@ func (a *applicationDaoImpl) Create(app *m.Application) error {
 }
 
 func (a *applicationDaoImpl) Update(app *m.Application) error {
-	result := a.getDb().Updates(app)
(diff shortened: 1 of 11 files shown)


lindgrenj6:
So it looks like this is the issue. I really regret not omitting the associations always before, heres the grafana graph of container memory usage for the last 24hrs after I updated prod to temporarily use this branch: 
image


Each spike is when the loop would happen leading to an OOMKill. They stopped after switching and it seems to have helped make the memory usage a little more "stable" too, the lines are flatter.

petracihalova (approved):
well done !

lpichler (approved):

lpichler (approved):
Nice finding!

PR #646, opened 2023-03-24
Prevent a 500 when passed a 'bad' filter

lindgrenj6:
https://redhat-internal.slack.com/archives/C0246P60U8H/p1679677836774999

Found here. Probably needed to do this a while ago but this is a great excuse for it!


petracihalova (approved):
nice !!!

lpichler (approved):

PR #647, opened 2023-04-04
Make DELETE call kafka message production asynchronous

lindgrenj6:
Producing all the messages to kafka on any DELETE request can take an incredibly long time, I've timed it up to 8 SECONDS locally. This has been the cause for us missing our SLO targets.

This PR makes those message produce events async like we had on the rails api, though it is a bit more complicated. The gist of it is we create a mutex at the beginning of the subresource destroy operation and run the appropriate destroy message production in separate goroutines, we need the mutex because we always destroy the authentications last. The lock ensures the rest of the items get destroyed in order before destroying the authentications and raising those events. 

---

This takes the destroy time from seconds down to the regular 20-30ms since it is basically just validating that the resource is there and then deleting it from the db while a goroutine raises the events.

### DIFF
--- a/service/subresource_destroyer.go
+++ b/service/subresource_destroyer.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/RedHatInsights/sources-api-go/dao"
 	"github.com/RedHatInsights/sources-api-go/kafka"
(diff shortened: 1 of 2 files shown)


petracihalova (approved):
lgtm 👍

service/subresource_destroyer.go
```diff
@@ -44,66 +49,70 @@ func DeleteCascade(tenantId *int64, userId *int64, resourceType string, resource
 			return fmt.Errorf(`could not completely delete the source: %s`, err)
```
lpichler, line 49: @lindgrenj6 is it ok when it fails here ? Should not we unlock mutex also here ?

lindgrenj6: Yes, at this point we haven't launched any goroutines yet. The async part only really kicks off once the "parent" resource gets deleted, in this case the Source would have to completely succeed in deleting before launching the kafka producer routines.

lpichler: thanks!!

lpichler (approved):

PR #652, opened 2023-06-09
Add provisioning ephemeral service account

ezr-ondrej:
For ephemeral environments, Provisioning uses a different service account then stage, so we need to add it to seeds and make qa or ci environment set in ephemeral.


petracihalova:
-> https://github.com/RedHatInsights/sources-api-go/pull/655

PR #653, opened 2023-06-26
Add provisioning service account

adiabramovitch:
This PR adds the provisioning service account in order to fetch it from sources UI as we do for was account.


lindgrenj6 (approved):

adiabramovitch:
@lindgrenj6 Thank you!

PR #655, opened 2023-07-10
add provisioning ephemeral service account

petracihalova:
the same as here https://github.com/RedHatInsights/sources-api-go/pull/652 but for ephemeral environment


ezr-ondrej:
:orange_heart: this works for me! :+1:

PR #684, opened 2024-01-25
[RHCLOUD-29528] Support multiple kafka brokers

lindgrenj6:
just updating the kafka package to support using multiple brokers.


petracihalova (approved):
thank you for that

abaiken (approved):

PR #738, opened 2025-01-20
Remove not used Kafka topic platform.topological-inventory.operations…

dscervantes:
Remove not used Kafka topic platform.topological-inventory.operation-satellite

RHCLOUD-31321

### DIFF
--- a/service/availability_check.go
+++ b/service/availability_check.go
@@ -16,7 +16,6 @@ import (
 	l "github.com/RedHatInsights/sources-api-go/logger"
 	h "github.com/RedHatInsights/sources-api-go/middleware/headers"
 	m "github.com/RedHatInsights/sources-api-go/model"
-	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
 )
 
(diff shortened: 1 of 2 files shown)


MikelAlejoBR (commented):
Looks good to me! But I would actually like to request you to please remove the following things too:

- The entire `EndpointAvailabilityCheck` method and the subsequent `publishSatelliteMessage` method too, since we are no longer going to publish any more messages there.
- That will force you to modify the `availabilityChecker` interface to remove the `EndpointAvailabilityCheck` method, which is fine. Go ahead and remove the `publishSatelliteMessage` method from the interface too.
- In the `RequestAvailabilityCheck`, you will have to remove the `else if` branch as well.
- Finally, make sure you remove the unused `satelliteAvailabilityMessage` struct too.

With that we will have cleaned up all the satellite stuff from Sources :+1:

MikelAlejoBR:
Cool, the only thing left is to remove the `EndpointCounter` from the `dummyCheker` in `availability_check_test.go` and run `go fmt` in the `availability_check.go` file, and I think everything should pass.

MikelAlejoBR (approved):
LGTM! Good job!

PR #752, opened 2025-01-29
RHCLOUD-37558 | refactor: remove the goroutine when performing availability checks

MikelAlejoBR:
We want to make sure we process an availability request at a time for now, to not overwhelm the downstream services.

## Jira ticket
[[RHCLOUD-37558]](https://issues.redhat.com/browse/RHCLOUD-37558)


dehort (approved):
Looks good to me!

We should be able to remove the  cancel() calls in the `if` statements.

dehort (approved):
Looks good to me!

PR #764, opened 2025-02-07
[RHCLOUD-35770] Convert cloud-connector availability checks to use v2 interface

jessicarod7:
## Jira ticket

https://issues.redhat.com/browse/RHCLOUD-35570

## Description

Updates the `service/availability_check.pingRHC()` function to use the new V2 availability check API:

- Before: `PUT /v1/connection_status`
- After: `GET v2/connections/{client_id}/status`

Since the ID is now part of the URL, it needs to be assembled at runtime. This required some changes to the ClowdApp params/env vars:

- **Param** `CLOUD_CONNECTOR_CHECK_PATH` is now `CLOUD_CONNECTOR_BASE_PATH`, which defaults to `/connector`
- **Added new param** `CLOUD_CONNECTOR_STATUS_PATH` which defaults to `/status`. This is also an **env var**, and is used in the function to assemble the final URL

## Compatibility

ClowdApp parameters will need to be revised as described above

## Testing

Ran `make alltest` with no new failures (`dao/application_dao_test.TestApplicationCreateBadRequest()` currently fails on the main branch).


MikelAlejoBR (approved):
Good job! LGTM!

PR #806, opened 2025-03-10
 RHCLOUD-38530 | fix: the "background worker" and the "status listener" reported as down

MikelAlejoBR:
Prometheus is not able to reach the "background worker" and the "status listener" because those two services don't have a "metrics" endpoint. This is causing some alerts to be fired because the aforementioned pods are marked as "down", not "up". By creating a simple metrics' exporter for all of the services we will ensure that they show up as "up and running" for Prometheus.

## Jira ticket
[[RHCLOUD-38530]](https://issues.redhat.com/browse/RHCLOUD-38530)

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,16 +46,17 @@ func main() {
 	interrupts := make(chan os.Signal, 1)
 	signal.Notify(interrupts, os.Interrupt, syscall.SIGTERM)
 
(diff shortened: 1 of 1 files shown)


gwenneg (approved):
LGTM, thanks!

PR #808, opened 2025-03-14
RHCLOUD-38735 | feature: skip Cost Management applications in availability checks

MikelAlejoBR:
Enables being able to request sources from the Sources Monitor which are not empty —a source with just Cost Management applications is considered empty too— and skipping the availability status checks for Cost Management applications, to avoid doing the same work twice. More details in the Jira ticket.

## Jira ticket
[[RHCLOUD-38735]](https://issues.redhat.com/browse/RHCLOUD-38735)

### DIFF
--- a/internal_handlers.go
+++ b/internal_handlers.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/RedHatInsights/sources-api-go/middleware/headers"
 	"net/http"
 
 	"github.com/RedHatInsights/sources-api-go/dao"
@@ -42,9 +43,17 @@ func InternalSourceList(c echo.Context) error {
 		return err
 	}
 
+	// Skip Sources which do not have associated applications or RHC Connections. Useful for the Sources Monitor since
+	// it prevents returning Sources that do not need to have availability checks performed for them. More information
+	// here: https://issues.redhat.com/browse/RHCLOUD-38735.
+	var skipEmptySources = false
+	if skip := c.Request().Header.Get(headers.SkipEmptySources); skip != "" {
+		skipEmptySources = skip == "true"
+	}
+
 	// The DAO doesn't need a tenant set, since the queries won't be filtered by that tenant
 	sourcesDB := dao.GetSourceDao(nil)
-	sources, count, err := sourcesDB.ListInternal(limit, offset, filters)
+	sources, count, err := sourcesDB.ListInternal(limit, offset, filters, skipEmptySources)
 
 	if err != nil {
 		return err
--- a/internal_handlers.go
+++ b/internal_handlers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"github.com/RedHatInsights/sources-api-go/middleware/headers"
 	"net/http"
+	"strings"
 
 	"github.com/RedHatInsights/sources-api-go/dao"
 	"github.com/RedHatInsights/sources-api-go/util"
@@ -48,7 +49,7 @@ func InternalSourceList(c echo.Context) error {
 	// here: https://issues.redhat.com/browse/RHCLOUD-38735.
 	var skipEmptySources = false
 	if skip := c.Request().Header.Get(headers.SkipEmptySources); skip != "" {
-		skipEmptySources = skip == "true"
+		skipEmptySources = strings.ToLower(skip) == "true"
 	}
 
 	// The DAO doesn't need a tenant set, since the queries won't be filtered by that tenant
--- a/internal_handlers.go
+++ b/internal_handlers.go
@@ -2,11 +2,11 @@ package main
 
 import (
 	"fmt"
-	"github.com/RedHatInsights/sources-api-go/middleware/headers"
 	"net/http"
 	"strings"
 
 	"github.com/RedHatInsights/sources-api-go/dao"
+	"github.com/RedHatInsights/sources-api-go/middleware/headers"
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
 )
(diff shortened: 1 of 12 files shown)


source_handlers_test.go
```diff
@@ -1718,6 +1720,74 @@ func TestAvailabilityStatusCheck(t *testing.T) {
...
+
+	// Set the environment variables so that the URL to which we need to send the request gets properly built.
+	if err := os.Setenv("COST_MANAGEMENT_AVAILABILITY_CHECK_URL", "invalidurl"); err != nil {
+		t.Errorf("unable to set the URL for the Cost Management application: %err", err)
+	}
+	defer func() {
+		if err := os.Unsetenv("COST_MANAGEMENT_AVAILABILITY_CHECK_URL"); err != nil {
+			t.Errorf("unable to unset the URL for the Cost Management application: %err", err)
+		}
+	}()
```
dehort, line 1745: Where does the COST_MANAGEMENT_AVAILABILITY_CHECK_URL env var actually get used?

It would be nice if we could modify this setting directly and then inject that setting (or object) into the thing we are testing, but that probably isn't possible at the moment.  (Something we can work on improving later on...)

source_handlers_test.go
```diff
@@ -1718,6 +1720,74 @@ func TestAvailabilityStatusCheck(t *testing.T) {
...
+		},
+	})
+
+	// Override the "mock source" so that we return the new mock with the source we just added above. Then, we need to
+	// make sure to revert it for the rest of the tests.
+	previousMockSourceDao := mockSourceDao
+	defer func() {
+		mockSourceDao = previousMockSourceDao
+	}()
+	mockSourceDao = &mocks.MockSourceDao{Sources: customSourceFixtures, RelatedSources: fixtures.TestSourceData}
```
dehort, line 1776: I'm not sure I understand how this works.  What is using mockSourceDao?

internal_handlers.go
```diff
@@ -42,9 +43,17 @@ func InternalSourceList(c echo.Context) error {
 		return err
 	}
 
+	// Skip Sources which do not have associated applications or RHC Connections. Useful for the Sources Monitor since
+	// it prevents returning Sources that do not need to have availability checks performed for them. More information
+	// here: https://issues.redhat.com/browse/RHCLOUD-38735.
+	var skipEmptySources = false
+	if skip := c.Request().Header.Get(headers.SkipEmptySources); skip != "" {
+		skipEmptySources = skip == "true"
```
dehort: Might be good to call `strings.ToLower()` on skip?

service/availability_check.go
```diff
@@ -72,8 +72,14 @@ func RequestAvailabilityCheck(c echo.Context, source *m.Source, headers []kafka.
 
 // sends off an availability check http request for each of the source's
 // applications
-func (acr availabilityCheckRequester) ApplicationAvailabilityCheck(source *m.Source) {
+func (acr availabilityCheckRequester) ApplicationAvailabilityCheck(source *m.Source, skipEmptySources bool) {
 	for _, app := range source.Applications {
+		// We do not want to send availability check requests for Cost Management applications, as detailed in
+		// https://issues.redhat.com/browse/RHCLOUD-38735.
+		if skipEmptySources && app.ApplicationType.Name == "/insights/platform/cost-management" {
```
dehort: Nit-picky...what about moving the `app.ApplicationType.Name == "/insights/platform/cost-management"` logic into a method?  Something like the following `isCostManagementApp(app)`?

```
func isCostManagementApp(app) bool {
  return app.ApplicationType.Name == "/insights/platform/cost-management"
}
```

Would that help communicate what that bit of logic is actually doing (kind of self-commenting) and hide the details about how it does (in case the logic changes in the future)?

dehort (approved):
I think this looks reasonable though I'm struggling to grok it all.  :)  I'm still learning sources...

Would it be more clear if the skip-empty flag was passed in as a query parameter instead of a header?  Just a though!  :thinking:

source_handlers_test.go
MikelAlejoBR: The environment variable gets generated here:

https://github.com/RedHatInsights/sources-api-go/blob/312de4d8016ce44e6077cf26a871c751b239e99a/model/application_type.go#L48-L69

Which then gets used in the following code:

https://github.com/RedHatInsights/sources-api-go/blob/312de4d8016ce44e6077cf26a871c751b239e99a/service/availability_check.go#L75-L86

This is why I needed to do this ugly thing to make the test work. I do agree that a refactor on that would be nice, so that we grab the environment variables when Sources starts and don't do the above env var dance all the time.

source_handlers_test.go
MikelAlejoBR: `mockSourceDao` is used by all the tests in the test suite. So if I just modify it by adding a new source to the fixtures, the tests which expect a very specific set of sources in the responses actually break.

internal_handlers.go
MikelAlejoBR: Sounds good!

MikelAlejoBR:
The reason why I didn't go with a query parameter is because the query parameters are automatically parsed and appended to the underlying queries that get run. Therefore, if I try to send a request with a `?skip-empty-sources=true` query parameter, the database complaints because there is no `skip-empty-sources` column in it. So in order to make it work I have to be removing the stuff from the filters' slice:

```go
	for i, filter := range filters {
		if (filter.Name == "skip-empty-sources") && len(filter.Value) == 1 && filter.Value[0] == "true" {
			skipEmptySources = true

			if len(filters) != 1 {
				if i == 0 {
					filters = filters[:i]
				}

				tmp := filters[:i-1]
				filters = append(tmp, filters[i:]...)
			}

			break
		}
	}
```

Which is way worse, I feel like. I could extract it to a function, though, but I went for the easiest thing which was a custom header. What do you think about it?

service/availability_check.go
MikelAlejoBR: Done as well! Good idea!

PR #859, opened 2025-04-24
RHCLOUD-39709 | refactor: RBAC base path creation and RBAC client creation

MikelAlejoBR:
We get the RBAC service's URL from an environment variable but we should really get it from Clowder's configuration file, in case the service's hostname or port changes.

Also, since the legacy RBAC client requires the whole base path including the "/v1" suffix, and we might need to send requests to the "/v2" API soon, I refactored the way the RBAC client gets created so that we can easily extend or refactor it in the future.

## Jira ticket
[[RHCLOUD-39709]](https://issues.redhat.com/browse/RHCLOUD-39709)

### DIFF
--- a/middleware/authorization.go
+++ b/middleware/authorization.go
@@ -1,25 +1,22 @@
 package middleware
 
 import (
-	"context"
 	"fmt"
 	"net/http"
-	"os"
 	"regexp"
-	"time"
 
-	"github.com/RedHatInsights/rbac-client-go"
 	"github.com/RedHatInsights/sources-api-go/config"
 	h "github.com/RedHatInsights/sources-api-go/middleware/headers"
+	"github.com/RedHatInsights/sources-api-go/rbac"
 	"github.com/RedHatInsights/sources-api-go/util"
 	"github.com/labstack/echo/v4"
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
 var (
-	psks            = config.Get().Psks
-	bypassRbac      = config.Get().BypassRbac
-	rbacClient Rbac = &RbacClient{client: rbac.NewClient(os.Getenv("RBAC_URL"), "sources")}
+	psks       = config.Get().Psks
+	bypassRbac = config.Get().BypassRbac
+	rbacClient = rbac.NewRbacClient(config.Get().RbacHost)
 )
 
 /*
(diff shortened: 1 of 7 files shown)


rbac/rbac.go
```diff
@@ -0,0 +1,56 @@
...
+// Client represents an RBAC client with which we can communicate with that service.
+type Client interface {
+	// Allowed checks whether the given principal in the "x-rh-identity" header has the "sources:*:*" permission or
+	// not.
+	Allowed(string) (bool, error)
+}
+
+// rbacClientImpl is the internal implementation of the RBAC client.
+type rbacClientImpl struct {
+	// baseURL is the base URL of the RBAC service without the
```
jessicarod7: Is there supposed to be more to this comment?

jessicarod7 (commented):
Looks good, just noticed an odd comment

rbac/rbac.go
MikelAlejoBR: Yes, good catch! Thank you so much!

jessicarod7 (approved):
lgtm!

middleware/authorization.go
```diff
@@ -1,25 +1,22 @@
...
 	"github.com/redhatinsights/platform-go-middlewares/identity"
 )
 
 var (
-	psks            = config.Get().Psks
-	bypassRbac      = config.Get().BypassRbac
-	rbacClient Rbac = &RbacClient{client: rbac.NewClient(os.Getenv("RBAC_URL"), "sources")}
+	psks       = config.Get().Psks
+	bypassRbac = config.Get().BypassRbac
+	rbacClient = rbac.NewRbacClient(config.Get().RbacHost)
```
dehort: It would be nice to explicitly inject these into the `PermissionCheck()` method and remove the globals.  This would likely make testing easier.

It doesn't have to be done now...just something to think about.

dehort (approved):
Looks good to me!

middleware/authorization.go
MikelAlejoBR: Yes, wholeheartedly agree! For Kessel's refactor I was telling Dalia to pass the Kessel dependencies like you suggested, but I was unsure if I should do it in this PR too in case it'd be out of scope. However your comment has been a great excuse to do it!

Refactored!

MikelAlejoBR:
Great, I broke the PR. I was trying to trigger the PR check again and I combined closing the PR with force pushing, which apparently [makes it difficult for GItHub to track whether your force push is actually related to what there already existed in GitHub](https://github.com/isaacs/github/issues/361). I tried following [a workaround](https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c) but it doesn't work :sweat: 

I pushed the suggested changes to a new PR: https://github.com/RedHatInsights/sources-api-go/pull/860. @dehort, could you please take a look there?

Thank you both for your reviews!

PR #860, opened 2025-04-25
RHCLOUD-39709 | refactor: RBAC base path creation and RBAC client creation

MikelAlejoBR:
We get the RBAC service's URL from an environment variable but we should really get it from Clowder's configuration file, in case the service's hostname or port changes.

Also, since the legacy RBAC client requires the whole base path including the "/v1" suffix, and we might need to send requests to the "/v2" API soon, I refactored the way the RBAC client gets created so that we can easily extend or refactor it in the future.

## Jira ticket
[[RHCLOUD-39709]](https://issues.redhat.com/browse/RHCLOUD-39709)

### DIFF
--- a/middleware/authorization.go
+++ b/middleware/authorization.go
@@ -24,7 +24,7 @@ import (
 //     operation on a subset of paths.
 //   - The request is a regularly authenticated one, so we will call RBAC to verify that the principal that comes in
 //     the header has the authorization to perform the operation in Sources.
-func PermissionCheck(bypassRbac bool, rbacPsks []string, rbacClient rbac.Client) echo.MiddlewareFunc {
+func PermissionCheck(bypassRbac bool, authorizedPsks []string, rbacClient rbac.Client) echo.MiddlewareFunc {
 	return func(next echo.HandlerFunc) echo.HandlerFunc {
 		return func(c echo.Context) error {
 			switch {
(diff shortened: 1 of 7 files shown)


middleware/authorization.go
```diff
@@ -1,142 +1,114 @@
...
+//   - When "bypassRbac" is "true", all the requests are authenticated and authorized.
+//   - When using a "psk" in the request, the latter gets authorized if the PSK is in our list of authorized PSKs that
+//     can send requests to Sources.
+//   - Lastly, the requests that come with an "x-rh-identity" header must fulfill one of the following two conditions:
+//   - The request has been authenticated with a certificate, and it's been sent with an allowed "GET", "POST" or
+//     "DELETE" http verb. In the case that it's a "DELETE" request, it will only be authorized to perform that
+//     operation on a subset of paths.
+//   - The request is a regularly authenticated one, so we will call RBAC to verify that the principal that comes in
+//     the header has the authorization to perform the operation in Sources.
+func PermissionCheck(bypassRbac bool, rbacPsks []string, rbacClient rbac.Client) echo.MiddlewareFunc {
```
dehort: Minor nit-pick (i'm sorry...i can't help it):  The name `rbacPsks` makes me think the psks that are passed in are somehow used with rbac.  However, I think they are just used as a list of allowed psks, right?

dehort (approved):
Looks good to me!

middleware/authorization.go
MikelAlejoBR: Very much agreed, yes! I renamed it to `authorizedPsks` so that it its more generic and not tied specifically to RBAC! Thanks for the suggestion!

PR #870, opened 2025-04-29
chore: enable the "static check analyisis" linter 

MikelAlejoBR:
The static check linter will help us with things like deprecations.
Turning it on forced us to update the AWS client's set up and, for now,
ignore some warning about a deprecated method for the certificates since
there is no better alternative.


dao/amazon/secrets_manager_client.go
```diff
@@ -46,13 +62,13 @@ type SecretsManagerClient interface {
 	DeleteSecret(arn string) error
 }
 
-type SecretsManagerClientImpl struct {
+type secretsManagerClientImpl struct {
```
dehort, line 65: Nice!  Avoiding making the impl public...I like it.  :)

dao/amazon/secrets_manager_client.go
```diff
@@ -2,39 +2,55 @@ package amazon
...
 	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
 	"github.com/aws/aws-sdk-go-v2/service/secretsmanager/types"
 	"github.com/google/uuid"
 )
 
 var (
-	// base config, populated from initial setup connecting to the sources account
-	sourcesConfig *aws.Config
+	conf           = config.Get()
+	secretsManager SecretsManagerClient
```
dehort, line 20: Does this need to be a global?

dehort (approved):
Looks good to me!

dao/amazon/secrets_manager_client.go
MikelAlejoBR: Not really, the thing is that initializing it in the `main` function and passing the dependency down incurred on so many changes that I thought about leaving it for another PR.

The thing is that the secrets manager was designed as a singleton, so `NewSecretsManagerClient` gets called wherever it is used.

If you don't mind me throwing here a couple more hundred lines I have no problem in refactoring it, but if not I can open a new PR to fix this.

dehort: I think leaving it is fine.  We can refactor it and change that later on.  :+1:

PR #891, opened 2025-05-19
Add Validation for Subscription ID

dscervantes:
Adding validation for subscription ID, so that it is a UUID only and not allow spaces, blanks or any other symbols.

### DIFF
--- a/service/authentication_validation.go
+++ b/service/authentication_validation.go
@@ -34,7 +34,7 @@ func ValidateAuthenticationCreationRequest(auth *model.AuthenticationCreateReque
 	auth.ResourceType = util.Capitalize(auth.ResourceType)
 
 	if err = ValidateAzureSubscriptionId(auth); err != nil {
-		return fmt.Errorf("subscription ID is invalid")
+		return fmt.Errorf("subscription ID is invalid: %w", err)
 	}
 
 	return nil
@@ -65,7 +65,7 @@ func ValidateAzureSubscriptionId(auth *model.AuthenticationCreateRequest) error 
 			return errors.New("username must not be blank or empty for Azure Source Types")
 		}
 		if _, err := uuid.Parse(trimmed); err != nil {
-			return fmt.Errorf("The username must be a valid UUID for Azure Source Types")
+			return fmt.Errorf("the username must be a valid UUID for Azure Source Types")
 		}
 	}
 
--- a/service/authentication_validation.go
+++ b/service/authentication_validation.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/RedHatInsights/sources-api-go/model"
 	"github.com/RedHatInsights/sources-api-go/util"
+	"github.com/google/uuid"
 )
 
 var validAuthenticationResources = []string{"source", "endpoint", "application", "authentication"}
@@ -32,6 +33,10 @@ func ValidateAuthenticationCreationRequest(auth *model.AuthenticationCreateReque
 	// capitalize it so it's always the same format.
 	auth.ResourceType = util.Capitalize(auth.ResourceType)
 
+	if err := ValidateAzureSubcriptionId(auth); err != nil{
+		return fmt.Errorf("Subscription ID is invalid")
+	} 
+
 	return nil
 }
 
@@ -43,4 +48,27 @@ func ValidateAuthenticationEditRequest(auth *model.AuthenticationEditRequest) er
 	}
 
 	return nil
+}
+
+func ValidateAzureSubcriptionId(auth * model.AuthenticationCreateRequest) error {
+	if auth == nil {
+		return errors.New("auth is nil")
+	}
+
+	if auth.AuthType == "provisioning_lighthouse_subscription_id" || auth.AuthType == "lighthouse_subscription_id" {
+		if auth.Username == nil{
+			return errors.New("username is required for Azure Source Types")
+		}
+
+		trimmed := strings.TrimSpace(*auth.Username)
+		if trimmed == "" {
+			return errors.New("username must not be blank or empty for Azure Source Types")
+		}
+		if _, err := uuid.Parse(trimmed); err != nil {
+			return fmt.Errorf("The username must be a valid UUID for Azure Source Types")
+		}
+	}
+
+	return nil
+
 }
--- a/service/authentication_validation.go
+++ b/service/authentication_validation.go
@@ -33,8 +33,8 @@ func ValidateAuthenticationCreationRequest(auth *model.AuthenticationCreateReque
 	// capitalize it so it's always the same format.
 	auth.ResourceType = util.Capitalize(auth.ResourceType)
 
-	if err := ValidateAzureSubcriptionId(auth); err != nil{
-		return fmt.Errorf("Subscription ID is invalid")
+	if err = ValidateAzureSubscriptionId(auth); err != nil{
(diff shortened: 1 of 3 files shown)


service/bulk_create.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/RedHatInsights/sources-api-go/kafka"
 	l "github.com/RedHatInsights/sources-api-go/logger"
 	m "github.com/RedHatInsights/sources-api-go/model"
+	"github.com/RedHatInsights/sources-api-go/service"
```
MikelAlejoBR: The import is not needed because we are in the same package

service/bulk_create.go
```diff
@@ -291,6 +292,12 @@ func parseEndpoints(reqEndpoints []m.BulkCreateEndpoint, current *m.BulkCreateOu
 func linkUpAuthentications(req m.BulkCreateRequest, current *m.BulkCreateOutput, tenant *m.Tenant, userResource *m.UserResource) ([]m.Authentication, error) {
 	authentications := make([]m.Authentication, 0)
 
+	for _, auth := range authentications {
+		if err := service.ValidateAzureSubcriptionId(auth); err != nil {
```
MikelAlejoBR:
```suggestion
		if err := ValidateAzureSubcriptionId(auth); err != nil {
```

service/authentication_validation.go
```diff
@@ -44,3 +49,26 @@ func ValidateAuthenticationEditRequest(auth *model.AuthenticationEditRequest) er
 
 	return nil
 }
+
+func ValidateAzureSubcriptionId(auth * model.AuthenticationCreateRequest) error {
```
MikelAlejoBR:
```suggestion
func ValidateAzureSubscriptionId(auth *model.AuthenticationCreateRequest) error {
```

service/authentication_validation.go
```diff
@@ -32,6 +33,10 @@ func ValidateAuthenticationCreationRequest(auth *model.AuthenticationCreateReque
 	// capitalize it so it's always the same format.
 	auth.ResourceType = util.Capitalize(auth.ResourceType)
 
+	if err := ValidateAzureSubcriptionId(auth); err != nil{
+		return fmt.Errorf("Subscription ID is invalid")
```
MikelAlejoBR: Error messages should not be capitalized

```suggestion
		return fmt.Errorf("subscription ID is invalid")
```

service/authentication_validation.go
```diff
@@ -32,6 +33,10 @@ func ValidateAuthenticationCreationRequest(auth *model.AuthenticationCreateReque
 	// capitalize it so it's always the same format.
 	auth.ResourceType = util.Capitalize(auth.ResourceType)
 
+	if err := ValidateAzureSubcriptionId(auth); err != nil{
```
MikelAlejoBR: Since the `err` variable is already defined we can simply reassign it with `=` instead of redeclaring it with `:=`.

```suggestion
	if err = ValidateAzureSubscriptionId(auth); err != nil{
```

MikelAlejoBR (changes requested):
Looking good! I would suggest fixing the issues and then running the `go fmt` utility to format the files correctly and so that the CI jobs don't complain.

I would also ask you to add some tests in the `authentication_validation.go` file, to make sure that invalid subscription IDs will be detected as such.

dscervantes:
Hey Mikel for bulk create I think i am getting an error, because auth is not the same type am i supposed to add the loop i made to the existing one? Since I think there it is where its making the auth requests

service/bulk_create.go
```diff
@@ -291,6 +291,12 @@ func parseEndpoints(reqEndpoints []m.BulkCreateEndpoint, current *m.BulkCreateOu
 func linkUpAuthentications(req m.BulkCreateRequest, current *m.BulkCreateOutput, tenant *m.Tenant, userResource *m.UserResource) ([]m.Authentication, error) {
 	authentications := make([]m.Authentication, 0)
 
+	for _, auth := range authentications {
+		if err := ValidateAuthenticationCreationRequest(auth); err != nil {
```
MikelAlejoBR: You are trying to loop though the local `authentications` variable, instead of the authentications from the `BulkCreateRequest`. I think you just need to change this to `range req.Authentications` and then pass the `auth.AuthenticationCreateRequest` to your function.

dscervantes: Oh gotcha !

service/authentication_validation.go
```diff
@@ -32,6 +33,10 @@ func ValidateAuthenticationCreationRequest(auth *model.AuthenticationCreateReque
 	// capitalize it so it's always the same format.
 	auth.ResourceType = util.Capitalize(auth.ResourceType)
 
+	if err = ValidateAzureSubscriptionId(auth); err != nil {
+		return fmt.Errorf("subscription ID is invalid")
```
MikelAlejoBR: Here the `err` is getting discarded, because you're creating a new `Errorf` but you're not wrapping the previous error.

```suggestion
		return fmt.Errorf("subscription ID is invalid: %w", err)
```

service/authentication_validation.go
```diff
@@ -44,3 +49,26 @@ func ValidateAuthenticationEditRequest(auth *model.AuthenticationEditRequest) er
...
+		if auth.Username == nil {
+			return errors.New("username is required for Azure Source Types")
+		}
+
+		trimmed := strings.TrimSpace(*auth.Username)
+		if trimmed == "" {
+			return errors.New("username must not be blank or empty for Azure Source Types")
+		}
+		if _, err := uuid.Parse(trimmed); err != nil {
+			return fmt.Errorf("The username must be a valid UUID for Azure Source Types")
```
MikelAlejoBR:
```suggestion
			return fmt.Errorf("the username must be a valid UUID for Azure Source Types")
```

service/bulk_create.go
```diff
@@ -291,6 +291,12 @@ func parseEndpoints(reqEndpoints []m.BulkCreateEndpoint, current *m.BulkCreateOu
 func linkUpAuthentications(req m.BulkCreateRequest, current *m.BulkCreateOutput, tenant *m.Tenant, userResource *m.UserResource) ([]m.Authentication, error) {
 	authentications := make([]m.Authentication, 0)
 
+	for _, auth := range req.Authentications {
+		if err := ValidateAuthenticationCreationRequest(&auth.AuthenticationCreateRequest); err != nil {
+			return nil, fmt.Errorf("validation failed for authentication: %w", err)
+		}
+	}
+
```
MikelAlejoBR: I think we're gonna have to move this code block around, because I realized that the authentications don't get their IDs set at this point. The code below is what it links them up and adds the proper IDs in the model, so I would suggest the following:

1. Remove this code block.
2. Before the `// checking to make sure the polymorphic relationship is set.` line, assign `a.ResourceID` to `auth.ResourceIDRaw` and `a.ResourceType` to `auth.ResourceType`, and then call your function with the `&auth.AuthenticationCreateRequest`.

That should validate the authentications' payloads properly.

dehort:
Do these failures get logged anywhere?

MikelAlejoBR (changes requested):

MikelAlejoBR:
> Do these failures get logged anywhere?

What failures are you referring to?

MikelAlejoBR (approved):
Great work!

PR #892, opened 2025-05-19
Ensure that the updateRHCStatus is called every time

dscervantes:
Deleted the if statement to ensure that the updateRhcStatus function is called every time.


MikelAlejoBR (approved):
LGTM!

PR #931, opened 2025-06-27
Change config stringer method to print out the values 

dehort:
Currently, the stringer method only prints out values from the global config instance.  This could be very confusing if you tried to create a new instance, change a value in that instance and then printed out the values.

Change config stringer method to print out the values from the receiver instance and not the values from the global on each invocation


MikelAlejoBR (approved):
Looking great! LGTM!

PR #937, opened 2025-07-10
Deletion of Marketplace in Sources

dscervantes:
Removing the RH marketplace in Sources  https://issues.redhat.com/browse/RHCLOUD-39495

Remove the marketplace package.
Remove all the references that pointed to the marketplace package's functions.
Remove Red Hat Marketplace as a source type from the seed files.
Remove references to the Red Hat Marketplace in the OpenAPI specification.


MikelAlejoBR (approved):
LGTM!

PR #962, opened 2025-07-28
RHCLOUD-40797 Introduce OIDC authentication

gwenneg:
Assisted-by: Claude-4-Sonnet (via Cursor and Claude Code)

  ## Summary

  Adds JWT authentication support using OIDC with JWKS key discovery. This provides an additional authentication method alongside existing PSK and x-rh-identity auth.

  ## What's Added

  - JWT authentication middleware with JWKS validation
  - Feature flag: `sources-api.oidc-auth.enabled`
  - Configuration: `JWKS_URL` environment variable
  - Subject-based authorization support

  ## How It Works

  1. Validates `Authorization: Bearer <token>` headers
  2. Fetches signing keys from JWKS endpoint
  3. Verifies JWT signature and claims
  4. Extracts subject for authorization

  Falls back to existing PSK/x-rh-identity auth if JWT validation fails or no token provided.

  ## Configuration

  ```bash
  JWKS_URL="https://your-oidc-provider/.well-known/jwks.json"
  ```

  ## Files Changed

  - middleware/jwt_auth.go - Main JWT authentication logic
  - middleware/jwks.go - JWKS fetching and caching
  - middleware/authorization.go - Authorization integration
  - config/config.go - Configuration support
  - routes.go - Middleware integration

  Fully backward compatible - existing authentication methods continue to work unchanged.

  See JWT_AUTH_SETUP.md for detailed setup instructions and security considerations.


deploy/clowdapp.yaml
```diff
@@ -210,11 +211,11 @@ objects:
               optional: true
         - name: RBAC_HOST
           value: ${RBAC_HOST}
-        - name: SOURCES_PSKS
```
gwenneg, line 213: This was a duplicate. `SOURCES_PSKS` still exists a few lines above.

gwenneg, line 213: This was a duplicate. `SOURCES_PSKS` still exists a few lines above.

middleware/jwks.go
```diff
@@ -0,0 +1,204 @@
...
+type jwksCache struct {
+	mu     sync.Mutex
+	keySet jwk.Set
+	expiry time.Time
+}
+
+var globalJWKSCache = &jwksCache{}
+
+// FetchJWKS fetches JWKS with intelligent caching to optimize performance and reliability.
+// Caches valid JWKS for 1 minute to reduce load on identity providers while ensuring fresh keys.
```
alechenninger: 1 minute is actually very very short time frame for key changes. I would set this to more like 10 minutes or more. Key rotation happens even in a very modern deployment over the course of hours or days, almost never minutes. I think SSO keys rotate _maybe_ once a _year_. 1 minute would cause unnecessary amounts of IO in my opinion.

middleware/jwks.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	// Single lock to prevent race conditions
+	globalJWKSCache.mu.Lock()
+	defer globalJWKSCache.mu.Unlock()
+
+	// Return cached JWKS if it's not expired
+	if globalJWKSCache.keySet != nil && time.Now().Before(globalJWKSCache.expiry) {
+		return globalJWKSCache.keySet, nil
+	}
```
alechenninger: minor: you might want to apply double-checked locking here as a minor optimization for the happy path (https://en.wikipedia.org/wiki/Double-checked_locking)

e.g., first check the expiry without a lock, then only acquire the lock if expired and redo the check afterwards

middleware/jwks.go
```diff
@@ -0,0 +1,204 @@
...
+	globalJWKSCache.expiry = time.Now().Add(1 * time.Minute)
+
+	return keySet, nil
+}
+
+// secureJWKSFetch performs secure JWKS fetching with comprehensive safety controls.
+// Implements multiple defenses against attacks: HTTP timeouts, status validation,
+// Content-Type verification, response size limits, and key count restrictions.
+// These protections prevent DoS attacks, SSRF exploitation, and malicious JWKS responses.
+func secureJWKSFetch(ctx context.Context, jwksURL string) (jwk.Set, error) {
```
alechenninger: Are the max size, RSA, key number checks specified in an RFC somewhere? I've not seen these before and I wonder if they are necessary, or could create their own availability problems. For example, if a trusted issuer innocently has 11 keys temporarily, you now have a potentially significant outage. Or if they decide to switch to elliptic curve.

The issuer has to be trusted (and trusted issuers should be defined up front). If there is some compromise of the issuer response, we probably have bigger problems I wonder?

There also seems to be some internal inconsistency between what's allowed as a key (only RSA) vs what's allowed as a JWT algorithm (includes some ECs).

deploy/clowdapp.yaml
```diff
@@ -210,11 +211,11 @@ objects:
...
+        - name: AUTHORIZED_JWT_SUBJECTS
```
alechenninger, line 214: You may want these to be subject/issuer pairs. You only trust a single issuer now, so it's not really necessary, but if another issuer was ever added, you could get into trouble if these did not explicitly call out the issuer for _that_ subject at that time. (The same subject from an unintended issuer could get unintended access)

alechenninger (commented):
Thanks for the ping!

For the JWT validation piece, it is technically missing a few important things:

1. You should validate the issuer ("iss") is a known, preconfigured value (and immediately reject unexpected issuers)
2. You should validate the "aud" is _equal to or contains_ (it may be a string OR array) some expected preconfigured value that identifies yourself or your own trust domain (there should be a common aud for tokens meant for console services, though I forget what it is off hand)

Both of those should be configurable ideally, on a per-issuer basis.

I realized 2 is actually not present in the RBAC implementation. There, I was concerned with keeping compatibility with existing validation. I didn't want to introduce new behavior at first, to ensure the change was done without regressions. It should be added there, too, now that the implementation has been successfully deployed without regression.

Following those, I have an optional but recommended suggestion. Given you should have issuer as a config value, you can then make JWKS_URL **optional**. If an IdP follows the [OIDC Discovery Endpoint](https://openid.net/specs/openid-connect-discovery-1_0.html) standard (most do), then all you need is the issuer URL. You can go from that to -> discovery document -> JWKS URL, without requiring additional configuration for the JWKS URL, because the discovery document MUST be at a specific path from the issuer URL. So, if you have the issuer (which you would have to), you will be able to get the discovery document, and from there the JWKS URL. This has some other marginal benefits such as, e.g., allowing the JWKS URL to change without having to update any other configuration. 

Lastly, you may be wondering about "scope" validation, if you are looking at other OAuth guides. You could check for expected scope(s) on the token. Honestly though I probably would not do this. We are leaning towards using credentials _only_ for S2S authentication / only identity or SVID token (and not access token – confusing / subtle difference). For authorization context, instead of scopes, we will probably use something like the transaction token standard, which defines a more narrow concept than scope that is per-request instead of per-session ("purpose"). All to say, ignoring scope for now is what I would recommend.

I added some other comments inline for consideration :)

middleware/jwks.go
```diff
@@ -0,0 +1,204 @@
...
+	keySet, err := secureJWKSFetch(ctx, jwksURL)
+	if err != nil {
+		// Invalidate cache on error to prevent poisoning
+		globalJWKSCache.keySet = nil
+		globalJWKSCache.expiry = time.Time{}
+
+		logger.Log.Errorf("JWKS fetch failed from %s: %v", jwksURL, err)
+
+		return nil, fmt.Errorf("JWKS fetch failed: %v", err)
+	}
```
alechenninger: You might want to adjust this policy a bit to prevent outages.

You are adding an artificial expiring to the JWKS, but then if there is any issue retrieving a fresh one, failing. This means if there is some outage in the IdP, you unnecessarily have an outage yourself, because you could be otherwise using a perfectly fine cached JWKS that is only merely minutes or hours old.

If there is a failure, I would retain the previously retrieved JWKS.

Also, you may want to kick off the JWKS refresh asynchronously if you already have a previously fetched value, so retrieving the JWKS is rarely in the hotpath of a request. You use a request to notice the JWKS should be refreshed, then immediately use the current one, meanwhile a new one is refreshed in the background. This also prevents JWKS endpoint problems from impacting requests. Even if you fallback to a cached keyset, the requesters would all be queued up while waiting for a JWKS response. If none ever comes, that's a lot of waiting and potentially manifests in its own cascading failure.

deploy/clowdapp.yaml
gwenneg: Fixed in https://github.com/RedHatInsights/sources-api-go/pull/962/commits/416c523c38a6fb8dd562bf5337f67a124b6dbb11.

middleware/jwks.go
gwenneg: Fixed in https://github.com/RedHatInsights/sources-api-go/pull/962/commits/73fe4d715f051f13d93d3f8d6b82689c002afe86.

middleware/jwks.go
gwenneg: Fixed in https://github.com/RedHatInsights/sources-api-go/pull/962/commits/33fdb5498ccf5a4aecf4cf63b580f8f2808a0802.

gwenneg:
Single issuer validation added in https://github.com/RedHatInsights/sources-api-go/pull/962/commits/dfffb99f6bdc5efd414017cb86a14a0a3d6a79f6. This will be enhanced to multiple issuers validation in the future if needed.

middleware/jwks.go
gwenneg: Fixed in https://github.com/RedHatInsights/sources-api-go/pull/962/commits/3ac4b225b9c8b88a238b37daadbd6b6bded9fc06.

middleware/jwks.go
gwenneg: The max keys limit was a suggestion from Claude Sonnet. I didn't get that from an RFC. I agree this creates a risk and shouldn't be needed considering we trust the issuer.

Fixed in https://github.com/RedHatInsights/sources-api-go/pull/962/commits/9fb4d2763c9d9b72ed4c4d6475b2f944058ffe8e.

gwenneg:
Tests are flaky because of the async JWKS fetch. I'm fixing them.

middleware/jwks.go
gwenneg: Additional commit to fix flaky tests: https://github.com/RedHatInsights/sources-api-go/pull/962/commits/878fa9418922af90430a81740d6c3523a232a315.

middleware/jwks.go
gwenneg: Same story for the response body size limit.

Fixed in https://github.com/RedHatInsights/sources-api-go/pull/962/commits/e72a8a3f90f7fbf895d8f6771c5e37b523ec77d5.

gwenneg: Same story for the crypto algorithms and key strength.

Fixed in https://github.com/RedHatInsights/sources-api-go/pull/962/commits/231f664ab8e2fe719ce3f6c2642b5d037d7dd549.

gwenneg: All concerns from your comment should now be address now. TL;DR: the issuer is whitelisted up front and we will trust it.

gwenneg:
The JWKS URL is now retrieved from the issuer's OIDC discovery endpoint. See https://github.com/RedHatInsights/sources-api-go/pull/962/commits/c7cf11488aceb2cffd866b8e1b73d48cf785d8cb.

middleware/jwt_auth.go
```diff
@@ -0,0 +1,145 @@
...
+// Controlled by feature flag "sources-api.oidc-auth.enabled".
+//
+// Security: JWKS caching, 30s clock skew tolerance, subject length limits, timeout protection.
+func JWTAuthentication() echo.MiddlewareFunc {
+	return func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			// Feature flag check
+			if !service.FeatureEnabled(FeatureFlagOIDCAuth) {
+				return next(c)
+			}
```
dehort, line 71: What if we did not enable this using a feature flag?  What if we just enabled it at boot time by reading a configuration option?  I think it would simplify things if we avoided using a feature flag here.

service/feature_flags.go
dehort, line 1: If we do not use feature flags, then we should be able to remove these changes.

If we use feature flags, then we should probably change this to a interface and have one implementation for the tests (in memory, non-global map) and have another implementation that reaches out to the feature flag service.

It looks like nothing is really using the `FeatureEnabled()` method, so I don't think we're actually using feature flags anywhere else.

routes.go
```diff
@@ -50,6 +50,7 @@ func setupRoutes(e *echo.Echo, metricsService metrics.MetricsService) {
 			middleware.HandleErrors,
 			middleware.IdValidation,
 			middleware.ParseHeaders,
+			middleware.JWTAuthentication(),
```
dehort, line 53: If we do not use a feature flag and use a config option, then we could just add this middleware during startup.

middleware/user.go
```diff
@@ -27,6 +27,14 @@ func UserCatcher(next echo.HandlerFunc) echo.HandlerFunc {
...
 			userIDFromContext = userID
 
+		case c.Get(h.JWTSubject) != nil:
+			jwtSubject, ok := c.Get(h.JWTSubject).(string)
+			if !ok {
+				return fmt.Errorf("failed to pull JWT subject from request")
+			}
+
+			userIDFromContext = jwtSubject
+
```
dehort: Its unfortunate the add authentication type causes a change in this file.   Hmm...

middleware/headers.go
```diff
@@ -51,6 +52,29 @@ func ParseHeaders(next echo.HandlerFunc) echo.HandlerFunc {
...
+						c.Logger().Warn("JWT token ignored: exceeds maximum length")
+					} else if !isValidJWTFormat(token) {
+						c.Logger().Warn("JWT token ignored: invalid format")
+					} else {
+						c.Set(h.JWTToken, token)
+					}
+				}
+			}
+		}
+
```
dehort: I think this should be moved into the jwt auth middleware.  I think we need to start  moving / consolidating the JWT related code into one place.

It feels like moving this code is kind of breaking a pattern, but I think its the right thing to do.  We should not have so many "ripple" changes.

middleware/headers.go
```diff
@@ -92,3 +116,22 @@ func ParseHeaders(next echo.HandlerFunc) echo.HandlerFunc {
...
+	// Check each part is non-empty (header.payload.signature)
+	for _, part := range parts {
+		if part == "" {
+			return false
+		}
+	}
+
+	return true
+}
```
dehort: Same.  I think this needs to be moved to jwt_auth.go.

middleware/jwks.go
```diff
@@ -0,0 +1,214 @@
...
+// JWKS cache with async refresh and fallback capabilities to prevent outages
+type jwksCache struct {
+	mu          sync.RWMutex
+	keySet      jwk.Set
+	expiry      time.Time
+	refreshing  bool
+	lastFetched time.Time
+}
+
+var globalJWKSCache = &jwksCache{}
```
dehort: Couple of things...

Instead of implementing our own cache, could you use an existing cache like this [1]?

I think we should avoid using a global here.  The cache instance could be moved so that it is an instance variable of the DefaultJWKSDiscoverer struct.

[1] https://github.com/hashicorp/golang-lru#expirable-lru-cache-example

config/config.go
```diff
@@ -103,10 +112,16 @@ func (s SourcesApiConfig) String() string {
...
+	fmt.Fprintf(&b, "%s=%v ", "JWTIssuer", s.JWTIssuer)
 
 	return b.String()
 }
 
+// Reset clears the cached configuration - primarily for testing purposes
+func Reset() {
+	parsedConfig = nil
+}
+
```
dehort, line 124: This makes me think we are using too many globals somewhere.  Maybe we should work to remove the need for this method?

middleware/authorization.go
```diff
@@ -40,6 +41,21 @@ func PermissionCheck(bypassRbac bool, authorizedPsks []string, rbacClient rbac.C
...
+
+				jwtSubject, ok := c.Get(h.JWTSubject).(string)
+				if !ok {
+					return fmt.Errorf("error casting jwt subject to string: %v", c.Get(h.JWTSubject))
+				}
+
+				if !jwtClaimsMatches(authorizedJWTSubjects, jwtIssuer, jwtSubject) {
+					return c.JSON(http.StatusUnauthorized, util.NewErrorDoc("Unauthorized Action: JWT issuer/subject not authorized", "401"))
+				}
+
```
dehort: Its unfortunate the add authentication type causes a change in this file. Hmm...

Can we move this logic into the jwt_auth.go file?

dehort (changes requested):
I'm a bit confused about the FetchJWKS.  Instead of having this method, should we just use the JWKSDiscoverer interface.

Let me know if you like a hand with the Golang side of this and making it more of a library that we can move into a standalone package that we can share between projects.

middleware/jwks.go
gwenneg: Thanks for this suggestion! I can definitely leverage an [expirable LRU cache](https://pkg.go.dev/github.com/hashicorp/golang-lru/v2#readme-expirable-lru-cache-example) and its [PeekOrAdd](https://pkg.go.dev/github.com/hashicorp/golang-lru/v2#Cache.PeekOrAdd) method in this PR. I'll make some changes around that later today.

middleware/authorization.go
gwenneg: Is it really a problem to split authorization (this file) and authentication (jwt_auth.go)? They are two different security steps.

I have a concern with the changes above though: we should return a 403 instead of a 401, but I chose to mimic the PSK authorization code which returns a 401 to keep PSK and OIDC consistent. I'm not sure this is the right move.

service/feature_flags.go
gwenneg: I'm a big fan of feature flags so I'd rather keep them and introduce the interface as you suggested. I'll give it a try.

middleware/user.go
gwenneg: I agree. I'm not happy with this whole file, but I tried to keep the changes as minimal and close to PSK as possible. Maybe this could be refactored later?

middleware/headers.go
gwenneg: I agree. I'll move this.

gwenneg:
I found a better and much lighter way to cache the JWKS. Implementing it.

routes.go
gwenneg: I moved this to internal routes only.

middleware/user.go
gwenneg: I removed this entirely.

middleware/headers.go
gwenneg: Simplified and moved to `jwt_auth.go`.

middleware/headers.go
gwenneg: I removed this. The JWT library will take care of rejected wrong tokens.

middleware/jwks.go
gwenneg: JWKS are now cached using the `jwk` library which has a built-in cache which does exactly what I initially implemented manually.

config/config.go
gwenneg: This function is called from tests to reset the config after it's been loaded with a specific env variable value (`JWT_ISSUER`). That looks reasonable to me. Do you have a better suggestion?

middleware/authorization.go
gwenneg: This was changed but `authorization.go` still contains a minor code change. Ideally, it shouldn't be needed but that's a consequence of how PSK was implemented. I'll provide more details privately.

service/feature_flags.go
gwenneg: I'm surprised feature flags are not used yet in this project while the required infrastructure is kinda present already. The current changes from this PR related to feature flags are indeed not perfect, but I feel like this is an orthogonal issue caused by the feature flags initial (and unused) implementation. Do you mind if we change that with a separate Jira ticket at a later stage?

gwenneg:
I pushed a major refactor. This is ready for another review. The `JWT_AUTH_SETUP.md` file should contain everything you need to help you understand the latest version of this PR.

config/config.go
dehort: Here's what I was thinking of:  https://github.com/gwenneg/sources-api-go/pull/1

We can inject the trusted issuer into the authenticator/validator.  We can do this during the tests instead of manipulating the global config.

Ideally, in my mind, I'd love to get rid of these types of globals.  We should inject a copy of the config into the areas that need the config object.

gwenneg:
This PR is completely outdated and no longer needed. Closing.

PR #978, opened 2025-08-04
RHCLOUD-37754 | fix: metrics are making Sources panic

MikelAlejoBR:
The Prometheus client requires us to set all the tags even if they're empty, as otherwise it throws an "inconsistent label cardinality" error.

## Jira ticket
[[RHCLOUD-37754]](https://issues.redhat.com/browse/RHCLOUD-37754)


Fynardo:
lgtm

PR #1022, opened 2025-09-15
Update Cost AWS accounts by env

Fynardo:
RHCLOUD-42269

Cost updated their AWS accounts, moving from one "global" account to different accounts per env, so we have to update Superkey metadata accordingly


MikelAlejoBR (approved):
LGTM!

PR #1033, opened 2025-09-23
Available status by default

dscervantes:
[RHCLOUD-40737]
The goal is to update the "Bulk assembly" method in Sources so that when a combination of "GCP & Cloud Meter" entities get created, they get created with an "available" status by default.


MikelAlejoBR (approved):
LGTM!

PR #1034, opened 2025-09-24
RHCLOUD-42273 | Migrate app seeding data into app-interface / configMap

Fynardo:
This PR configures the service side of the migration of the application metadata from a local file into appInterface. The seeding logic stays the same but:

- Now metadata is expected to live in a file inside `/tmp/sources` folder (`/tmp` was the best fitting path I could find but subject to change)
- The metadata.yml file will be populated through a volume (both in appInterface using configMaps and docker-compose)
- Metadata file cannot be read with `embed` now so switching back to `os.Readfile`
- Metadata object has been simplified and has now one less "layer" as the env is not needed anymore. Each env will provide the correct associated metadata

### DIFF
--- a/dao/seeding.go
+++ b/dao/seeding.go
@@ -289,7 +289,7 @@ func seedSuperkeyMetadata(appTypes map[string]*m.ApplicationType) error {
 }
 
 func seedAppMetadata(appTypes map[string]*m.ApplicationType) error {
-	seeds := make(appMetadataSeedMap)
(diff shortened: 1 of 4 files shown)


MikelAlejoBR (approved):
LGTM! The only nitpick is that I would maybe put the file in `/mnt/sources` instead of `/tmp/sources`, but other than that this looks good!

PR #1042, opened 2025-09-29
Fix app_metadata seeding and testing in local env

Fynardo:
Migration of metadata into appInterface  / containers caused an issue in the local env where the app_metadata.yml is not correctly loading as it is not present in /mnt.

This PR addresses the conditional loading of the app_metadata based on env


MikelAlejoBR (approved):
LGTM!

PR #1052, opened 2025-10-14
Changing event types to int64

dscervantes:
[RHCLOUD-42371]

Changing event id from string to int 64.

### DIFF
--- a/model/event_models.go
+++ b/model/event_models.go
@@ -25,7 +25,7 @@ type ApplicationEvent struct {
 }
 
 type AuthenticationEvent struct {
-	ID        string    `json:"id"`
+	ID        int64     `json:"id"`
 	CreatedAt time.Time `json:"created_at"`
 
 	Name                    string                 `json:"name"`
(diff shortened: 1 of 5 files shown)


jessicarod7 (approved):
LGTM!

model/event_models.go
```diff
@@ -103,7 +103,7 @@ type SourceEvent struct {
 }
 
 type RhcConnectionEvent struct {
-	ID                      *string        `json:"id"`
+	ID                      *int64         `json:"id"`
```
dehort, line 106: Nit-pick (i'm sorry!):  Does this need to be a pointer?  I usually try to avoid using pointers unless really needed otherwise you end up needing to put `nil` pointer checks in the code, etc

In this case, I would think an RhcConnectionEvent would always have an ID so this would not need to be a pointer.

dehort (commented):
Do we know if other teams are currently using these events as they are being produced today?

This appears to be a change that will break an interface (the interface is basically a contract that defines how we are producing / formatting these events).  By changing the data type from `string` to `int`, then we might break existing clients which could be expecting the IDs to be strings.

If no one is using these events currently, then making a change like this is not a big deal.

PR #1107, opened 2025-12-08
Combined dependabot branch

dscervantes:
Might be too many updates, but want to check why the pull is braeaking


dehort (approved):
Looks good to me!

PR #1182, opened 2026-02-25
Redis to valkey

dscervantes:
The Redis library used in Sources is outdated, as the repository was renamed to [redis/go-redis](https://github.com/redis/go-redis). Additionally, we are migrating from using Redis to Valkey. In order to prevent issues from future incompatible Valkey releases, Sources should switch to using the Valkey client library, [valkey-io/valkey-go](https://github.com/valkey-io/valkey-go).

https://issues.redhat.com/browse/RHCLOUD-44598

### DIFF
--- a/db/migrations/migrations.go
+++ b/db/migrations/migrations.go
@@ -50,13 +50,13 @@ func Migrate(db *gorm.DB) {
 	// Using UUID as the lock value since it's a safe way of obtaining a unique string among all the clients.
 	uuid, err := uuidpkg.NewUUID()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


jessicarod7 (approved):
LGTM, thanks Dalia!

dehort (approved):
Looks good to me!

I wish the redis.Client was not a global, but maybe that is something we can cleanup later on.

I don't see where we are setting any timeouts for the valkey calls.  Does the valkey set a default timeout?  If it doesn't, then it might be a good time to add some timeouts.

PR #1212, opened 2026-03-30
Fix/logging

dscervantes:
Logs will include request_id, source_id, application_id and authentication_id. Handle errors logs every api error with those fields before returning the json response

https://redhat.atlassian.net/browse/RHCLOUD-45192

### DIFF
--- a/application_handlers.go
+++ b/application_handlers.go
@@ -115,6 +115,8 @@ func ApplicationCreate(c echo.Context) error {
 		return err
 	}
 
+	c.Logger().Infof("created application: id=%d source_id=%d application_type_id=%d", application.ID, application.SourceID, application.ApplicationTypeID)
+
 	accountNumber, err := getAccountNumberFromEchoContext(c)
 	if err != nil {
 		c.Logger().Warn(err)
@@ -202,6 +204,8 @@ func ApplicationEdit(c echo.Context) error {
 		return err
 	}
 
+	c.Logger().Infof("updated application: id=%d", id)
+
 	setNotificationForAvailabilityStatus(c, previousStatus, app)
 	setEventStreamResource(c, app)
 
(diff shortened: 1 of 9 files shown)


middleware/error_handling.go
```diff
@@ -1,14 +1,67 @@
...
+			fields["request_id"] = v
+			hasRequestID = true
+		}
+	}
+
+	if !hasRequestID {
+		if v := c.Request().Header.Get(h.EdgeRequestID); v != "" {
+			fields["request_id"] = v
+		}
+	}
```
dehort: I'm afraid this might get confusing as we'd be logging one request id (the akamai one?) as our request id.  🤔

dehort (changes requested):
This is looking good.  Can you add some more logging to the handler code to log when the various components (sources, applications, authentications) are created, deleted, modified?

Also, verify that the ids are printed out as part of the log messages.

This should help us to be able to more easily track what is happening, where things are failing, etc when troubleshooting production issues.

application_handlers.go
```diff
@@ -115,6 +115,8 @@ func ApplicationCreate(c echo.Context) error {
 		return err
 	}
 
+	c.Logger().Infof("created application: id=%d source_id=%d application_type_id=%d", application.ID, application.SourceID, application.ApplicationTypeID)
```
dehort: For each one of these log messages, it would be really helpful if we set the different ids as fields in the logging system like this:

`
logger.Log.WithFields(logrus.Fields{"tenant_id": *s.TenantID, "source_id": application.SourceID, "application_id": application.ID}).Infof("created application")
`

Some of the fields might already be set because we are pulling the `logger` instance from the `echo.Context`.   (Please verify this)

This becomes very useful when viewing our logs in kibana, cloudwatch, etc as those systems parse out these "structured" logs and then make those fields searchable.

We need to be careful and use the same field names throughout the code.  

Once this is implemented, then we can go into kibana (or cloudwatch) and search using the `source_id` field and it will show us all the logs that have a matching `source_id` field.  This is very useful when troubleshooting issues.

I hope that makes sense.  Let me know if it doesn't and we can hop on a call and discuss.

dehort (changes requested):

application_handlers.go
dscervantes: yes, when middleware.LoggerFields runs, the logger ont he echo context is already a *logrus.Entry with base fields.

dehort (approved):
Looks good to me!

PR #1231, opened 2026-04-22
fix(openapi): correct dependent_applications type

dscervantes:
## Summary

- Fixes the `dependent_applications` field type in the `ApplicationType` schema from `"object"` to `"array"` across all four OpenAPI spec versions (v1.0, v2.0, v3.0, v3.1)
- The API returns `dependent_applications` as an array (e.g. `[]`), but the spec incorrectly declared it as `"object"`, causing strict OpenAPI client validators (such as pydantic in OpenAPI Generator v7) to reject valid responses
- All existing Go tests continue to pass

Fixes [RHCLOUD-46824](https://issues.redhat.com/browse/RHCLOUD-46824)

## Test plan

- [ ] Verify `GET /application_types` response validates against the updated OpenAPI spec
- [ ] Confirm OpenAPI Generator v7 Python bindings no longer raise `ValidationError` for `dependent_applications`
- [ ] Run existing test suite (`go test ./...`) — all pass ✅

[RHCLOUD-46824]: https://redhat.atlassian.net/browse/RHCLOUD-46824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/model/application_type.go
+++ b/model/application_type.go
@@ -1,6 +1,7 @@
 package model
 
 import (
+	"encoding/json"
 	"net/url"
 	"os"
 	"strconv"
@@ -32,6 +33,14 @@ type ApplicationType struct {
 func (a *ApplicationType) ToResponse() *ApplicationTypeResponse {
 	id := strconv.Itoa(int(a.Id))
 
+	var depApps []string
+	if len(a.DependentApplications) > 0 {
+		_ = json.Unmarshal(a.DependentApplications, &depApps)
+	}
+	if depApps == nil {
+		depApps = []string{}
+	}
+
 	// returning the address of the new struct.
 	return &ApplicationTypeResponse{
 		Id:                           id,
@@ -39,7 +48,7 @@ func (a *ApplicationType) ToResponse() *ApplicationTypeResponse {
 		UpdatedAt:                    util.DateTimeToRFC3339(a.UpdatedAt),
 		Name:                         a.Name,
 		DisplayName:                  a.DisplayName,
-		DependentApplications:        a.DependentApplications,
+		DependentApplications:        depApps,
 		SupportedSourceTypes:         a.SupportedSourceTypes,
 		SupportedAuthenticationTypes: a.SupportedAuthenticationTypes,
 	}
--- a/model/application_type.go
+++ b/model/application_type.go
@@ -37,6 +37,7 @@ func (a *ApplicationType) ToResponse() *ApplicationTypeResponse {
 	if len(a.DependentApplications) > 0 {
 		_ = json.Unmarshal(a.DependentApplications, &depApps)
 	}
+
 	if depApps == nil {
 		depApps = []string{}
 	}
--- a/model/application_type_http.go
+++ b/model/application_type_http.go
@@ -12,7 +12,7 @@ type ApplicationTypeResponse struct {
 
 	Name                         string         `json:"name"`
 	DisplayName                  string         `json:"display_name"`
-	DependentApplications        datatypes.JSON `json:"dependent_applications"`
+	DependentApplications        []string       `json:"dependent_applications"`
 	SupportedSourceTypes         datatypes.JSON `json:"supported_source_types"`
 	SupportedAuthenticationTypes datatypes.JSON `json:"supported_authentication_types"`
 }
--- a/model/application_type_test.go
+++ b/model/application_type_test.go
@@ -2,8 +2,99 @@ package model
 
 import (
 	"os"
+	"reflect"
 	"testing"
+	"time"
+
+	"gorm.io/datatypes"
 )
+
+func TestToResponseWithDependentApplications(t *testing.T) {
+	now := time.Now()
+	at := ApplicationType{
+		Id:                    1,
+		CreatedAt:             now,
+		UpdatedAt:             now,
+		Name:                  "/insights/platform/cost-management",
+		DisplayName:           "Cost Management",
+		DependentApplications: datatypes.JSON(`["app1","app2","app3"]`),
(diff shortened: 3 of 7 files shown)


dehort:
This is an interesting issue.  The issue is that the dependent_applications is basically untyped in the code and in the database.

Here is the db model:  https://github.com/RedHatInsights/sources-api-go/blob/main/model/application_type.go#L22
Notice it is datatypes.JSON....so basically, store whatever is passed in as json in the database.

Here is the type that is returned by the REST requests:  https://github.com/RedHatInsights/sources-api-go/blob/main/model/application_type_http.go#L15
Notice it is again a datatype.JSON...so basically, any json type.

:thinking: 

I don't think it is possible to add `application_types` through the REST api.
The `application_type` data appears to be seeded by us and that is the only way it is set.  The `dependent_applications` is returned as an array because that is how it is seeded:  https://github.com/RedHatInsights/sources-api-go/blob/main/dao/seeds/application_types.yml#L4
Notice each `dependent_application` is set to `[]`.

So I think the question is...should we require that the dependent_applications data is always an array? (an array of strings)?  or should they be a json `any` type?

If we require dependent_applications to be an array, then we should probably modify the ApplicationTypeResponse to use `[]string` as the type for DependentApplication.

:thinking:

dehort:
Is dependent_applications used/set anywhere else in the code?

dscervantes:
Great observations! I dug into this more to answer your questions.

**Is `dependent_applications` used/set anywhere else in the code?**

Here's every place it appears:

| Location | Usage |
|----------|-------|
| `model/application_type.go:22` | DB model — `datatypes.JSON` |
| `model/application_type_http.go:15` | HTTP response — `datatypes.JSON` |
| `graph/schema.graphqls:123` | GraphQL schema — `Any` |
| `graph/schema.resolvers.go:66` | GraphQL resolver — returns the field as-is |
| `dao/seeding_types.go:20` | Seed struct — `interface{}` |
| `dao/seeding.go:169` | Seed logic — `json.Marshal` → stores to DB |
| `dao/seeds/application_types.yml` | All 5 entries — always `[]` |
| `application_type_handlers.go` | Only read endpoints (List, Get, SubCollectionList) — **no create/update** |

Note: `config.go`'s `findDependentApplication` is unrelated — it finds Clowder dependency endpoints, not this field.

**Should we require it to be `[]string`?**

I think yes — it would be safe and correct given:
- All seed data uses `[]` (empty arrays)
- No REST endpoint creates or modifies `ApplicationType` records
- The data is only ever set through seeding

I can update the PR to also change `ApplicationTypeResponse.DependentApplications` from `datatypes.JSON` to `[]string`. This would enforce the type at the Go level, not just in the OpenAPI spec. The DB model (`ApplicationType`) could stay as `datatypes.JSON` since GORM handles the JSONB→`[]string` scanning, or I could add a custom scanner type.

Would you prefer:
1. **Minimal** — keep the PR as-is (OpenAPI spec only)
2. **Stronger** — also change the HTTP response type to `[]string` to enforce the contract in code

Let me know which direction you'd like and I'll update accordingly.

dscervantes:
we can go with option 2

dscervantes:
Done! I've implemented option 2 — `ApplicationTypeResponse.DependentApplications` is now `[]string` instead of `datatypes.JSON`.

**Changes in this commit:**
- `model/application_type_http.go`: Changed `DependentApplications` field type from `datatypes.JSON` to `[]string`
- `model/application_type.go`: Updated `ToResponse()` to unmarshal the JSONB data into a `[]string` slice, with nil-safety (defaults to empty `[]string{}`)

The DB model (`ApplicationType`) stays as `datatypes.JSON` since that's what GORM needs for the JSONB column. The conversion happens at the response boundary in `ToResponse()`.

All tests and `go vet` pass.

dehort:
Should the openapi spec be adjusted to specify that the items within the array are strings?

Can we add a test to the ToResponse() method to verify that it works as expected?  Hopefully this test will help us catch issues where our implementation and openapi spec do not line up in the future.

I am very confident that the seed data is the only thing setting the dependent_applications data in the database, but it might be good to verify this with gabi.

dscervantes:
Done! Here's what I've added:

**OpenAPI spec update**: All four public OpenAPI specs (`v1.0`, `v2.0`, `v3.0`, `v3.1`) now specify `dependent_applications` items as `type: string` instead of the empty `items: {}`.

**ToResponse() tests**: Added four test cases to `model/application_type_test.go`:
- `TestToResponseWithDependentApplications` — verifies valid JSON array `["app1","app2","app3"]` unmarshals correctly
- `TestToResponseWithEmptyDependentApplications` — verifies `[]` produces an initialized empty slice (serializes to `[]` not `null`)
- `TestToResponseWithNilDependentApplications` — verifies nil input produces an initialized empty slice
- `TestToResponseWithInvalidDependentApplications` — verifies invalid JSON degrades gracefully to an empty slice

All tests pass. This should help catch any future mismatches between the OpenAPI spec and our implementation.

Regarding the gabi verification — I agree it's worth confirming that seed data is the only source setting `dependent_applications`. That check would need to be done by someone with gabi access to prod/stage environments.

dehort (dismissed):
Looks good to me!

dscervantes:
@dehort Your approval was dismissed because I pushed a minor whitespace fix after your review (added blank lines before `if` statements to satisfy the `wsl` linter). No logic changes — just style. Could you re-approve when you get a chance? Thanks!

dehort (approved):
Looks good to me!

PR #1232, opened 2026-04-23
refactor(kafka): reuse superkey producer

dscervantes:
## Summary

- Refactors the superkey Kafka producer to reuse a single long-lived writer instead of creating and destroying one per message
- The previous per-message create/destroy pattern reset the internal round-robin partitioner state, causing all messages to land on the same Kafka partition
- The shared writer is lazily initialized via `sync.Once` and closed during graceful shutdown

RHCLOUD-43917

## Test plan

- [x] All existing Go tests pass (`go test ./...`)
- [x] `go vet ./...` clean
- [x] `gofmt` clean
- [ ] Verify in stage that superkey messages are distributed across partitions (check Kafka consumer group lag per partition)


dehort (changes requested):
I would like to avoid having the `superkeyWriter` as a global.  I see a couple of ways to accomplish this.  While this requires a more significant change, I think it will also make automated testing easier.

What do you think?   What all changes would be required to avoid having the superkeyWriter be a global?

dscervantes (commented):
Thanks for the feedback! I've refactored the superkey producer to remove the global writer.

**What changed:**
- Removed the package-level `superkeyWriter`, `sync.Once`, `getSuperkeyWriter()`, and `CloseSuperkeyProducer()` from `service/superkey.go`
- Added an exported `ProduceSuperkeyMessage` function variable that gets injected from `main.go` — following the same dependency injection pattern the codebase uses for DAO functions
- The Kafka writer is now created and owned by `main.go`, keeping the service package free of global writer state
- Tests can swap `service.ProduceSuperkeyMessage` with a mock function, just like DAO function variables are swapped in existing tests

All tests pass. Let me know if you had a different approach in mind!

dehort (changes requested):
This looks better.  I know how the DAO injects functions using the function variables, but this is not as straightforward / clear as I would like.

What would it look like to move the SuperKey methods to a member function of a struct (SuperKey) and inject the kafka producer into an instance of the struct?  The SuperKey struct should be created in the main startup area of the code.

dscervantes (commented):
Good call — I've moved the superkey methods to a `SuperKeyService` struct and injected the Kafka producer (writer) into the struct instance.

**What changed:**
- Added `SuperKeyService` struct in `service/superkey.go` with a `producer *kafka.Writer` field
- `NewSuperKeyService(producer *kafka.Writer)` constructor
- `SendCreateRequest` and `SendDeleteRequest` are now methods on `*SuperKeyService`
- Removed the `ProduceSuperkeyMessage` function variable
- `SuperKeyService` is created in `main.go` during startup and assigned to `service.SuperKey`
- All callers (`application_handlers.go`, `service/bulk_create.go`, `jobs/superkey.go`) updated to use `service.SuperKey.SendCreateRequest()` / `service.SuperKey.SendDeleteRequest()`

Tests can create a `SuperKeyService` with a test writer, keeping dependency injection clean and explicit.

service/superkey.go
```diff
@@ -1,22 +1,33 @@
...
+// NewSuperKeyService creates a SuperKeyService with the given Kafka producer.
+// The producer should be a long-lived writer so the internal round-robin
+// partitioner distributes messages across all partitions.
+func NewSuperKeyService(producer *kafka.Writer) *SuperKeyService {
+	return &SuperKeyService{producer: producer}
+}
 
-var superkeyTopic = config.Get().KafkaTopic(superkeyRequestedTopic)
+// SuperKey is the shared instance, initialised in main.go during startup.
+var SuperKey *SuperKeyService
```
dehort: This makes the `SuperKey` instance a global.  What about creating the kafka producer and superkey instance in main and injecting the superkey into the methods where it is needed?

I would like to avoid globals where possible.  Injecting components in like this should make writing automated tests easier later on.

dehort (changes requested):

service/superkey.go
dscervantes: Good point — removed the `service.SuperKey` global entirely. The `SuperKeyService` is now created in `main()` and injected into each consumer:

- **Handlers** (`ApplicationCreate`, `BulkCreate`): receive it via a package-level var in `main`, set during server startup alongside the DAO function variables
- **`SendBulkMessages`**: takes `*SuperKeyService` as a direct parameter
- **Background worker**: `jobs.Run()` now accepts the `*SuperKeyService` and stores it for `SuperkeyDestroyJob` to use

The `service` package no longer has any package-level `SuperKey` state — the struct, constructor, and methods are all that remain. Tests can create a `SuperKeyService` with a mock writer and pass it directly.

dscervantes:
Hi @dehort — friendly bump on this PR. All feedback from your May 5 review has been addressed (SuperKey global removed, struct-based DI injected via main). Tests and lints are passing. Would you have a chance to take another look? Thanks!

dehort:
Can the bot handle me making a PR against its PR?  :grin:

hey bot, i'm thinking of an approach that looks like this:
https://github.com/dscervantes/sources-api-go/pull/1

Let's inject the dependencies in and start removing globals.  Hopefully this helps us add more testability in the future.

dscervantes:
Hey @dehort — great approach! I've merged your changes from [dscervantes/sources-api-go#1](https://github.com/dscervantes/sources-api-go/pull/1) into this branch.

The closure-based DI with the `SuperKeyProducer` interface is much cleaner — it eliminates the module-level `superKeySvc` variable entirely, routes dependencies through `setupRoutes`, and gives us a proper mock for tests. Build and `go vet` pass cleanly.

Let me know if you'd like any further adjustments!

dehort:
Bot, we still have a global.  Its in jobs/worker.go.  It is there because the middleware uses the superkeyservice.

The question is why is the delete logic split into being handled in the application_handler.go and the middleware/superkey_destroy.go.  The seems to be spreading the delete logic out into multiple places.  Is that right?  Why?  

Don't make any changes yet.  Am I looking at this correctly?  Wouldn't it make sense to move the delete logic from the middleware and put it into the application_handler.go.  This way the delete logic is more centeralized and maybe it makes it easier to test?

dscervantes:
Great questions — you're reading this correctly. Let me break down the current flow:

### The delete split

**Non-superkey application DELETE:**
1. Request → middleware chain → `SuperKeyDestroyApplication` checks `IsSuperkey(id)` → false → calls `next(c)` → `ApplicationDelete` handler → `service.DeleteCascade(...)` → 204 No Content (immediate DB delete)

**Superkey application DELETE:**
1. Request → middleware chain → `SuperKeyDestroyApplication` checks `IsSuperkey(id)` → true → enqueues `SuperkeyDestroyJob` to valkey → returns 202 Accepted (**handler never runs**)
2. Background worker picks up the job → `sendForApplication()` → `superKeySvc.SendDeleteRequest()` sends a Kafka message to the superkey-worker to tear down the cloud resources (e.g. AWS) → enqueues `AsyncDestroyJob` with a 15-second delay → `AsyncDestroyJob` eventually calls `service.DeleteCascade()` to clean up the DB

So the logic is split because the middleware acts as a gate — superkey deletes need to be async (cloud teardown first, then DB cleanup after a delay), while non-superkey deletes can be immediate.

### Moving delete logic to the handler

Your suggestion makes sense — the handler could check `IsSuperkey(id)` itself:
- Superkey → enqueue destroy job, return 202
- Not superkey → cascade delete, return 204

This would centralize the delete logic and make it more explicit. The same approach would apply to `SourceDelete` too, since `SuperKeyDestroySource` follows the same middleware pattern.

### The `jobs/worker.go` global

`var superKeySvc *service.SuperKeyService` in the jobs package is still a package-level variable, but it's different from the previous `service` package global — it's set once at startup in `Run(shutdown, sks)` and only used by `SuperkeyDestroyJob.sendForApplication()`. The reason it's a package-level var is that jobs are deserialized from JSON (from valkey), so the service can't be injected into the job struct directly.

Possible alternatives:
1. A job runner struct that holds the service and dispatches jobs
2. A `Run`-scoped closure passed into the job

But these would be more significant restructuring. Should I explore one of these approaches, or do you want to focus on centralizing the delete logic in the handler first?

dehort:
Bot, we are already in here.  Let's start cleaning this up.  Step by step.

Build some tests around the SuperKey delete rest api calls.  Test that the DELETE calls handled by the superkey_destroy middleware return the correct response (202 accepted).

Once that is complete, then move the logic from the SuperKeyDestroyApplication and SuperKeyDestorySource from the middleware into the handlers.  Make sure there are tests when doing this.

Make sure this change increases the testability of this logic.  It would be nice to get to a point where we can have tests that verify the handler logic (application delete, source delete, super key application delete, super key source delete) and also have some automated tests around the async super key delete logic.

Make the job runner an interface with an implementation of the interface based on the current code.

dscervantes:
Hey @dehort — I've implemented all the steps you outlined:

### 1. SuperKey delete tests (`superkey_delete_test.go`)
- `TestSuperKeyDeleteApplicationReturns202` — verifies superkey application DELETE returns 202 Accepted
- `TestSuperKeyDeleteSourceReturns202` — same for sources
- `TestNonSuperKeyApplicationDeleteReturns204` — non-superkey goes through cascade delete, returns 204
- `TestNonSuperKeySourceDeleteReturns204` — same for sources
- `TestSuperKeyDeleteApplicationJobContents` — verifies the enqueued job has correct identity/tenant/model
- `TestSuperKeyDeleteApplicationBadId` — bad id returns 400
- `TestSuperKeyDeleteApplicationNotFound` — missing app returns 404

Tests mock `jobs.Enqueue` (function variable) and the DAO layers to run without DB/valkey.

### 2. Moved delete logic from middleware to handlers
- `ApplicationDelete` and `SourceDelete` now check `IsSuperkey(id)` directly
- Superkey → `enqueueSuperKeyDelete()` → 202 Accepted
- Non-superkey → `service.DeleteCascade()` → 204 No Content
- Shared logic extracted to `superkey_handlers.go` (`enqueueSuperKeyDelete`)
- `SuperKeyDestroyApplication`/`SuperKeyDestroySource` middleware removed from route registration

### 3. JobRunner interface (`jobs/runner.go`)
- `JobRunner` interface with `Enqueue(j Job)` method
- `ValkeyJobRunner` struct implementing it (backed by valkey list)
- `jobs.Enqueue` is a function variable defaulting to `ValkeyJobRunner` — tests swap it with a mock

All tests passing. Let me know what you'd like to tackle next!

dehort (approved):
I think this looks like a step in the right direction!

Looks good to me!

PR #1255, opened 2026-06-12
Rhcloud-48159 mitigation

bonscji1:
mitigate know vulnerabilities, change go-version to go-version-file


Dockerfile
```diff
@@ -3,6 +3,10 @@ WORKDIR /build
 
 RUN dnf --assumeyes --disableplugin=subscription-manager install go
 
+#This was added to use never patch 1.26.4 required by mitigations,
+#it is safe to remove if go-toolset has correct version for our build
+ENV GOTOOLCHAIN=go1.26.4+auto
+
```
dehort: I think let's remove this for now and use golang 1.26.3 as that is what ships with the latest release of the ubi9 image.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/RedHatInsights/sources-api-go
 
-go 1.25.0
+go 1.26.4
```
dehort: Let's use golang 1.26.3 for now as that is what ships with the latest release of the ubi9 image.

dehort (changes requested):
Let's move to golang 1.26.3 to match what is on the latest ubi9 image.  I think that should fix our high/critical issues.

You can view the vulnerabilities per image here: https://quay.io/repository/redhat-user-workloads/hcc-integrations-tenant/sources/sources-api

dehort (approved):
Looks good to me!

PR #1257, opened 2026-06-22
Add Procedure to remove obsolete auth 

bonscji1:


dscervantes:
In line 36-39  and line 44-46 we have to full table scans when I think we can scan once. I think we can fix it by using the potgreSQL returning clause 
`  WITH deleted AS (
      DELETE FROM authentications
      WHERE authtype = 'notifications-basic-authentication'
      RETURNING id
  )
  SELECT ARRAY_AGG(id), COUNT(*)
  INTO deleted_ids_array, deleted_count
  FROM deleted;
`
Everything else looks good to me, I would just need to ask someone else for a pass before Merging it in!

dehort (changes requested):
Instead of putting this in a stored-procedure, what do you think about making it into a psql script (similiar to [1]) that is run as a clowder job [2]?

Clowder jobs run when the application gets deployed.  We could add the script to the repo, add the job to the deploy/clowdapp.yml, then push the change out to stage and prod, then remove the job and script once its no longer needed.

I feel like db migrations and stored-procedures are not the right place for one-off cleanup jobs.

I'd be happy to hop on a call and discuss, walk you through my idea, etc.  Let me know what you think!



[1] https://github.com/RedHatInsights/payload-tracker-go/blob/master/deployments/clowdapp.yml#L121
[2] https://inscope.corp.redhat.com/catalog/default/component/clowder/docs/usage/jobs/#job-and-cronjob-support

bonscji1:
Lets schedule a call then and discuss this, i would love to hear more about your proposed approach.

PR #1332, opened 2026-08-11
Add application subresource to REST filter parsing

dscervantes:
## Summary
- The REST middleware `parseFilter` only recognized `source_type` and `application_type` as subresources, missing `application`
- When the UI sends `filter[application][application_type_id]=1`, the middleware misparsed it as a column filter on `application`, which then failed the column allowlist check introduced in #1329, returning a 400 error
- This caused the UI to crash with `TypeError: can't access property "filter", e is undefined` because it received an error response where it expected data

## Test plan
- [ ] Verify `filter[application][application_type_id][eq]=1` correctly parses as a subresource filter
- [ ] Verify existing source_type and application_type subresource filters still work
- [ ] Verify the UI can filter sources by application without errors

### DIFF
--- a/middleware/filtering.go
+++ b/middleware/filtering.go
@@ -39,8 +39,7 @@ func parseFilter(c echo.Context) ([]util.Filter, error) {
 			}
 
 			// matching filter[subresource][field][operation]
-			// we only support filtering on source type/application type
-			if matches[1][0] == "source_type" || matches[1][0] == "application_type" {
+			if matches[1][0] == "source_type" || matches[1][0] == "application_type" || matches[1][0] == "application" {
 				filter = util.Filter{
 					Subresource: matches[1][0],
 					Name:        matches[2][0],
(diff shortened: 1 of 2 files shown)


dehort:
@dscervantes is this the correct way to call/test this?

  GET /sources?filter[application][application_type_id][eq]=1

It looks like i'm getting a 500 error.

dscervantes:
@dehort Yes, that's the correct format! GET /sources?filter[application][application_type_id][eq]=1 should work.

The 500 error is unexpected - the application subresource support was already implemented in the DAO layer (dao/filtering.go:149-159), this PR just adds it to the middleware's allowed list.

Could you share the error logs from the 500? That would help debug what's happening. A few things to check:

1. GORM relationship: The Source model does have Applications []Application defined, and the filter uses Joins("Applications") - but the join might need to be plural vs singular
2. Test data: Make sure there are sources with applications in your test environment
3. Database state: The join might be failing if there's a schema mismatch

If you can paste the stack trace, I can investigate further. The unit test in middleware/filtering_test.go:169 passes, so the parsing is correct - this looks like a runtime issue in the query execution.

